Issue #8, opened 2022-04-01
Update GEFs minimal Java version


vogella:
In platform we moved to Java 11 a long time ago, we should also update GEF to Java11.

azoitl:
done

PR #19, opened 2022-04-01
#8 Add automatic module name entry

azoitl:
In preparation for upgrading to Java 11 all manifest files have now an automatic module name entry.


vogella (commented):
Can we have this splitted by project? The version also needs to be increased, I suggested to update y the x.y.z by 1, e.g. 3.10.100 to 3.11.100

azoitl:
> Can we have this splitted by project? The version also needs to be increased, I suggested to update y the x.y.z by 1, e.g. 3.10.100 to 3.11.100

Sure. I split the commits such that the whole repo is always in the same state. But splitting per project sounds even better for me. 

I forgot to correctly set the API baseline by now I also noticed the version problem. Will fix that.

Issue #20, opened 2022-04-02
Replace all GEF Legacy mentions with GEF Classic


azoitl:
In many names and text through the GEF classic repo the old name GEF Legacy is used. For consistency and clean-up this should be updated.

vogella:
I see this text also in feature.xml. Can you also update the feature text?

azoitl:
Sure in the next days I wanted to work through each project step by step and update any description if possible. Can you please assign this issue to me.

azoitl:
Would you like again to have the changes per project? For most projects it would be just a single line.

Should the update site:
`http://download.eclipse.org/tools/gef/updates/legacy/releases/`
 also be changed as part of that issue. If yes I would change legacy to classic, or?

vogella:
> Would you like again to have the changes per project? For most projects it would be just a single line.

Feel free to combine.
> http://download.eclipse.org/tools/gef/updates/legacy/releases/

We should also create new update sites but I assume this will be done via #9 The old update sites should remain as they are to avoid breakage in GEF users.

vogella:
Looks like you fixed that @azoitl if you find more, just send an PR, no need for a new issue.

PR #24, opened 2022-04-02
[Issue #20] Cleaned-up README.md

azoitl:
Replaced all GEF legacy mentions with GEF classic. Updated the target
platform and fixed some minor typos.


vogella:
Thanks @azoitl

PR #26, opened 2022-04-04
Shorter description for target platform update

vogella:


vogella:
@azoitl please review

azoitl:
Unfortunately we should keep the original text. Because this section is for adopters who would like to build solutions on top of GEF Classic and use the snapshot builds. It will need some polishing after we are again able to provide snapshot builds and have a dedicated update side but for now I think it should stay as it was.

PR #38, opened 2022-04-05
Fix thumbnail not working on Mac

### LINKED ISSUE #4: Problem with thumbnail class on Mac OS
Hi Team,

I have a problem with [Thumbnail.java](https://github.com/eclipse/gef-legacy/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/parts/Thumbnail.java) in org.eclipse.draw2d.parts. On Mac the thumbnail image will be "rendered" multiple times (on Windows and Linux this problem is not appear). I supposed the problem was because [thumbnailGC.drawImage - line 199](https://github.com/eclipse/gef-legacy/blob/14563a9e1f2af636a5364d195cf07dbff6f35fa6/org.eclipse.draw2d/src/org/eclipse/draw2d/parts/Thumbnail.java#L199) like described in [https://bugs.eclipse.org/bugs/show_bug.cgi?id=543796](https://bugs.eclipse.org/bugs/show_bug.cgi?id=543796). 

My question; is there any plan to fix this issue or maybe create a workaround? (I read in README.md that this project is a legacy project/component which still maintained but will not developed further)

Phillipus:
Draft PR for discussion in order to workaround #4.

There is overhead in allocating new instances of GC, SWTGraphics, and ScaledGraphics, and there may be a better workaround, but this one is certainly tried and tested and at least shows the problem.

Refactor so that new instances of `GC`, `SWTGraphics`, and `ScaledGraphics` are created in the `createTileGraphics()` method that can be called in the `run()` method if on Mac

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/parts/Thumbnail.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/parts/Thumbnail.java
@@ -34,10 +34,15 @@ import org.eclipse.draw2d.geometry.Rectangle;
  * Figure.
  * 
  * @author Eric Bordeau
- * @author Alexander Ny√üen (anyssen)
+ * @author Alexander Nyﬂen (anyssen)
  */
 public class Thumbnail extends Figure implements UpdateListener {
-
+	
+	// Bug on Mac - images are cached
+	// See https://bugs.eclipse.org/bugs/show_bug.cgi?id=543796
+	// This affects macOS 10.14 and greater but we'll patch it for all macOS versions
+	private static final boolean isMac = "cocoa".equals(SWT.getPlatform()); //$NON-NLS-1$
+	
 	/**
 	 * This updates the Thumbnail by breaking the thumbnail {@link Image} into
 	 * several tiles and updating each tile individually.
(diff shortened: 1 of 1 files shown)


vogella:
@nnemkin and @azoitl any concerns here from your side?

azoitl:
I tested it under Linux, using the draw2d examples as well as the logic editor and they work fine with this changes. Also I don't see anything critical in the code.

Phillipus:
The changes are conditional on the OS being Mac, so Windows and Linux implementations remain the same.

We've been using this version in our RCP app for some years now without any issues. It probably could be done better but effectively does the job.

azoitl:
@Phillipus the main reason I wanted to test was that I saw some additional dispose calls. I find this calls good, I think it makes the Thumbnail code better behaving. But I wanted to be sure that there are no side effects.

Phillipus:
I used to have an even simpler workaround for the Thumbnail bug but that stopped working at Eclipse 4.13. I just tested that with latest Eclipse and it works. I might create a new PR with that one so we can compare.

vogella:
Thanks @Phillipus

vogella:
And thanks @azoitl for reviewing. I'm merging based on your review and the information from @Phillipus

PR #41, opened 2022-04-06
Adding signing to the build, similar as platform does it

vogella:
I copied the profile from platform and added it to the main pom for GEF.
Unfortunately AFAIK this cannot be tested locally.


vogella:
Unfortunately I' not aware of a way to test this without merginig..

PR #51, opened 2022-04-11
Change log

vogella:


azoitl:
The images in the .changelog directory are used in the CHANGELOG.md. so removing the renders this file invalid. The question is what to the with CHANGELOG.md in general. Should this be turned into something like release notes? 

But by scrolling through it I noticed there is a feature that I was not aware of, which would help me in my GEF usage.

vogella:
Ah, having a changelog is actually nice. Of course lets keep the images.

Issue #70, opened 2022-06-16
Please keep original feature IDs intact


merks:
I'm not sure I understand why the feature IDs have changed.

From this:

![image](https://user-images.githubusercontent.com/208716/174059083-1b869a20-a62c-4374-9138-97c4c535f8af.png)

To this:

![image](https://user-images.githubusercontent.com/208716/174059126-17dd4953-67a4-48c2-8132-5ee432a8af4c.png)

That's what causes this failure:

![image](https://user-images.githubusercontent.com/208716/174059174-607ab9e8-421b-42f8-80ca-8f6fa7e930bb.png)

It would be better to avoid changing the feature IDs.  

These direct dependencies are broken by that:

![image](https://user-images.githubusercontent.com/208716/174059206-48baa424-3495-40e6-a13d-d28137ef38b6.png)

If you start disabling those, you need to disable what's downstream from those, and eventually you'll get this far:

![image](https://user-images.githubusercontent.com/208716/174059011-efdcb966-c81e-47ec-a0d4-2f3357137584.png)

At that point you'll notice you need to disable EMF because it depends on Data Tools which is broken, and then you can pretty much disable 2/3 of the train.

This doesn't sound like a viable approach...

laeubi:
> From this:
> To this:

Maybe i get it worn,g but should "gef-classic" not actually be using *totally different* feature names to not interfere with "non classic" and not change any existing "legacy" stuff?

That would make most sense to me... so consumers of "non classic" won't be affected and "legacy" consumers can migrate?

laeubi:
@vogella just curious: Why does tycho-pomless requires feature id changes at all? the id should be derived from the feature itself so I could understand that the **artifactId** might change but not the other way round...

howlger:
> @vogella just curious: Why does tycho-pomless requires feature id changes at all? the id should be derived from the feature itself so I could understand that the **artifactId** might change but not the other way round...

As far as I know, a feature with the same ID as a plugin (e.g. `org.eclipse.draw2d` feature and `org.eclipse.draw2d` plugin) causes a conflict in POM-less Tycho, having two artifacts with the same ID.

laeubi:
You can define an alternative artifactId in the `build.properties`
```
artifactId = whateveryoulike
```

merks:
[...]

PR #71, opened 2022-06-16
Keep  keep original feature IDs intact

merks:
Also add the org.eclipse.draw2d.sdk to the category.xml

https://github.com/eclipse/gef-classic/issues/70


vogella:
LGTM, thanks @merks

PR #77, opened 2022-06-17
Add api baseline target for 3.14.0 release

azoitl:
Added a target platform with the 3.14 release for use as api base line in pde tools.


vogella:
I would have assumed that https://github.com/eclipse/gef-classic/blob/caf52ec022514d1fd0632f7f82ba7774b656d40c/pom.xml#L19 would have to be changed to update our baseline.

azoitl:
This is a change that I also have to do. This pull request is a convenience pull request. GEF Classic always had the baseline as part of its repo so that you could set it in your pde preferences. Up to know this were always the full binaries. I noticed that I now can use a target platform file. So I wanted to try this out. May should have wrote more text in the pull request.

PR #88, opened 2022-07-06
[Modernize Code] typefy the list in groupRequest

azoitl:


azoitl:
This work needs a major update and in its current form it is not mergeable.

PR #81, opened 2022-06-25
[Modernize Code] Turn Editpart children into typed List

azoitl:
One of the biggest drawbacks of current GEF Classic code is that it is not yet using typed generics. This requires a lot of checks and casts in client code. This commit is a start to move GEF Classic to a more modern java.

PR #89, opened 2022-07-10
[Modernize Code] change getEditpart children so that it gets easier

azoitl:
castable


vogella:
@[pcdavid](https://github.com/pcdavid) can you check if that fixes GMF?

azoitl:
This pull request is not needed anymore as the base commit it depends on was reverted. See discussion at PR #81.

PR #95, opened 2022-08-05
Workaround thumbnail not working on Mac

Phillipus:
- See https://github.com/eclipse/gef-classic/issues/4
- Draw a copy of tileImage to GC


Phillipus:
This is an alternative method to workaround issue #4 

Instead of re-creating new instances of `tileGC`, `tileGCGraphics`, and `tileGraphics` as in #38 we simply make a copy of `tileImage` if on Mac and use that copy.

This might be a more lightweight workaround than #38

Phillipus:
I did some timing tests and copying the image is much slower than creating new instances of `tileGC`, `tileGCGraphics`, and `tileGraphics` so I'll close this one.

PR #99, opened 2022-08-12
removed unnecessary redrawing of zest Graph upon scrolling

BauePhil:
As the Viewport already redraws the Graph on demand redrawing upon moving the scrollbars is unnecessary and reduces performance in bigger graphs.


azoitl:
Checked a bit further the update and redrawing is already handled by the base class and the LWS so it is safe to merge this pull request.

PR #101, opened 2022-08-19
[Modernize Code] use generic version of getAdapter

azoitl:
This pull request updates all getAdapter methods to the generic version also used in the Eclipse Platform. This greatly improves the usability of the Adapter interface in GEF, as clients do not need to to perform instanceof and casts any more. 

@vogella @Phillipus if you have time I would be happy for a quick look. But I think I followed the Eclipse Platforms style and therefore it shouldn't have an impact on users. 

I tested with Eclipse 4diac IDE where we make quite some use of the getAdapter interface and it worked as expected.

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/LogicEditor.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/LogicEditor.java
@@ -174,10 +174,9 @@ public class LogicEditor extends GraphicalEditorWithFlyoutPalette {
 					"org.eclipse.gef.examples.logic.outline.contextmenu", //$NON-NLS-1$
 					provider, getSite().getSelectionProvider());
 			getViewer().setKeyHandler(getCommonKeyHandler());
-			getViewer()
-					.addDropTargetListener(
-							(TransferDropTargetListener) new TemplateTransferDropTargetListener(
-									getViewer()));
+			getViewer().addDropTargetListener(
+					(TransferDropTargetListener) new TemplateTransferDropTargetListener(
+							getViewer()));
 			IToolBarManager tbm = getSite().getActionBars().getToolBarManager();
 			showOutlineAction = new Action() {
 				public void run() {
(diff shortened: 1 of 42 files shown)


Phillipus:
I've just gone through all of our GEF forked code and made the same changes. All our unit tests pass, and the RCP app seems to be working all OK.

azoitl:
@Phillipus thanks for testing. Would it be a lot of work and duplicate your test setup so that I have a second RCP?

Phillipus:
> Would it be a lot of work and duplicate your test setup so that I have a second RCP?

Not sure what you mean? Archi code is here:

https://github.com/archimatetool/archi

We have our version of GEF/Draw2d/Zest as bundles that I copied last year.

azoitl:
> > Would it be a lot of work and duplicate your test setup so that I have a second RCP?
> 
> Not sure what you mean? Archi code is here:
> 
> https://github.com/archimatetool/archi
> 
> We have our version of GEF/Draw2d/Zest as bundles that I copied last year.

Ah yes. Sorry didn't remember that. But for testing I would always need to merge my changes into your repo. I guess this is not the right place here, but what can we do that you don't need a full GEF classic copy anymore?

Phillipus:
> but what can we do that you don't need a full GEF classic copy anymore?

My copy has bug fixes, theme colors for palette, and most importantly an option not to use `ScaledGraphics` in Draw2d:

See https://github.com/archimatetool/archi/commits/master/org.eclipse.draw2d

azoitl:
> > but what can we do that you don't need a full GEF classic copy anymore?
> 
> My copy has bug fixes, theme colors for palette, and most importantly an option not to use `ScaledGraphics` in Draw2d (it causes problems with font clipping at higher zoom and scaling):
> 
> See https://github.com/archimatetool/archi/commits/master/org.eclipse.draw2d

Ah yes the infamouse ScaledGraphics problems. Thanks for tracking this down. We also deactivated the ScaledGraphics in 4diac IDE based on your findings. But it seems you had more issue then we had.

Phillipus:
> Ah yes the infamouse ScaledGraphics problems. Thanks for tracking this down. We also deactivated the ScaledGraphics in 4diac IDE based on your findings. But it seems you had more issue then we had.

Opened #106 so we can carry on there...

azoitl:
Given the feedback and my tests I'll merge this PR now.

PR #108, opened 2022-08-23
Check for possible null value to avoid NPE

Phillipus:
- Under some circumstances currentState.relativeClip can be null and will throw a NPE
- This checks for a null value

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java
@@ -335,7 +335,10 @@ public class SWTGraphics extends Graphics {
 			sharedClipping = false;
 
 			boolean previouslyApplied = (appliedState == currentState.relativeClip);
-			currentState.relativeClip = currentState.relativeClip.getCopy();
+			// Fix: currentState.relativeClip can be null and lead to NPE
+			if (currentState.relativeClip != null) {
+				currentState.relativeClip = currentState.relativeClip.getCopy();
+			}
 			if (previouslyApplied) {
 				appliedState.relativeClip = currentState.relativeClip;
 			}


Phillipus:
We encountered this NPE in our RCP app here:

https://github.com/archimatetool/archi/issues/431

It can happen in a diagram if a connection line Figure is disabled.

Phillipus:
The value can be set to null in `appliedState.relativeClip = currentState.relativeClip = null;` here:

```java
	public void setClip(Path path) {
		initTransform(false);
		if (((appliedState.graphicHints ^ currentState.graphicHints) & FILL_RULE_MASK) != 0) {
			// If there is a pending change to the fill rule, apply it first.
			gc.setFillRule(((currentState.graphicHints & FILL_RULE_MASK) >> FILL_RULE_SHIFT)
					- FILL_RULE_WHOLE_NUMBER);
			// As long as the FILL_RULE is stored in a single bit, just toggling
			// it works.
			appliedState.graphicHints ^= FILL_RULE_MASK;
		}
		gc.setClipping(path);
		appliedState.relativeClip = currentState.relativeClip = null;
	}
```

org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java
```diff
@@ -335,7 +335,10 @@ private void checkSharedClipping() {
 			sharedClipping = false;
 
 			boolean previouslyApplied = (appliedState == currentState.relativeClip);
-			currentState.relativeClip = currentState.relativeClip.getCopy();
+			// Fix: currentState.relativeClip can be null and lead to NPE
+			currentState.relativeClip = currentState.relativeClip != null
```
azoitl, line 339: wouldn't a simple if be nicer here? setting it to null when it is already null seems for me a bit confusing.

Phillipus:
> wouldn't a simple if be nicer here? setting it to null when it is already null seems for me a bit confusing.

Yep, Force pushed a change.

azoitl:
Thanks for the rework and the patch.

Argh, I still don't know how to merge and comment at the same time.

Phillipus:
Thanks for merge!

Issue #109, opened 2022-08-23
Changes to the code adds extra formatting


Phillipus:
When I make a change in the GEF code and save it, it is formatted according to the `.settings` files. That's OK, but the problem is when you submit a PR it's harder to see the relevant diff amongst all the other changes. I guess we just have to live with it?

azoitl:
The problem I have with the `.settings`files is that the settings there are a bit out dated. Therefore I created the PR #103 to use the default Eclipse settings. Not sure if this is a good idea. But as you point out above I think it would be a good idea to do one pull request with just reformatting all the java files to whatever setting we are choosing. I did the same in 4diac IDE and this reduced this formatting noise a lot and we got more modern formatted code.

Phillipus:
Sounds like a good idea.

azoitl:
Then I will do that after my current pull requests are merged. Probably will do that in the next days.

In 4diac IDE we have a common set of workspace settings that we share and import. This ensure the same formatting but we also have some save actions like clean-up and formating. Not sure if this is also something for GEF. I'll attach the 4diac settings here. Happy to get some feedback.

[4diacSettings.epf.txt](https://github.com/eclipse/gef-classic/files/9410189/4diacSettings.epf.txt)

Had to add the txt file ending otherwise Github was not happy.

azoitl:
As discussed I just created PR #110, in which I reformatted all the Java files to the current default eclipse style. I hope this reduces the noise in future commits.
I sofar didn't do any automatic clean-ups that Eclipse would be capable of.

azoitl:
With PR  #110 merged I tihnk it is safe to close this issue.

Phillipus:
>With PR https://github.com/eclipse/gef-classic/pull/110 merged I tihnk it is safe to close this issue.

OK, thanks for the PR!

PR #110, opened 2022-08-24
Addresses #109 reformated all java files

azoitl:
With removing the project specific formatters and migrating to default eclipse formatting this PR reformated all java file. This should ensure a uniform formatting as well as reducing future formatting noise.


azoitl:
Fixes #109

PR #112, opened 2022-09-13
enable custom graph nodes and connections

BauePhil:
Seems like the original Idea was to make the graph nodes and connections extendable (i.e. [CGraphNode](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/CGraphNode.java)) and just have not been implemented in a way to accomplish this. I added exposed create methods and added the original object creation in the overridden methods in the GraphViewer.

### DIFF
--- a/org.eclipse.zest.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.zest.core/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.zest.core;singleton:=true
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
-Bundle-Version: 1.6.200.qualifier
+Bundle-Version: 1.7.0.qualifier
 Require-Bundle: org.eclipse.zest.layouts,
  org.eclipse.ui;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.draw2d;visibility:=reexport
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
@@ -230,7 +230,7 @@ public abstract class AbstractStructuredGraphViewer extends AbstractZoomableView
 	HashMap getNodesMap() {
 		return this.nodesMap;
 	}
-
+	
 	GraphNode addGraphModelContainer(Object element) {
 		GraphNode node = this.getGraphModelNode(element);
 		if (node == null) {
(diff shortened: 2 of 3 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
```diff
@@ -250,12 +250,18 @@ GraphNode addGraphModelNode(IContainer container, Object element) {
 		}
 		return node;
 	}
+	
+	/** Implement and return the new node object, enables to define custom graph nodes*/
+	protected abstract GraphNode createNodeObject(final Graph graphModel, final int style, IFigure figure);
```
azoitl, line 255: why is this method here abstract and not as the original implementation does return the new CGraphNode. If you don't do this here clients using a different viewer wouldn't work anymore. So I think the changes you did in the GraphViewer should be done here and the GraphViewer should stay unchanged.

org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
```diff
@@ -250,12 +250,18 @@ GraphNode addGraphModelNode(IContainer container, Object element) {
...
+
+	/** Implement and return the new connection object, enables to define custom graph connections*/
+	protected abstract GraphConnection createConnectionObject(final Graph graphModel, final int style, final GraphNode source, final GraphNode destination);
```
azoitl, line 258: See my comment above

org.eclipse.zest.core/META-INF/MANIFEST.MF
```diff
@@ -4,7 +4,7 @@ Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.zest.core;singleton:=true
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
-Bundle-Version: 1.6.200.qualifier
+Bundle-Version: 1.7.0.qualifier
```
azoitl, line 7: I think we also need to update the feature version in this case.

azoitl (changes requested):

org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
```diff
@@ -250,12 +250,23 @@ GraphNode addGraphModelNode(IContainer container, Object element) {
 		}
 		return node;
 	}
+	
+	/** Implement and return the new node object, enables to define custom graph nodes*/
+	protected GraphNode createNodeObject(final Graph graphModel, final int style, IFigure figure) {
```
azoitl, line 255: unfortunately with your update the since tag is missing can you please also add some descrpition of the parameters and the return value

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
```diff
@@ -614,7 +614,10 @@ private void doUpdateFigure(Connection connection) {
 		connection.setToolTip(toolTip);
 	}
 
-	private PolylineArcConnection createFigure() {
+	/**
+	 * @since 1.7
+	 */
+	protected PolylineConnection createFigure() {
```
azoitl, line 620: as it is now API please add some description.

PR #120, opened 2022-10-07
Updated target platform to 2022-09 and to Tycho 3.0.0

azoitl:
@vogella as we got the Jenkins problem I did the long due upgrade to the latest Eclipse platform. As part of that I would have a best practice question. Does it make sense to have an individual target file per Eclipse release or is it simpler to just have a gef target plaform file that we regularly update?


azoitl:
\maven

PR #121, opened 2022-10-07
Update maven.yml

azoitl:
upgraded to java 17 as tycho 3.0. needs Java 17

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -11,11 +11,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 11
+    - name: Set up JDK 17
       uses: actions/setup-java@v2
       with:
         distribution: 'adopt'
-        java-version: '11'
+        java-version: '17'
     - name: Cache local Maven repository
       uses: actions/cache@v3
       with:


azoitl:
@vogella @Phillipus do you see any issues building GEF with Java 17? I would love to have our build on the latest Eclipse Tycho.

Phillipus:
Will it still run under 11 for end users?

azoitl:
> Will it still run under 11 for end users?

Yes definitely. But I may have to change something in my other pull request.

Phillipus:
@azoitl If Tycho is building with Java 17 and we have:

```
<maven.compiler.source>11</maven.compiler.source>
<maven.compiler.target>11</maven.compiler.target>	
```

Will it actually create 11 compliant byte code? Normally one would use a `toolchains.xml` file to specify a path to JDK 11 otherwise Tycho complains with:

`[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml`

azoitl:
I hope so. But I'm not a maven expert. Any suggestions are warmly welcome here.

Phillipus:
> I hope so. But I'm not a maven expert. Any suggestions are warmly welcome here.

I think it should be OK from my tests with building Archi with these settings. But also not an expert. @vogella WDYT?

Issue #122, opened 2022-10-31
Replace deprecated methods isEnqueued()


Phillipus:
There are two instances of `isEnqueued()`. This method has been deprecated since Java 16:

https://github.com/eclipse/gef-classic/blob/9e560cf2b2d63e03d6721a9346895d4ce44471f2/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalViewerKeyHandler.java#L218

https://github.com/eclipse/gef-classic/blob/9e560cf2b2d63e03d6721a9346895d4ce44471f2/org.eclipse.gef/src/org/eclipse/gef/tools/SelectionTool.java#L141

The JavaDoc says:

> Deprecated. This method was originally specified to test if a reference object hasbeen cleared and enqueued but was never implemented to do this test.This method could be misused due to the inherent race condition or without an associated ReferenceQueue. An application relying on this method to release critical resources could cause serious performance issue. An application should use `ReferenceQueue` to reliably determine what reference objects that have been enqueued or `refersTo(null)` to determine if this reference object has been cleared.

It seems to me the easiest thing would be to replace with `refersTo(null)` like this:

`cachedNode.refersTo(null)` and `cachedHandlePart.refersTo(null)`

JavaDoc:

> Tests if the referent of this reference object is obj. Using a null obj returns true if the reference object has been cleared.

WDYT?

Phillipus:
> It seems to me the easiest thing would be to replace with `refersTo(null)` like this:

Scrap that idea because I just realised that `refersTo(null)` was introduced in Java 16 so this won't compile on earlier versions.

vogella:
Should we move GEF to Java 17? Simrel already uses this one.

Phillipus:
> Should we move GEF to Java 17? Simrel already uses this one.

Are there downstream clients still on 11?

azoitl:
Yes. Especially when you are using Xtext or Xtend your are currently stuck with Java 11 max. This will change with the next release of Xtext. But not sure yet how stable it will be.

Phillipus:
I'm not even sure this line does anything useful anyway:

```
if (cachedNode.isEnqueued())
            return null;
```

The line following is this:

```
return (GraphicalEditPart) cachedNode.get();
```

JavaDoc says about `get()`

> Returns this reference object's referent. If this reference object has been cleared, either by the program or by the garbage collector, then this method returns null.

So it will return null anyway.

azoitl:
[...]

PR #123, opened 2022-11-01
Remove redundant isEnqueued() checks

Phillipus:
- The isEnqueued() check is not needed because get() will return null if this reference object has been cleared

- See https://github.com/eclipse/gef-classic/issues/122

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/tools/SelectionTool.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/tools/SelectionTool.java
@@ -55,7 +55,7 @@ public class SelectionTool extends TargetingTool {
 	private DragTracker dragTracker;
 	private LocationRequest hoverRequest;
 
-	private WeakReference cachedHandlePart;
+	private WeakReference<EditPart> cachedHandlePart;
 
 	/**
 	 * Default constructor.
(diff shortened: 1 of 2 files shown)


org.eclipse.gef/src/org/eclipse/gef/tools/SelectionTool.java
```diff
@@ -616,7 +613,7 @@ private void setLastHandleProvider(EditPart part) {
 		if (part == null)
 			cachedHandlePart = null;
 		else
-			cachedHandlePart = new WeakReference(part);
+			cachedHandlePart = new WeakReference<EditPart>(part);
```
azoitl, line 616: This type here is not needed and <> is sufficient

org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalViewerKeyHandler.java
```diff
@@ -527,7 +525,7 @@ private void setCachedNode(GraphicalEditPart node) {
 		if (node == null)
 			cachedNode = null;
 		else
-			cachedNode = new WeakReference(node);
+			cachedNode = new WeakReference<GraphicalEditPart>(node);
```
azoitl, line 528: This type information is not needed and the <> is sufficient.

azoitl (changes requested):

Phillipus:
Yep. Eclipse auto-magic doing its thing there. :-)

azoitl:
> Yep. Eclipse auto-magic doing its thing there. :-)

As good as they are most of the time, they can get out of hand sometimes. Thx for the patch.

azoitl (approved):

Issue #124, opened 2022-11-02
Maintain Canvas Position under Mouse during Zooming


azoitl:
Currently the ZoomManager tries to maintain the center of the canvas while zooming. This can lead to strange zooming behavior and users loosing track where they are. This happens especially in larger diagrams. 

In 4diac IDE we implemented a different behavior where the point under the mouse is preserved. So if your mouse is over an element and you zoom in or out this element is not moving.

We find that a very convenient behavior and therefore I'm consider to upstream it. However before doing so I would like to ask if this behavior is a better general behavior. 

@Phillipus, @pcdavid what do you think about it?

Phillipus:
Could it be optional behaviour?

azoitl:
Good point. As it is currently rather deeply woven into the zoom logic I haven't thought about it But yes it could be nicely separated and made configurable. For that I would propose an IZoomScrollPolicy interface which the ZoomManager would use to calculate the new scrollposition and apply it. The current implementation could be a default policy and be used unless something else is set. And I could then provide my new one rather easily.

azoitl:
@Phillipus with your inspiration I created a first PR #125 to show a potential solution. What to you think? Is this in the direction you imagined. I applied the new ZoomScroolPolicy to the logic editor for testing. What should be marked is while it is working it is not behaving as naturally as in 4diac IDE. The reason is there we have a border around the drawing canvas. This has to reasons: 1. it gives the user visual feedback where the diagram ends
2. for zooming and scrolling there is enough spare space that the content under the mouse always stays.

I'm considering to upstream that as well if people are interested.

Phillipus:
@azoitl Yes, that looks great! I think your idea of using a policy pattern fits in with the existing GEF design patterns.

azoitl:
> @azoitl Yes, that looks great! I think your idea of using a policy pattern fits in with the existing GEF design patterns.

Thx. It is also much nicer then what we currently have in 4diac IDE. However after sleeping it over I think we should really first decide on #100. Because I guess this feature is something that we shouldn't have duplicated in GEF and ZEST.

PR #125, opened 2022-11-02
Addresses #124: A first proposal for IZoomScrollPolicies

azoitl:
With a zoom scroll policy a client can adjust how the scroll position should be maintained during zooming. Two implementations are provided:
  - The default which is the original GEF Classic behavior (i.e. maintain the center of the diagram)
  - The mouse location based which tries to maintain the location under the mouse

This is a first POC for getting feedback. More documentation is needed.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/zoom/AbstractZoomManager.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/zoom/AbstractZoomManager.java
@@ -66,14 +66,35 @@ public abstract class AbstractZoomManager {
 	DecimalFormat format = new DecimalFormat("####%"); //$NON-NLS-1$
 
 	/**
+	 * The zoom scroll policy associated with this zoom manager. Zoom scroll
+	 * policies regulate how the position of the scroll bars are treated during
+	 * zooming.
+	 */
+	private IZoomScrollPolicy scrollPolicy;
+
+	/**
+	 * Creates a new ZoomManager.
+	 * 
+	 * @param pane         The ScalableFigure associated with this ZoomManager
+	 * @param viewport     The Viewport associated with this ZoomManager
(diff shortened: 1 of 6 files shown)


azoitl:
@Phillipus I think this PR is also now ready to be merged. Would be great if you could have a quick look and check if I haven't missed anything.

azoitl:
Sorry during more extended tests in 4diac IDE I noticed there is an offset in the mouse position calculation. Fixed in my latest commit. Now it should be really ready.

Phillipus:
> @Phillipus I think this PR is also now ready to be merged. Would be great if you could have a quick look and check if I haven't missed anything.

@azoitl Looks good to me. The changes are optional so existing code won't be affected.

BTW - what's the best practice if a PR contains multiple commits where one of the commits is a fix to the previous commits? I've been force pushing the PR rather than adding a new commit. WDYT?

vogella:
Force pushing is fine to update a PR.

azoitl:
@Phillipus thanks for noting this. In this case I really should have updated to original commit. Should I still do that? I can also squash the three commits for this PR

What I'm not sure if splitting a PR in different commits as I often do it's a good idea or not. Personally I like it as it splits bigger tasks in smaller chunks and keep these also when merging the PR. This has the advantage that if there are later issues identified it is easier narrow down the problem. But I'm not sure if this is a good practice.

Phillipus:
I think multiple commits for a PR makes sense if each commit represents a discrete change and makes the PR clearer in intent. But if it's a fix or change to one of the PR commits then I would go for a force push on the commit affected. But if the commit affected is not the last commit then you'd have to re-do the order of commits (cherry picking etc) and re-push.

azoitl:
Thx for the feedback. Based on it I squashed everything together I think for this PR it makes more sens.

Phillipus:
> Thx for the feedback. Based on it I squashed everything together I think for this PR it makes more sens.

Looks good to me. 👍

PR #126, opened 2022-11-03
Update Tycho extensions file to 3.0.0

Phillipus:
As Tycho 3.0.0 is used this should be set to the same version


azoitl:
Thx for finding that. I always forget about the .mvn directory. In my other project we don't have that.

PR #127, opened 2022-11-04
Fix drag operation unwanted artifacts

Phillipus:
Don't call canvas.update() in NativeGraphicsSource#getGraphics

This is needed to avoid SWT/GEF bug 137786 where drag over feedback (ie the drag feedback provided by the drag source) and drag under feedback (ie the feedback provided by the drop target) interfere with each other, causing flickering and more importantly ugly graphical artifacts.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=137786


azoitl:
After reading all the comments in Bug 137786 I feel a bit nervous about this PR. @Phillipus do you have any tips how I could test these in different setings?

Phillipus:
Well, we've had that patch for the past 10 years or so in Archi. Without it we get cheese artefacts when dragging over a node.  If I remember many other people also had to workaround it.

azoitl:
Ah that explains a lot. I would then take it.

Phillipus:
Thanks!

azoitl:
Thx for upstreaming!

Issue #106, opened 2022-08-21
Make ScaledGraphics optional


Phillipus:
Opening this issue as a placeholder for discussion about the `ScaledGraphics` class and making it optional, why it has problems, and what we can do about it.

More to follow...

Phillipus:
The [`ScaledGraphics`](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/ScaledGraphics.java) class is a wrapper around another `Graphics` instance and seems to add some additional features like Font scaling. But some fonts are clipped at different zoom levels.

We encountered this in our RCP app:

https://github.com/archimatetool/archi/issues/621

And other discussions:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=470334
https://bugs.eclipse.org/bugs/show_bug.cgi?id=442442

I don't know whether the underlying cause of the font problem could be fixed or whether `ScaledGraphics` should be removed or made optional. We made it optional:

https://github.com/archimatetool/archi/commit/98b038e346275da27146c7eb95c801f3ef71b6c6
https://github.com/archimatetool/archi/commit/2f364562c93a2d7913cfece92d0a1862ddd20702

Phillipus:
And another RCP tool, Modelio, made the same decision

> * Same as {@link ScalableFreeformLayeredPane} but don't use a ScaledGraphics
> * to paint the area, this class sucks at zooming texts.

https://github.com/ModelioOpenSource/Modelio/blob/9348609fab75011467b9f04f6965390e2471e5aa/modelio/app/app.diagram.elements/src/org/modelio/diagram/elements/core/figures/freeform/ScalableFreeformLayeredPane2.java

Phillipus:
One downside of not using ScaledGraphics:

https://github.com/archimatetool/archi/issues/851

azoitl:
In the 4diac IDE we also replaced the scaled graphics with the graphics [1]. Since then I have the impression the graphics looks nicer under Linux however we also noticed some problems:

- XOR drawing mode seems not to work under Windows. XOR drawing is used by GEF for selection higlight, marquee select border and D&D feedback
- On some zoom-factors the clipping of connections was to small and some connection segments (mostly vertical ones) where not drawn. We could compensate that with increasing the bounding margins. Also it seemed that mostly windows was affected.

As for both windows was mostly affected it could also be an SWT issue.

[...]

PR #132, opened 2022-11-24
Add option to not use Scaled Graphics

Phillipus:
Draft PR to further the discussion on https://github.com/eclipse/gef-classic/issues/106


Phillipus:
This PR is no longer needed, so closing.

azoitl:
Although we still have no solution for the printing.

PR #136, opened 2022-12-07
Update to Tycho 3.0.1

Phillipus:
- Contains an important fix
- See https://github.com/eclipse-tycho/tycho/issues/1814


Phillipus:
@azoitl Builds will fail until this is committed.

azoitl:
> @azoitl Builds will fail until this is committed.
thx for noting this.

Issue #128, opened 2022-11-09
Palette not colored correctly in dark theme


kjsmita6:
cc @vogella 

Eclipse GEF	GEF Classic Draw2d	3.14.0.202206170857	org.eclipse.draw2d
Eclipse GEF	GEF Classic GEF (MVC)	3.14.0.202206170857	org.eclipse.gef

Coming here from https://github.com/eclipse-platform/eclipse.platform.ui/issues/259. Our Eclipse RCP product uses a GEF-based editor and palette. The palette is not colored correctly when the platform is in dark theme:
![image](https://user-images.githubusercontent.com/9405706/200885821-39d3177d-7e96-4b01-b4fd-ab507134524e.png)
(notice the rest of the workbench is dark correctly)

The main editor canvas is dark correctly, it is just the palette view which does not get colored.

However, on rare occasions, I have gotten it to work. It seems that (sometimes) putting the OS in dark mode and also having Eclipse in dark mode can help, although doing this as I'm writing this issue isn't working.

Phillipus:
> when the platform is in dark theme

Is this on Windows or Mac?

Phillipus:
I have a custom version of GEF and Draw2d. This version can set the colors of the palette using a theme ColorDefinition set in a customised copy of Eclipse's dark theme. When I have more time I'll try to contribute this to GEF code here.

kjsmita6:
> Is this on Windows or Mac? Do you mean OS dark theme or Eclipse / RCP dark theme?

This is Mac, and I meant Eclipse dark theme.

> I have a custom version of GEF and Draw2d. This version can set the colors of the palette using a theme ColorDefinition set in a customised copy of Eclipse's dark theme. When I have more time I'll try to contribute this to GEF code here.

We have our own dark theme color overrides as well, including for the GEF editor. Does this not work for the palette then without your changes?
 
> I found, though, that a very dark palette background does not work for some toolbar icons. So it may be that you need another icon set for the palette in dark theme.

I have found the opposite for the times it has worked. On those rare occasions when the palette shows in dark mode, the icons and text are correctly matching the background.

Phillipus:
> I have found the opposite for the times it has worked. On those rare occasions when the palette shows in dark mode, the icons and text are correctly matching the background.

We have some black icons like arrows and so doesn't work for us. Something to bear in mind if the palette is changed when in dark mode.

azoitl:
[...]

Issue #134, opened 2022-12-02
GEF 3.16 Release for Eclipse 2023-06


azoitl:
This issue is used to track and coordinate the work for the GEF Classic 3.15.0 release to be contributed to the Eclipse 2022-12 release train. It shall serve as checklist as well as template for the upcoming release.

- [x] Update version number and target platforms:
     - [x] Update the version number of all features to 3.16.0
     - [x] Update the target platform for the API baseline to GEF Classic 3.15.0
     - [x] Update the pom file target platform checking to GEF Classic 3.15.0
     - [x] Update the development target platform to Eclipse 2022-12
     - [x]  Update the development target platform to Eclipse 2023-03
- [x] 12.04.2023: Release 3.16.0 M1 for 2023-03 M1 due on 14.04.2023
- [x] 03.05.2023: Release 3.16.0 M2 for 2023-03 M2 due on 05.05.2023
- [x] 24.05.2023: Release 3.16.0 M3 for 2023-03 M3 due on 26.05.2023
- [x] 31.05.2023: Release 3.16.0 RC1 for 2023-03 RC1 due on 02.06.2023
- [x] 07.06.2023: Release 3.16.0  for 2023-03 RC2 due on 09.06.2023
- [x] 14.06.2023: 2023-03 Release due on 14.06.2023
   - [x] tag release commit and mark it as release in github

As an experience from 2022-12 release, I set the release dates always two days before the release data so that Aggrcon can pick it up and it is included in the simrel release sites.

azoitl:
As I forgot it to note it in the check list: For all plugins at least the maintenance version number needs to be bumped for the start commit.

vogella:
@azoitl is this one done?

azoitl:
> @azoitl is this one done?

Unfortunately we didn't do much in the last 3 months. I wanted to move that issue as contribution for 2023-06.

vogella:
> 

Sounds good to me.

azoitl:
Just did 3.16.0M2: https://download.eclipse.org/tools/gef/classic/releases/3.16.0M2/

Aggcron update: https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/201702

azoitl:
3.16.0M3 is available at: https://download.eclipse.org/tools/gef/classic/releases/3.16.0M3/

Aggcron update: https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/202093

[...]

PR #140, opened 2022-12-22
Adresses #128 add color customization and icon overloads

Destrolaric:
Pr adds an ability to customize colors via `org.eclipse.ui.themes`. Colors could be configured via preferences or theme settings. Additionally adds an ability to override configured images to support more extensive customization. This improvement will allow for the palette and other components to be configured for a black theme.

### DIFF
--- a/org.eclipse.draw2d/.settings/.api_filters
+++ b/org.eclipse.draw2d/.settings/.api_filters
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<component id="org.eclipse.draw2d" version="2">
+    <resource path="src/org/eclipse/draw2d/ColorConstants.java" type="org.eclipse.draw2d.ColorConstants">
+        <filter comment="Adding fields to constant interface is fine" id="403767336">
+            <message_arguments>
+                <message_argument value="org.eclipse.draw2d.ColorConstants"/>
+                <message_argument value="lineForeground"/>
+            </message_arguments>
+        </filter>
+        <filter comment="Adding fields to constant interface is fine" id="403767336">
+            <message_arguments>
+                <message_argument value="org.eclipse.draw2d.ColorConstants"/>
+                <message_argument value="listHoverBackgroundColor"/>
+            </message_arguments>
+        </filter>
+        <filter comment="Adding fields to constant interface is fine" id="403767336">
+            <message_arguments>
+                <message_argument value="org.eclipse.draw2d.ColorConstants"/>
+                <message_argument value="listSelectedBackgroundColor"/>
+            </message_arguments>
+        </filter>
+    </resource>
+</component>
--- a/org.eclipse.draw2d/META-INF/MANIFEST.MF
+++ b/org.eclipse.draw2d/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
-Bundle-SymbolicName: org.eclipse.draw2d
+Bundle-SymbolicName: org.eclipse.draw2d;singleton:=true
 Bundle-Version: 3.12.100.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
@@ -15,7 +15,8 @@ Export-Package: org.eclipse.draw2d,
  org.eclipse.draw2d.widgets,
  org.eclipse.draw2d.zoom
 Import-Package: com.ibm.icu.text
-Require-Bundle: org.eclipse.swt;visibility:=reexport;bundle-version="[3.4.0,4.0.0)"
+Require-Bundle: org.eclipse.swt;bundle-version="[3.4.0,4.0.0)";visibility:=reexport,
+ org.eclipse.ui;bundle-version="3.201.100"
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Automatic-Module-Name: org.eclipse.draw2d
--- a/org.eclipse.draw2d/META-INF/MANIFEST.MF
+++ b/org.eclipse.draw2d/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.draw2d;singleton:=true
-Bundle-Version: 3.12.100.qualifier
+Bundle-Version: 3.13.0.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.draw2d,
--- a/org.eclipse.draw2d/build.properties
+++ b/org.eclipse.draw2d/build.properties
@@ -12,7 +12,8 @@ bin.includes = about.*,\
                META-INF/,\
                .,\
                plugin.properties,\
-               gef_eclipse_logo_32.png
+               gef_eclipse_logo_32.png,\
+               plugin.xml
 source.. = src/
 src.includes = src/org/eclipse/draw2d/images/,\
                about.html
--- a/org.eclipse.draw2d/css/gef_dark.css
+++ b/org.eclipse.draw2d/css/gef_dark.css
@@ -1,0 +1,12 @@
+IEclipsePreferences#org-eclipse-ui-workbench:org-eclipse-draw2d {
+	preferences:
+		'org.eclipse.draw2d.color.line.foreground=64,64,64'
+		'org.eclipse.draw2d.color.list.selected.background=90,90,124'
+		'org.eclipse.draw2d.color.list.hover.background=80,80,100'
+		'org.eclipse.draw2d.color.list.background=57,66,72'
+		'org.eclipse.draw2d.color.list.foreground=224,224,224'
+		'org.eclipse.draw2d.color.menu.background=57,66,72'
+		'org.eclipse.draw2d.color.menu.foreground=224,224,224'
+		'org.eclipse.draw2d.color.menu.foreground.selected=105,105,105'
+		'org.eclipse.draw2d.color.button=64,64,64'
+}
--- a/org.eclipse.draw2d/css/gef_dark.css
+++ b/org.eclipse.draw2d/css/gef_dark.css
@@ -1,11 +1,11 @@
 IEclipsePreferences#org-eclipse-ui-workbench:org-eclipse-draw2d {
 	preferences:
 		'org.eclipse.draw2d.color.line.foreground=64,64,64'
-		'org.eclipse.draw2d.color.list.selected.background=90,90,124'
-		'org.eclipse.draw2d.color.list.hover.background=80,80,100'
-		'org.eclipse.draw2d.color.list.background=57,66,72'
+		'org.eclipse.draw2d.color.list.selected.background=90,90,90'
+		'org.eclipse.draw2d.color.list.hover.background=70,70,70'
+		'org.eclipse.draw2d.color.list.background=60,60,60'
 		'org.eclipse.draw2d.color.list.foreground=224,224,224'
-		'org.eclipse.draw2d.color.menu.background=57,66,72'
+		'org.eclipse.draw2d.color.menu.background=40,40,40'
 		'org.eclipse.draw2d.color.menu.foreground=224,224,224'
 		'org.eclipse.draw2d.color.menu.foreground.selected=105,105,105'
 		'org.eclipse.draw2d.color.button=64,64,64'
--- a/org.eclipse.draw2d/plugin.properties
+++ b/org.eclipse.draw2d/plugin.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2007, 2009 IBM Corporation and others.
+# Copyright (c) 2007, 2022 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License v1.0
 # which accompanies this distribution, and is available at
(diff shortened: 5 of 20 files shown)


azoitl:
@Destrolaric thx for the patch. I'm no export in the themeing thing but I noticed a few things that I kindly ask to consider:

1.  You need to sign the Eclispe contributor agreement and link your github account with your Eclipse account. 

2. For all the files you did bigger changes you need to update the copyright header (ie., 200x, 2022). For files you created you need to attribute them to the copyright holder, I think in your case dbeaver. 

3. I'm normally a big fan of several smaller commits but in this case and as there are some changes required I think it is better to sqash the commits.

Destrolaric:
@azoitl Hi, thank you. I signed the eca. And by the check status, it looks like everything is good. Thank you for the info about the copyright header. It will be improved.

Destrolaric:
@azoitl Updated the copyright and squashed the first two commits.

azoitl:
Today I finally found time to experiment a bit with your patch. There are a few smaller backward compatibility issues which should be rather straight forward to address. However the bigger issue I have is that several colors are now off. These include the  background color of the editor and palette as well as the grid. Furthermore the palette now has a bad contrast between folder items and folder content and the selection color is black and not the selection color used in the rest of Eclipse.

Finally switching to the dark theme is only partly updating the colors. Even after restart.

I tested with the GEF example applications and with 4diac IDE.

vogella:
> However the bigger issue I have is that several colors are now off. 

@azoitl can you suggest new default values for the colors? Using the preferences you should be able to set the colors relatively easy.

> Finally switching to the dark theme is only partly updating the colors. Even after restart.

@Destrolaric could you reproduce this issue?

Destrolaric:
@vogella Hi, sorry for waiting. Yes, several default colors are off. It will be fixed. I also found a way to bind colors to dark theme automatically. I am currently working on the implementation. I will write back as soon as it is ready.

vogella:
@Destrolaric thanks for the answer. Great to hear that you are already working on an improvement.

An easy way of binding colors to the dark theme is to use a CSS extension. For example JDT defines this extension for the dark theme: https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/plugin.xml#L6971 This file allows you to set the preference values https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/css/e4-dark_jdt_syntaxhighlighting.css which basically sets the preferences based on the theme.

A nice tool for working with preference key / values is the preference spy which you can open in a recent IDE with Ctrl+3 -> Preference spy.

Toggle the trace to see all preference changes 

image

The option to exchange icons based on the theme would be also super useful for the IDE / JFace. So once this has landed in GEF, we may want to move / add this also for JFace. Would be awesome to have themes specific icons in the IDE.

Destrolaric:
@vogella Hi, sorry for waiting, I have added default colors for dark theme.

Destrolaric:
> An easy way of binding colors to the dark theme is to use a CSS extension. For example JDT defines this extension for the dark theme: https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/plugin.xml#L6971 This file allows you to set the preference values https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/css/e4-dark_jdt_syntaxhighlighting.css which basically sets the preferences based on the theme.

Thank you for the advice. Added a new CSS file with the configuration.

vogella:
@azoitl can you retest?

azoitl:
Yes will do. Was traveling this week therefore didn't participate in the discussions.

azoitl:
I did some test today. It definitely looks better. Especially in the dark mode. However I noticed a few things and I have some general issue with the patch which I’m not sure how to address.

First of all the patch breaks the background color of the editing area and the palette (see screenshot below). This was before white and should IMHO the editor background color. Especially in the palette we loose with that the difference between palette drawer and palette entry. In dark mode it is a bit better.

The next element is that the selection color in the palette is now different then in the rest of the platform (see screenshot below). This is even worse in darkmode.

This brings me also to my biggest concern. With the latest changes but also a bit before we are introducing a full own color set for GEF Classic. So if the platform updates we will always diverge. I find this a bad. With your infrastructure in place we should be able to hook on existing colors of the Eclipse platform and use them appropriately, am I right? 

@vogalla what would be from the Eclipse Platform perspective? How can we get the right color definition. I know for example that the editor background color is found at a very strange place. In 4diac IDE I had to use the following piece of code:

```
final String background = InstanceScope.INSTANCE.getNode("org.eclipse.ui.editors") //$NON-NLS-1$
				.get("AbstractTextEditor.Color.Background", null); //$NON-NLS-1$

if (background != null) {
	// we have a color in the preferences set
	final RGB rgb = StringConverter.asRGB(background, null);
	return new Color(rgb);
}

// if not in the preferences try to get it from the current theme
final ColorRegistry colorRegistry = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme().getColorRegistry();
return colorRegistry.get("org.eclipse.ui.editors.backgroundColor"); //$NON-NLS-1$
```

the two different approaches are needed because the color is differently provided in light and dark mode. But I may messed up here.


Screenshot from 2023-01-24 16-05-43

kjsmita6:
@azoitl I've noticed that `PlatformUI.getWorkbench().getThemeManager().getCurrentTheme().getColorRegistry().get("MY_COLOR_ID")` doesn't always get the color correctly. Like if we are in dark theme it might get the color in light theme. We use the following code as the first resort when finding colors programmatically:
```java
        String themeId = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme().getId();
        IThemeDescriptor themeDesc = WorkbenchPlugin.getDefault().getThemeRegistry().findTheme(themeId);
        if (themeDesc != null) {
            // Find the theme from the descriptor. This tends to be the most accurate.
            Optional<ColorDefinition> colorDef = Arrays.asList(themeDesc.getColors()).stream().filter(def -> "MY_COLOR_ID".equals(def.getId())).findFirst();
            if (colorDef.isPresent()) {
                return new Color(colorDef.get().getValue());
            }
        }
```
This method seems to always work, although we are using `IThemeDescriptor` which is internal.

I've never accessed preferences to get colors though so maybe that could work instead.

azoitl:
@kjsmita6 thx for pointing that out. Not sure if we should use Eclipse internal API. But what bothers me most is where do I get a list of color ids. Do you use your own or the one from the platform. If the later how did you find them?

azoitl:
Ah and @kjsmita6 my approach with the preferences has the big drawback that I need to know the ID of the color defining plugin otherwise this
`InstanceScope.INSTANCE.getNode("org.eclipse.ui.editors")`
can not be used. So your approach sound from that perspective better

kjsmita6:
> @kjsmita6 thx for pointing that out. Not sure if we should use Eclipse internal API. But what bothers me most is where do I get a list of color ids. Do you use your own or the one from the platform. If the later how did you find them?

We have both. Some colors we have defined in plugin.xml (via `colorDefinition` or `colorOverride` extension point) as RGB (like ID `com.example.my_blue` has a `value` of `"100,100,100"`), but some we have the `value` set to platform colors (like ID `com.example.my_red` has a `value` of `COLOR_RED`).

The system colors are found in `org.eclipse.swt.SWT` as `COLOR_*` constants. The name of the constant is the ID. I'm not sure where the RGB of these colors is defined. Like most things SWT it is probably platform-specific. I assume other plugins, like `org.eclipse.ui`, would define their own colors in their own plugin.xmls.

Destrolaric:
@azoitl I have looked into the commentaries regarding adding additional colors, most of them have a binding to system colors by default, so they will not deviate from the eclipse color palette. The dark theme overrides some colors. This is due to two possible reasons: 
1. They previously had a hardcoded color value like `ColorConstants.white,` so they have some specific RGB color value 
2. They previously used system values that are not changed by theme change, so we need to define them in the CSS.
There can be possible color differences, but they could be fixed by setting the proper ones. 

I am not sure I understand the proposed solutions for extracting color definitions. Are there any edge cases when `themeManager.getCurrentTheme().getColorRegistry().get()` could not work out?.

kjsmita6:
@Destrolaric 
> I am not sure I understand the proposed solutions for extracting color definitions. Are there any edge cases when `themeManager.getCurrentTheme().getColorRegistry().get()` could not work out?.

"Edge cases" I'm not sure. In my case, that code wouldn't get the colorOverride for dark theme and instead get the default one (for light theme). For example, define a color ID `CANVAS_BACKGROUND_COLOR` set to `COLOR_WHITE` for light theme and `COLOR_BLACK` for dark theme. That `getCurrentTheme().getColorRegistry().get("CANVAS_BACKGROUND_COLOR")` would return `COLOR_WHITE` regardless of the theme. Always the chance we are doing something wrong though ;) we also aren't an e4 app thus not using CSS styling.

azoitl:
I totally agree that hard coded colors have to go. I had the same experience as @kjsmita6 that `themeManager.getCurrentTheme().getColorRegistry().get()`didn't got us the right color in all cases. That's the reason I posted the code above. But for now I have no clear solution on how we go forward here.
I'm maybe also over-complicating things. But as my screenshot shows above the editor looks completely different in the light theme then before. So I think we have to do something more. 

But I also don't think we have to solve everything at once.

Destrolaric:
Hi @azoitl, thanks for the clarification. I agree that the case is not that simple, but perhaps we could try to look at these problems individually and find a constructive solution.

- The latest commit should fix the wrong colors on the white theme. Color Registry shouldn't be the cause for that. Some mismatched colors may be left, but they should be easily fixable.
- As for the dark colors, the dark theme sets colors by RGB values, which can also be corrected if needed.
- Regarding the issue with adding additional color endpoints, for most cases, default values are the same as previous ones.
- I am not quite sure what to do about the `ColorRegistry` situation, but it would be great if you have any suggestions.

If any acceptance criteria can be formulated for this pr, that would also be great. I'm ready to make improvements, but I'd like to understand what's worth doing.

azoitl:
@Destrolaric sorry for the late reply. I was traveling and got distracted. First of all again thx for your support here. I kindly ask to be a patient with me. This is the first time I have to handle such a PR in GEF Cl, and I'm learning on the job.

To answer your question. I think as soon as we have roughly the same (or improved) appearance to existing GEF Classic I would say the main work is done. What I also noticed that some API clean-up will be necessary. For that you need to set the API baseline in your Eclipse dev workspace. I setup API base-line target platform. The API base line is needed to get feedback which versions we need to bump and where we need to reconsider API. One thing that we are currently not able is to bump the major version. So for some of your changes we need to find a workaround (e.g., the new colors need to go into a new class, unfortunately to orginal design was not the best idea, but we have to live with it).  After that I think we are ready.

What I can offer is that I'll try to do some of the API fixes. As I have some experience with the ZoomManager work I did in the last release.  

What do you think?

Destrolaric:
@azoitl Thanks for reply. I glad that this PR is starting to move forward. Regarding colors situation I will look which colors are off and will fix it. Also, it will be great to get help for the API to be inline with other Eclipse parts. API baseline will be helpful. I didn't use it before, so I need time to look into it.

Destrolaric:
@azoitl Hi, fine tuned the colors. Now both themes should look properly colored. 
javaw_j3WIVAOh4l
javaw_n0jEbxz7nN

azoitl:
@Destrolaric this looks really nice. Have you also tested the grid? But if you want I can do that later this week. I also don't see it super important. BTW I really like how the palette drawers are looking with your changes. Fits much better to the current design of the Eclipse platform. 

I would say we go with that and fine tune over in other commits. 

So the final steps would be to get the API correct. A help for that are the API tools of PDE. For that you need to setup the API baseline in your Eclipse workspace. In the preferences under API Baselines you need to add a new one and set it as active. In the GEF Classic repo there is a baseline directory there you can find baselines for the different versions since GEF classic 3.14.0 I provide a target platform file as baseline. Just choose it and set it as active. This is how it looks for in my GEF Classic workspace:
Screenshot from 2023-02-28 21-43-31

This will give you additional errors and warnings that we have to consider. Currently with GEF Classic we need to stay API compatible.  That means for any error that requires to bumb the major version we need a work around. 

Would you like to have a look first and then discuss what we should do, or should I have a look?

Destrolaric:
@azoitl Got following errors
image

vogella:
> @azoitl Got following errors image

Sorry for being to unavailable, I'm currently occupied with other things. If you have PDE installed you should have quickfixes available.

Adding fields to  an constant interface should be backward compabile, so I think you can just add an API filter for this via the quickfix.

The @since tags can also be added via quickfixes.

The removal of the field might be a problem if someone uses it. Not sure if that field is used somewhere, if in doubt you could leave it and deprecated it and / or make it a no-opt.

vogella:
> @azoitl Hi, fine tuned the colors. Now both themes should look properly colored. javaw_j3WIVAOh4l javaw_n0jEbxz7nN

Is GEF still using GIFs? Typical GIFs look really bad in the dark theme and PNG files should be preferred. @Destrolaric if case you have better looking images and the time and desire to contribute these to GEF, please open a new PR to update them.

Destrolaric:
@vogella We have a better palette icon, but there are two things, it uses blue color, which may look not uniform with other parts of Eclipse and it uses Apache License.
palette@2x

vogella:
> 



> @vogella We have a better palette icon, but there are two things, it uses the standard DBeaver blue color, which may not look uniform with other parts of Eclipse, and it is under Apache License. palette@2x

Looks better to me, IIRC Apache Licence is fine. @azoitl WDYT? I think the overall icon set of Eclipse looks in general a bit outdated and I personally would be happy to see new icons in GEF.

Destrolaric:
@vogella Is it ok to bump minor version?

Destrolaric:
I fixed the base-line check errors. @vogella, could you take a look?

vogella:
> I fixed the base-line check errors. @vogella, could you take a look?

Lets wait and see that @azoitl thinks, he is much deeper into this PR and of course also in GEF.

Destrolaric:
`Error: [ERROR] Internal error: java.lang.NullPointerException: Cannot invoke "org.apache.maven.repository.internal.ModelCacheFactory.createCache(org.eclipse.aether.RepositorySystemSession)" because "this.modelCacheFactory" is null -> [Help 1]` 
Something wrong with maven build

Destrolaric:
@azoitl Hi, I would like to hear your thoughts about the current state of the pull request. There are some issues with maven that look like CI/CD issues. Can you take a look?

azoitl:
@Destrolaric sorry for my late response I was occupied with other stuff the last week. I'll retriggered the build job and hope this fixes it. Last week there was some issue with Eclipse build infra. 

Regarding the icon @Destrolaric & @vogella  I would like to disucss this in the dedicated issue (#144) that was opend for this.

azoitl:
@Destrolaric the maven build passes now. However when I check it out my Eclipse PDE I still get three errors regarding the version problem:  
Screenshot from 2023-03-12 21-13-00

How I see the problem is that you added three new fields to the ColorConstants interface. This is an API breakage. For me the only solution would be to move these three values to a new final class implementing the ColorConstants interface and add there the three new color constants. We could mark the ColorConstants then deprecated and use the new class everywhere. 

Or do you have a better idea?

Destrolaric:
@azoitl Hmm, weird. For my baseline configuration, a minor version change led to that error going away. Can we increment the minor version? These are fields so they should be backward compatible in most cases.

vogella:
> @azoitl Hmm, weird. For my baseline configuration, a minor version change led to that error going away. Can we increment the minor version? These are fields so they should be backward compatible in most cases.

Yes, increasing the minor version is fine.

azoitl:
> > @azoitl Hmm, weird. For my baseline configuration, a minor version change led to that error going away. Can we increment the minor version? These are fields so they should be backward compatible in most cases.
> 
> Yes, increasing the minor version is fine.

@vogella @Destrolaric already changed the minor version for his patch. Therefore I'm currently not sure what to do. Now we are getting to the end of my java knowledge but is it possible to mark this interface to be not implemented by clients. Because this is exactly what should be done. As said above it was from the beginning a bad idea to collect these constants in an interface. But this was done 20 years ago and we have to live with it.

vogella:
> > > @azoitl Hmm, weird. For my baseline configuration, a minor version change led to that error going away. Can we increment the minor version? These are fields so they should be backward compatible in most cases.
> > 
> > 
> > Yes, increasing the minor version is fine.
> 
> @vogella @Destrolaric already changed the minor version for his patch. Therefore I'm currently not sure what to do. Now we are getting to the end of my java knowledge but is it possible to mark this interface to be not implemented by clients. Because this is exactly what should be done. As said above it was from the beginning a bad idea to collect these constants in an interface. But this was done 20 years ago and we have to live with it.

I added an API filter and changes the Javadoc to the new version. API wise it should be fine now to merge. @Destrolaric could you rebase onto master?

vogella:
> > I added an API filter and changes the Javadoc to the new version. API wise it should be fine now to merge. @Destrolaric could you rebase onto master?
> 
> Thx a lot @vogella was not aware about that. In the next weeks I should again have more time for GEF Classic and be more responsive.

Sorry, the usage of API filters is very common in platform so I incorrectly assumed that my comment in https://github.com/eclipse/gef-classic/pull/140#issuecomment-1450148507 was clear enough.

Destrolaric:
@vogella I made a rebase

vogella:
> > Sorry, the usage of API filters is very common in platform so I incorrectly assumed that my comment in [#140 (comment)](https://github.com/eclipse/gef-classic/pull/140#issuecomment-1450148507) was clear enough.
> 
> I read it but didn't full get the implication. Also given the history with some of my commits I was a bit nervous about such a change.

Understandable. So are we ready to merge this one?

azoitl:
I just tested it for me works perfectly fine. 
@Destrolaric sorry that it took so long for getting this PR merged. Next time it will be faster. There is still so much I have to learn.

vogella:
Thanks a bunch @Destrolaric for your work. Would be nice to see a few more contributions from you so that we can nominate you as GEF committer so that you can work directly on the code base.

vogella:
Thanks @azoitl for doing the testing and code review. Again sorry for my only little involvement.

Destrolaric:
@vogella We are planning to transfer generic support from DBeaver fork in the near future. So there should be more prs soon.

Destrolaric:
@vogella, Is there any estimation of when will be a release containing these changes?

azoitl:
@Destrolaric these are great news. 

Regarding your question. I would try to hop on the simultanouse release train again. so the next official GEF release would then be with Eclipse 2023-06 which is scheduled for June 14th. I will update issue #134 accordingly later today.

But we have a latest build update page you can use for testing in DBeaver if you want to.

azoitl:
Oh I missed that. Yes that may make people unhappy. @Destrolaric is there a possibility to move your new color provider to GEF and use the old colors as fallback for a non eclipse usage of Draw2D.

kjsmita6:
Are there any specific steps required to use the new functionality? The original bug (the white palette on black background - why #128 was opened) is still there with the latest snapshot. I would've expected it to "just work".

azoitl:
> @azoitl I think it could be possible. I will see how it can be done.

@Destrolaric this would be amazing.

kjsmita6:
I am saying even with dark mode and restarts and everything needed my
palette is still white as before.

I don't have it in my environment anymore but let me know if there is
anything I can do/show to provide more info and I will add it back.

PR #149, opened 2023-03-26
[Modernice Code] Typify IFigure/Figure's Children

azoitl:
In my attempt to modernize GEF classic code I worked on the IFigure and more important the Figures infrastructure by typifying the children list in Figure and change the return value of getChildren to List<? extends IFigure>. With that we can get rid of many casts and instanceof checks. Also Client code can take advantage of it. In order to test it I compiled GEF classic and all its examples as well as with Eclipse GMF runtime and Eclipse 4diac. To ensure that I haven't broken any core features (e.g., selection) I ran the GEF classic logic editor and Eclipse 4diac. 

Finally to show the advantages I completely reworked the Draw2d Tree example. 

I hope with that caution it should be a save PR.  

\cc @Phillipus  @pcdavid  @Destrolaric


azoitl:
> @azoitl Hmm, maybe we can split the parts each of us could refactor. I think it may be viable in that case.

How about creating an issue where we can discuss that? This could also help us to document solutions and best practices and how to evaluate changes.

For now I would be most grateful for a fix to the Draw2d eclipse platform dependency problem as this is breaking our tests and breaking for potential users.

Destrolaric:
@azoitl Yes, it will be a good idea to create an issue where it could be discussed. I will try to work out the fix in a short time.

Issue #150, opened 2023-03-26
Unit Tests are failing


Phillipus:
I just noticed that even though the build action completes successfully, there are many failing tests. For example, here's a recent build:

https://github.com/eclipse/gef-classic/actions/runs/4518703236/jobs/7958746089

The raw log on that shows several failures.

azoitl:
Interesting. I see here two issues.

1.  Failing tests should be reported. Maybe @vogella when he set-up the build infra disabled the reporting as we have some tests that are constantly failng
2. The current log shows that the tests are not executed as eclipse plugin tests. AFAIK this is the reason for the `java.lang.NoClassDefFoundError: Could not initialize class org.eclipse.draw2d.ColorConstants` errors. Because when I run the tests locally on my machine these are not showing.

But I'm not sure what we should do with that.

Phillipus:
> Because when I run the tests locally on my machine these are not showing.

Are you running the tests locally using Maven or in Eclipse's JUnit runner? If the latter, the `Display` class might get initialised correctly but not when running from Maven.

azoitl:
Locally in Eclipse. Not yet in Maven. I thought there is something for also running with maven correctly. But I'm very bad in that thing.

Phillipus:
> Locally in Eclipse.

I came across this problem before with Archi. Tests would run fine in Eclipse but not in Maven. I found that Eclipse was initialising`Display#getCurrent` but Maven wasn't.

If we look at this commit we can see that `ColorConstants` used to initialise the Display:

https://github.com/eclipse/gef-classic/commit/93042dca3e62ed4f0ff0aedbd58e58327d98fc12

```java
class SystemColorFactory {
		private static Color getColor(final int which) {
			Display display = Display.getCurrent();
			if (display != null)
				return display.getSystemColor(which);
			display = Display.getDefault();
			final Color result[] = new Color[1];
			display.syncExec(new Runnable() {
				public void run() {
					synchronized (result) {
						result[0] = Display.getCurrent().getSystemColor(which);
					}
				}
			});
			synchronized (result) {
				return result[0];
			}
		}
	}
```

But that initialisation has been removed. Perhaps this is the case here?

azoitl:
This code has been re-added as it was API. However Draw2D shouldn't use it anymore. Hmh. Have to dig deeper.

Phillipus:
> This code has been re-added as it was API. 

Ah, OK, I see that in a later build. Perhaps then it is not called?

[...]

PR #151, opened 2023-03-27
Improve draw2d test setup

azoitl:


azoitl:
As discussed in #150 this PR shadows the real problem. Therefore I'm closing it without merging.

PR #153, opened 2023-03-27
GH verification build on Linux and Windows and multiple Java versions

vogella:
Currently we use Java 11 and Java 17, if GEF moves to Java 17 we may want to use latest and greatest in addition to Java 17.

PR #156, opened 2023-03-30
Use correct path to javadoc.exe on Windows

Phillipus:
Should fix problem building Javadoc on Windows


vogella:
Thanks @Phillipus, we can see with a rebase of https://github.com/eclipse/gef-classic/pull/153

PR #158, opened 2023-03-30
Add macOS to build matrix and use coactions/setup-xvfb

Phillipus:
- Build on macOS
- GabrielBB/xvfb-action is deprecated


Phillipus:
@vogella One thing not working is the `${{ matrix.config.native }}` part in this line:

```
 name: test-results-${{ matrix.config.native }}-java${{ matrix.java }}
```

Phillipus:
Could use this `name: test-results-${{ matrix.os }}-java${{ matrix.java }}`

See https://github.com/eclipse/windowbuilder/pull/363

azoitl:
> Adding macOS to the matrix might slow down these builds. Currently, the build is queued with the message:
> 
> ```
> The agent pool assigned to this job has hit their MacOs concurrency limits
> Requested labels: macos-latest
> ```
> 
> According to [this](https://stackoverflow.com/questions/73141152/macos-github-actions-queued-the-agent-pool-assigned-to-this-job-has-hit-their), it may be that we are limited on macOS concurrency builds.

According to some mail I read today or yesterday on an eclipse mailing list the agent pool and its quota is shared across an organization. The GEF Classic repo is currently in the Eclipse foundation org. EF is no starting to set-up an own organization for each project. Maybe that is one of the reasons. Not sure if it makes for now sense for GEF Classic. But maybe something to be considered.

Phillipus:
This is working now on all platforms.

Phillipus:
> According to some mail I read today or yesterday on an eclipse mailing list the agent pool and its quota is shared across an organization. The GEF Classic repo is currently in the Eclipse foundation org. EF is no starting to set-up an own organization for each project. Maybe that is one of the reasons. Not sure if it makes for now sense for GEF Classic. But maybe something to be considered.

I found that discussion [here](https://www.eclipse.org/lists/eclipse-dev/msg12141.html). It seems the GH EF organization is on the free plan which limits actions to 20 concurrent jobs. There are currently 464 repositories in the EF organization (granted not all use workflows). There's going to be quite some queuing if more come one board.

PR #159, opened 2023-04-03
[WB] Integrate geometry extensions from the WindowBuilder project.

ptziegler:
The Draw2D fork used by the WindowBuilder project includes several convenience methods. Because we want to migrate to the official GEF release, those changes need to be merged back.

In addition to several convenience methods, a new Interval class is introduced, describing the interval within 1-dimensional space. This class is used by the Rectangle, in order to determine the width or the height of a widget, depending on its orientation.

Test cases have been extended to cover this new functionality.


ptziegler:
Those 2000 lines of code added might look intimidating, but that's 95% the new JUnit tests :)

azoitl:
Thanks a lot for providing these improvements. This is amazing. 

First of all we currently have an issue with the unit tests of Draw2D. I would like to have this fixed before we are merging any new things to Draw2d. 

Yes I think your proposal of splitting this PR into individual PRs for each Draw2D class would be helpful. I had a quick look at the changes and for some I think some discussion on function naming would be needed. For me the most critical changes are the changes for Rectangle as you are changing the behavior of existing methods. Not sure if we should do this or give the whole set of functions a distinct name.

Also you need to add the @since tag to make PDE tooling happy.

ptziegler:
Thanks for the quick response!

> First of all we currently have an issue with the unit tests of Draw2D. I would like to have this fixed before we are merging any new things to Draw2d.

You're talking about #150, I assume? I'm currently in the process of getting the WB tests up & running again. Perhaps I can also take a look at what's going on with the GEF jobs.

In any case, I'll hold off on creating new pull requests, until those tests are green again. I'll keep this pull-request open as a placeholder, to document which changes have already been merged and which are still pending. :)

> For me the most critical changes are the changes for Rectangle as you are changing the behavior of existing methods. Not sure if we should do this or give the whole set of functions a distinct name.

That's a good point, and something that was bothering me as well. WB calls `setX()` and `setY()` only in a single class. So it's probably a better idea to refactor the WB code, rather than to introduce new side-effects in GEF.

That aside, I would even go so far and claim that we don't need to add additional methods and instead use `moveX()` and `moveY()`.

azoitl:
> You're talking about https://github.com/eclipse/gef-classic/issues/150, I assume? I'm currently in the process of getting the WB tests up & running again. Perhaps I can also take a look at what's going on with the GEF jobs.

Unfortunately with the recent work on getting the correct theme colors we accidentally a dependency to eclipse.ui to Draw2d. This needs to be fixed. @Destrolaric is working on it. 

> That's a good point, and something that was bothering me as well. WB calls setX() and setY() only in a single class. So it's probably a better idea to refactor the WB code, rather than to introduce new side-effects in GEF.
>
> That aside, I would even go so far and claim that we don't need to add additional methods and instead use moveX() and moveY().

I'm definitely the wrong one  to comment on your code base. But I'm happy if our collaboration here gets both Draw2D and WB better. And I found the functionality of the methods interesting. Maybe a name that better conveys the intention can help.

azoitl:
@ptziegler thanks for upstreaming these improvements. Hope that your GEF Classic migration will go smoothly.

PR #160, opened 2023-04-03
Add a new job to the GitHub action for publishing the test results.

ptziegler:
With this change, test failures are no longer silently ignored. When a failure has been detected, all remaining test cases continue to be executed, then the Maven build throws an error.

Regardless of the successful build, the test results for each job are uploaded to GitHub.

Once the job has completed, a new workflow action is executed, which aggregates all test results and publishes them within the GitHub action.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -24,8 +24,9 @@ jobs:
       uses: coactions/setup-xvfb@v1
       with:
        run: >- 
-        mvn -V -B -D maven.test.failure.ignore=true clean verify
+        mvn -V -B -fae clean verify
     - name: Upload Test Results for Java-${{ matrix.java }}
+      if: always()
       uses: actions/upload-artifact@v3
       with:
         name: test-results-${{ matrix.os }}-java${{ matrix.java }}
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,37 @@
+name: Publish Unit Test Results
+
+on:
+  workflow_run:
+    workflows: ["GEF Class verification build"]
+    types:
+      - completed
+
+jobs:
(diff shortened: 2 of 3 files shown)


ptziegler:
In short, this should deal with the issue where the builds are marked green, despite test failures.

ptziegler:
Looks like I missed the `maven.test.failure.ignore` flag in the pom.xml.

In any case,it looks like the `Publish Unit Test Results` is only performed, if the script is already on the master.  Though I've tested it on my fork and in principle, everything should work as expected.
image

azoitl:
As the tests are now reported correctly as failing, I'm merging this PR.

azoitl:
@ptziegler I tried to bring this setup to our new 4diac IDE repository. However there I always get a write exception. Do you have any tip would could be wrong?

https://github.com/eclipse-4diac/4diac-ide/actions/runs/8616565434

ptziegler:
Are you perhaps using Otterdog there? If so, write access is disabled by default and needs to be explicitly requested.I can say more once I'm back home.On Apr 9, 2024 15:33, azoitl ***@***.***> wrote:
@ptziegler I tried to bring this setup to our new 4diac IDE repository. However there I always get a write exception. Do you have any tip would could be wrong?
https://github.com/eclipse-4diac/4diac-ide/actions/runs/8616565434

ptziegler:
That really looks like the permissions being disabled by default, due to Otterdog. It's a safety mechanism to prevent malicious tools from gaining implicit access to the repository, from what I remember.

Given that this is related to unit tests, I _think_ you have to enable permissions like this:
```
jobs:
  unit-test-results:
    permissions:
      pull-requests: write
```

azoitl:
Thx for the tip. I took the liberty and look at the otterdog file of WindowBuilder.

PR #161, opened 2023-04-04
Disable fail-fast flag to avoid premature job cancellation.

ptziegler:
By default, the failure of a single job in the build matrix triggers the cancellation of all other jobs. This means we may miss out on test cases, because the other jobs might not have finished yet.

From the workflow documentation:

You can control how job failures are handled with
jobs.<job_id>.strategy.fail-fast and jobs.<job_id>.continue-on-error.

jobs.<job_id>.strategy.fail-fast applies to the entire matrix. If jobs.<job_id>.strategy.fail-fast is set to true, GitHub will cancel all in-progress and queued jobs in the matrix if any job in the matrix fails. This property defaults to true.

jobs.<job_id>.continue-on-error applies to a single job. If jobs.<job_id>.continue-on-error is true, other jobs in the matrix will continue running even if the job with jobs.<job_id>.continue-on-error: true fails.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         java: [ 17 ]


ptziegler:
Because I noticed that thje MacOS job got cancelled by the test failures on the Windows/Linux jobs, before its tests finished. I don't think that's desirable.

azoitl:
@ptziegler do you know why macos got again cancled?

ptziegler:
My best guess there's a 6h time limit, in order to prevent jobs from being stuck indefinitely.

ptziegler:
> OS macos-latest Java 17 compile
> The job running on runner GitHub Actions 5 has exceeded the maximum execution time of 360 minutes.

PR #164, opened 2023-04-07
Stabilized Draw2d Tests

azoitl:
ConnectionEndPointMove test did correctly wait for all updates to be dispatched. This resulted in different test results depending if executed alone or as part of Draw2d's test suite. This is fixed now.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ConnectionEndPointMoveTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ConnectionEndPointMoveTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


azoitl:
After some poking around I think I got the not so stable Draw2d tests into a stage that they are giving us reasonable feedback. or that I had to deactivate some tests: only flowpages and their breaking up  of text. 

What do you think @Phillipus @Destrolaric.

@ptziegler with this PR we would be ready for your improvements of Draw2d.

Phillipus:
> What do you think @Phillipus @Destrolaric.

I think you did a good job to stabilize this. :-)

Moving forward we need a stable test suite that passes so we can test against new commits. I think we're at that point now.

azoitl:
> > What do you think @Phillipus @Destrolaric.
> 
> I think you did a good job to stabilize this. :-)
> 
> Moving forward we need a stable test suite that passes so we can test against new commits. I think we're at that point now.

Thx. I only fear that I fixed some symptoms. @Phillipus do you have any idea how we can get the macos build not canceled? Or is this because some quotas?

Phillipus:
> do you have any idea how we can get the macos build not canceled? Or is this because some quotas?

It could be because of the quota. You can test this by running the tests on your fork.

azoitl:
> > do you have any idea how we can get the macos build not canceled? Or is this because some quotas?
> 
> It could be because of the quota. You can test this by running the tests on your fork.

Ah didn't think about that. Yes in my fork macos builds.

PR #165, opened 2023-04-10
Introduce getChildrenRevIterable for Figure

azoitl:
With that the points which are iterating in reverse over all children are clearer marked.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionLayer.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionLayer.java
@@ -85,13 +85,10 @@ public class ConnectionLayer extends FreeformLayer {
 	 */
 	public void setConnectionRouter(ConnectionRouter router) {
 		connectionRouter = router;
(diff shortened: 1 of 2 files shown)


azoitl:
@Destrolaric thx in advance for any feedback.

Destrolaric:
@azoitl Hmm, is the second class needed? If we add interface to existing class this will probably be backwards compatible.

azoitl:
> @azoitl Hmm, is the second class needed? If we add an interface to the existing class, this will probably be backward compatible.

unfortunately if I added the interface to the first one API analysis tool is not happy and I didn't want to add a new exception to it. Also I didn't like the name of the first class as it didn't convey the meeting. I now changed it to  give the second class a better name and deprecated the first one. 

Within Draw2d I'm no only using the iterable and foreach loops. I think this will minimize the impact.

azoitl:
@Destrolaric sorry for poking you again. I should do today the 3.15 M1 build and would love to have this commit on board. Did you found time to look at my update?

Destrolaric:
@azoitl I think it should be okay.

Issue #155, opened 2023-03-28
Use Java Generics where possible


azoitl:
GEF Classic's dates back to a time without generics. While this is in general nothing bad it makes maintaining GEF Classic hard. Also clients and GEF Classic have to perform more instanceof checks and cast then necessary. Therefore GEF Classic is also hard to use. 

However just type all generics used is also not the solution as it may break client code. See for example the discussion in PR #81.

In the other side there are partly hidden issues in GEF Classic that a cleanup can reveal (e.g., Issue #154, or the planned upcoming PR https://github.com/azoitl/gef-classic/commit/69194fbf80e0c9b545788f662807c775d26ad6af). 

As now several people are interested in improving the GEF Classic code base in this respect this issue is here to coordinate any effort. But also discuss how to best test and evaluate all proposed changes to make the transition for  our clients as smooth as possible. 

Furthermore the starting point for this has already been done. There are already several PRs merged for the 3.15 release but there is still some way ahead of us. 

For starting this I've created a first PR #149 and based on that there are several commits in the pipeline that can be moved to PRs as soon the first one is merged:
     - https://github.com/azoitl/gef-classic/commit/7d1ad001ebb9590694ee27d4fae9c8eaf7bca9c2
     - https://github.com/azoitl/gef-classic/commit/588d8775e73821fe82eca6dd061de7c75356612f
     - https://github.com/azoitl/gef-classic/commit/19c84c8d8eb9204bfb3717a17ce721f53607e435
     - https://github.com/azoitl/gef-classic/commit/69194fbf80e0c9b545788f662807c775d26ad6af

**First set of Validation and Check rules**
In order to ensure a smooth transition for our users I would suggest at least the following tests and rules to be performed:
 - PRs should be as small as possible each addressing one distinct feature
 - It should pass our unit tests
 - Our examples (e.g., logic editor) do  behave as expected (e.g., for PR#149 I found an issue by doing this step)
 - Compile with additional clients (e.g., I do at least Eclipse GMF runtime and 4diac IDE). The more the merrier.

Destrolaric:
@azoitl We discussed this with our team. We can accept the #149. And we are planning to split other parts of our https://github.com/Destrolaric/gef-classic/pull/2 into several smaller prs. Is it acceptable?

vogella:
> And we are planning to split other parts of our [Destrolaric#2](https://github.com/Destrolaric/gef-classic/pull/2) into several smaller prs. Is it acceptable?

Yes

[...]

PR #170, opened 2023-04-12
Addresses #155 Typify LayoutListener list

azoitl:
@Destrolaric this would be my next one in the chain. I hope it is ok for you.


Destrolaric:
@azoitl LGTM

PR #172, opened 2023-04-12
[WB] Extend Translatable interface with additional default methods.

ptziegler:
Extend the interface to also allow calling the performTranslate method for points, dimensions and insets. For the sake of backwards-compatibility, provide a default implementation.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Translatable.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Translatable.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Translatable.java
```diff
@@ -15,6 +15,38 @@
...
+	}
+
+	/**
+	 * Translates this object horizontally by <code>insets.left</code> and
+	 * vertically by <code>insets.top</code>.
+	 * 
+	 * @param insets Insets which provides translation information
+	 * @since 3.13
+	 */
+	default void performTranslate(Insets insets) {
```
azoitl, line 47: I'm not sure about the translate with an Insets object. An Insets has in four values why take left and top? Or should this method have then a different name?

ptziegler: > why take left and top

When it comes to painting Figures, all coordinates have to be relative to the top-left corner of the canvas.

So if you want to e.g. only draw the client area of a Figure, you need to convert all coordinates, relative to the Figure (or their parent Figure), to their absolute coordinates of the panel. Meaning the bounds of the Figure as well as its Insets.

So in a way, the (left, top) coordinates of an Insets correspond to the (x, y) coordinates of a Point, which is why I treat them in the same way.

azoitl: Thx for the explanation. I think I was to tired when I reviewed the code yesterday evening. With your explanation it totally makes sense. Sorry for the disturbance.

PR #176, opened 2023-04-13
added gdk backend setting to jenkins file

azoitl:


azoitl:
will not wait for github builds as the jenkins file is not considered by them.

PR #181, opened 2023-04-13
Addresses #155: Typifysearch

azoitl:
With this cleanup the figure and edipart search was typified. As part of
this work it was identified that the TargetingTool documentation was
wrong. While there it was written that getExclusion set provides
editparts all its children provided a list of IFigures. Also the search
expected IFigures.

Furthermore AbstractTransferDropTargetListener delivered EditParts but
the search required IFigures. The code was fixed here was well and by
typifying the Collections it is now clearer.

@Destrolaric this would be my last commit for Issue #155 for now.


azoitl:
As there were no objections I merged it.

PR #182, opened 2023-04-14
[WB] Create a new Interval class to describe distances in 1-d space.

ptziegler:
The Interval is the counterpart to the Rectangle, for 1-dimensional space. Its primary use is to describe the length of rows and columns of layouts, in most cases the GridLayout.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
@@ -1,0 +1,230 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2023 Google, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Google, Inc. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.draw2d.geometry;
+
+import java.io.Serializable;
+
+/**
+ * Represents interval in 1-dimensional space.
+ *
+ * @author mitin_aa
+ * @author lobas_av
+ * @author scheglov_ke
+ * @coverage gef.draw2d
+ */
+public record Interval(int begin, int length) implements Serializable {
+
+	private static final long serialVersionUID = 0L;
+	public static final Interval INFINITE = new Interval(0, Integer.MAX_VALUE);
+
+	/**
+	 * Creates a new Interval with zero length and starting at the origin.
+	 * 
+	 * @since 3.13
+	 */
+	public Interval() {
+		this(0, 0);
+	}
+
+	/**
+	 * Creates a new Interval exact same parameters as the given Interval.
+	 * 
+	 * @param interval The interval to copy
+	 * @since 3.13
+	 */
+	public Interval(Interval interval) {
+		this(interval.begin, interval.length);
+	}
+
+	/**
+	 * Returns the center of this Interval. The center is the middle of the begin
+	 * and the end.
+	 * 
+	 * @return the center of this {@link Interval}.
+	 * @since 3.13
+	 */
+	public int center() {
+		return begin + length / 2;
+	}
+
+	/**
+	 * Returns whether the given value is within [begin, begin + length).
+	 * 
+	 * @param value A position somewhere within 1-dimensional space.
+	 * @return <code>true</code> if the value is contained by this interval.
+	 * @since 3.13
+	 */
+	public boolean contains(int value) {
+		return begin <= value && value < end();
+	}
+
+	/**
+	 * Calculates the distance to given <code>point</code>.
+	 *
+	 * Example:
+	 *
+	 * <pre>
+	 *  Let x=[10, 100]:
+	 *  	distance for point 3 is 7.
+	 *  	distance for point 10 is 0.
+	 *  	distance for point 50 is 0.
+	 *  	distance for point 110 is 0.
+	 *  	distance for point 150 is 40.
+	 * </pre>
+	 *
+	 * @param point the point to calculate the distance to.
+	 * @return the calculated distance.
+	 * @since 3.13
+	 */
+	public int distance(int point) {
+		if (point < begin) {
+			return begin - point;
+		} else if (point > end()) {
+			return point - end();
+		}
+		return 0;
+	}
+
+	/**
+	 * Returns the end of this Interval. The end is the sum of the begin and the
+	 * length of this Interval.
+	 * 
+	 * @return end of this Interval
+	 * @since 3.13
+	 */
+	public int end() {
+		return begin + length;
+	}
+
+	/**
+	 * Returns a new Interval which has the exact same parameters as this Interval.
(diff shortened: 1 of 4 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+ * Represents interval in 1-dimensional space.
+ *
+ * @author mitin_aa
+ * @author lobas_av
+ * @author scheglov_ke
+ * @coverage gef.draw2d
+ */
+public class Interval implements Serializable {
```
azoitl, line 23: to make our live easier for future enhancements of this class should it be final?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	/**
+	 * Creates a new Interval exact same parameters as the given Interval.
+	 * 
+	 * @param interval The interval to copy
+	 * @since 3.13
+	 */
+	public Interval(Interval interval) {
+		begin = interval.begin;
```
azoitl, line 56: I think calling this(interval.begin, interval.end) would be nicer here?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
```diff
@@ -0,0 +1,293 @@
...
+		length = interval.length;
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (o instanceof Interval) {
```
azoitl, line 62: if you prefer to not have Interval not final I think comparing the classes would be better. 

Should we add an this == o check?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
```diff
@@ -399,6 +399,22 @@ public Rectangle getIntersection(Rectangle rect) {
...
+	 * with the width as length is returned. Otherwise a new Interval beginning at
+	 * the y coordinate with the height as length is returned.
+	 * 
+	 * @param isHorizontal {@code true} for the horizontal Interval, {@code false}
+	 *                     for the vertical.
+	 * @return the {@link Interval} of projection on X or Y axis, depending on
+	 *         <code>isHorizonal</code> param.
+	 * @since 3.13
+	 */
+	public Interval getInterval(boolean isHorizontal) {
```
azoitl, line 414: not sure where you need it but would a getHorInterval and getVerInterval be better readable in client code?

azoitl (commented):
All in all I like this new Interval class a lot. I can see how it can also be used in Draw2D layout code cleanups! Thx a lot for providing it.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
ptziegler: That's a good point. I don't see the use case of extending geometric concepts like intervals.

I think you mentioned updating to Java 17 for the M2. So perhaps it would make sense to transform this class into a record instead? Intervals would then not only be final, but also immutable.

What do you think?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
ptziegler: I agree. Calling `getHorInterval()` looks a lot more descriptive than `getInterval(true)`

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
azoitl: I think for value classes like this one it totally makes sense. The grow and start method would then return new instances or?

ptziegler: > The grow and start method would then return new instances

Exactly.

azoitl: I would be interested to get more experience with records. So if it is for you, as many user of this new class, the right way I would go for it. Just started my Eclipse IDE to prepare the Java 17 patch. Hope to have something for you by today.

azoitl: If I remember correctly for record classes you also don't need to implement your own hash and equals methods or?

ptziegler: > If I remember correctly for record classes you also don't need to implement your own hash and equals methods or?

That's true. It also creates a simple toString() method.

In addition, it also automatically generates getter methods for all fields. At first glance, that should get rid of a lot of boilerplate code inside this class.

ptziegler: I've just pushed the changes. I'm not sure how Tycho will respond to this, but at least the IDE currently shows an error, because the compiler compliance is still set to Java 11.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
ptziegler: Done. With records, it's a requirement to always call the canonical constructor. :)

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
ptziegler: The equals method is now generated directly by Java.

ptziegler:
I've also took the liberty to sort the Interval methods alphabetically, in order to match the style of the other classes.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Interval.java
azoitl: Java 17 PR is underway PR #183. I really like the reduced boilerplate code.

azoitl:
I just noticed the copy right headers for the two new classes. Is the reason they are dated 2011, 2023 that they are migrated from window builder? Should we document that somehow?

ptziegler:
> Is the reason they are dated 2011, 2023 that they are migrated from window builder? Should we document that somehow?

That's right. We created this class back in 2011. But for GEF, it probably makes sense to only keep the 2023.

I think the fact that this class comes from the WindowBuilder project becomes clear from the commit message. Though I'll edit the message, just to be sure.

azoitl:
> That's right. We created this class back in 2011. But for GEF, it probably makes sense to only keep the 2023.

If it was originally created in 201, we have to keep that date. Because this is where the original copyright was created.

ptziegler:
> If it was originally created in 201, we have to keep that date. Because this is where the original copyright was created.

Understood. I've restored the original header and updated the commit message. That should hopefully avoid any misunderstandings.

azoitl:
In Eclipse 4diac we document this also in the contributor section of the copyright header. But I think the commit message is also good. Thx for bringing this feature to Draw2D.

PR #184, opened 2023-04-18
[WB] Add methods to modify the size and location of Rectangles.

ptziegler:
Instead of only being able to shrink a Rectangle symmetrically (meaning top/bottom and left/right), it should also be possible to only shrink one direction at a time.

Methods of the form shrinkX(), have been added, where X stands for the direction that is supposed to be shrunk. Shrinking a rectangle modifies both the x/y coordinate (if necessary) and the width/height.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -706,6 +706,16 @@ public class Rectangle implements Cloneable, java.io.Serializable, Translatable 
 	}
 
 	/**
+	 * Returns the X-coordinate of the left side of this Rectangle.
+	 * 
+	 * @return The X coordinate of the left side
+	 * @since 3.13
+	 */
+	public int left() {
+		return x;
+	}
+
+	/**
 	 * Returns <code>true</code> if this Rectangle's width or height is less than or
 	 * equal to 0.
 	 * 
@@ -900,6 +910,20 @@ public class Rectangle implements Cloneable, java.io.Serializable, Translatable 
 	}
 
 	/**
+	 * Updates the height to match the specified {@link #bottom()} coordinate. If
+	 * the new Y coordinate of the bottom happens to be smaller (i.e. above) than
+	 * the Y coordinate of the top, the height is set to {@code 0}.
+	 * 
+	 * @param newBottom The Y coordinate of the bottom
+	 * @return <code>this</code> for convenience
+	 * @since 3.13
+	 */
+	public Rectangle setBottom(int newBottom) {
+		height = Math.max(0, newBottom - y);
+		return this;
+	}
+
+	/**
 	 * Sets the height of this Rectangle to the specified one.
 	 * 
 	 * @param height The new height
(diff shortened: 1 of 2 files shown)


ptziegler:
I'd have to double-check, but I think this is the last contribution we need to get started with our GEF migration.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
```diff
@@ -1050,6 +1060,70 @@ public Rectangle shrink(int h, int v) {
...
+	/**
+	 * Shrinks the height of the rectangle by the given amount, keeping
+	 * {@link #top()}.
+	 * 
+	 * @param deltaY The vertical reduction amount
+	 * @return <code>this</code> for convenience
+	 * @since 3.13
+	 */
+	public Rectangle shrinkBottom(int deltaY) {
+		height = deltaY - y;
```
azoitl, line 1072: from the description I would have expected that here `height -= deltaY;' would be the right code. If I understand your code correctly it takes deltaY as an absolute Y coordinate. But this would be different to the other shrink methods.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
```diff
@@ -1050,6 +1060,70 @@ public Rectangle shrink(int h, int v) {
...
+	/**
+	 * Shrinks the width of the rectangle by the given amount, keeping
+	 * {@link #left()}.
+	 * 
+	 * @param deltaX The horizontal reduction amount
+	 * @return <code>this</code> for convenience
+	 * @since 3.13
+	 */
+	public Rectangle shrinkRight(int deltaX) {
+		width = deltaX - x;
```
azoitl, line 1099: I think here is the same problem as shrinkbottom

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
```diff
@@ -1050,6 +1060,70 @@ public Rectangle shrink(int h, int v) {
...
+	 * Shrinks the width of the rectangle by the given amount, keeping
+	 * {@link #right()}.
+	 * 
+	 * @param deltaX The horizontal reduction amount
+	 * @return <code>this</code> for convenience
+	 * @since 3.13
+	 */
+	public Rectangle shrinkLeft(int deltaX) {
+		x += deltaX;
+		width -= deltaX;
```
azoitl, line 1086: should there be some check or assertion that deltaX is smaller then width?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
```diff
@@ -1050,6 +1060,70 @@ public Rectangle shrink(int h, int v) {
...
+	 * Shrinks the height of the rectangle by the given amount, keeping
+	 * {@link #bottom()}.
+	 * 
+	 * @param deltaY The vertical reduction amount
+	 * @return <code>this</code> for convenience
+	 * @since 3.13
+	 */
+	public Rectangle shrinkTop(int deltaY) {
+		y += deltaY;
+		height -= deltaY;
```
azoitl, line 1113: should there be some check or assertion that deltaY is smaller then height?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
ptziegler: Hm, I think you're right. I suppose I was a little bit too eager to in having every method start with "shrink". `shrinkBottom` and `shrinkRight` allow you set the y-coordinate of the bottom and the x-coordinate side.

Originally, I had the methods named `setBottom` and `setRight`, which I've now went back to. I've also adjusted the checks, to make sure that neither width/height can never become negative.

azoitl: Reading the code I thought that. With the updates the patch is for me ready to merge

azoitl:
+1

PR #185, opened 2023-04-20
Clean-up in Zest Graph file

sebHollersbacher:


org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -160,42 +154,36 @@ public void dispatchMouseMoved(org.eclipse.swt.events.MouseEvent me) {
...
+
+		revealListeners = new ArrayList<RevealListener>(1);
+		this.addPaintListener(event -> {
+			if (!revealListeners.isEmpty()) {
+				// Go through the reveal list and let everyone know that the
+				// view is now available. Remove the listeners so they are
+				// only
+				// called once!
+				Iterator<RevealListener> iterator = revealListeners.iterator();
+				while (iterator.hasNext()) {
```
azoitl, line 175: as you are anyhow cleaning can you please change this to a for each loop

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -160,42 +154,36 @@ public void dispatchMouseMoved(org.eclipse.swt.events.MouseEvent me) {
...
-					}
+		this.connections = new ArrayList<>();
+		this.constraintAdapters = new ArrayList<>();
+		this.selectedItems = new ArrayList<>();
+		this.selectionListeners = new ArrayList<>();
+		this.figure2ItemMap = new HashMap<>();
+
+		revealListeners = new ArrayList<RevealListener>(1);
+		this.addPaintListener(event -> {
+			if (!revealListeners.isEmpty()) {
```
azoitl, line 169: not sure if this if is needed in modern java, the for each loop would anyhow perform the same check.

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -223,7 +211,7 @@ public void removeSelectionListener(SelectionListener selectionListener) {
 	 * 
 	 * @return List of GraphModelNode objects
 	 */
-	public List getNodes() {
+	public List<GraphNode> getNodes() {
```
azoitl, line 214: from my experience with other generics users would work better if you you use here ? extends GraphNode

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -299,7 +287,7 @@ public int getNodeStyle() {
 	 * 
 	 * @return list of GraphModelConnection objects
 	 */
-	public List getConnections() {
+	public List<GraphConnection> getConnections() {
```
azoitl, line 290: As above  ? extends GraphConnection

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -320,15 +308,15 @@ public void setSelection(GraphItem[] items) {
...
+		setSelection(nodes.toArray(new GraphItem[] {}));
 	}
 
 	/**
 	 * Gets the list of currently selected GraphNodes
 	 * 
 	 * @return Currently selected graph node
 	 */
-	public List getSelection() {
+	public List<GraphItem> getSelection() {
```
azoitl, line 319: ? extends GraphItem

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -977,7 +941,7 @@ void addNode(GraphContainer graphContainer) {
 	}
 
 	void registerItem(GraphItem item) {
-		if (item.getItemType() == GraphItem.NODE) {
+		if (item instanceof GraphNode) {
```
azoitl, line 944: not sure if this change is save. Would prefere if we could keep to original version.

azoitl: Or did you combine here the container and the node as both have the same code?

azoitl: but even then I think I would combine both of the original ifs

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
sebHollersbacher: I ignored this loop due to the remove call in it

azoitl: Oh sorry missed that.

PR #187, opened 2023-04-22
[JUnit] Stabilize PaintDamageEraseTest and split up test cases

ptziegler:
It appears as if pumping the SWT event queue only once is not sufficient to process the update of the UpdateManager instance.

Define a new method which pumps the SWT events for an arbitrary but fixed amount of time.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PaintDamageEraseTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PaintDamageEraseTest.java
@@ -38,119 +38,84 @@ public class PaintDamageEraseTest extends TestCase implements UpdateListener {
 	protected IFigure container;
 	private String errMsg;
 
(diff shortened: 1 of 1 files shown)


ptziegler:
The tests are executed and working, but a sample size of one is probably not enough to say whether this sporadic test failure have been fixed :/

azoitl:
Thanks for finding and fixing that issue. I tested also on my machine. I still have one not stable test. But this one is good. So I'll merge at it looks like an improvement.

ptziegler:
> I still have one not stable test.

Are you talking about the ConnectionEndPointMoveTest, by any chance? I think the same approach with the `waitEventLoop` can be used there as well.

azoitl:
> > I still have one not stable test.
> 
> Are you talking about the ConnectionEndPointMoveTest, by any chance? I think the same approach with the `waitEventLoop` can be used there as well.

Yes this is the flanky one on my side.

PR #188, opened 2023-04-23
[JUnit] Stabilize ConnectionEndPointMoveTest via waitEventLoop method

ptziegler:
Use the same approach used to stabilize the PaintDamageErase tests. When pumping the SWT event loop only once, the update of the UpdateManager may have not been fully performed. In order to ensure that it has been fully processed, pump the events for an arbitrary but fixed amount of time.

For the sake of reusability, the waitEventLoop method has been moved to the BaseTestCase class. Both test cases extend this base class in order to access this method. The ConnectionEndPointMove test will pump SWT events for a total of 100ms.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/BaseTestCase.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/BaseTestCase.java
@@ -23,6 +23,7 @@ import org.eclipse.draw2d.geometry.Point;
 import org.eclipse.draw2d.geometry.Rectangle;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Shell;
 
 /**
  * @author Pratik Shah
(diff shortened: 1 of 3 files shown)


azoitl:
You are amazing. Finally the all tests run smoothly also on my machine.

I'll tried to fix the merge conflict directly in github. did this the first time here. therefore could you be so kind and check if I didn't mess up.

ptziegler:
> I'll tried to fix the merge conflict directly in github. did this the first time here. therefore could you be so kind and check if I didn't mess up.

Yeah... the merge editor on GitHub is a little weird.

It seems like the formatting broke and one line is using spaces instead of tabs. I'll have a closer look this afternoon.

ptziegler:
The merge editor also messed up the Git history because changes were merged from the master to the branch. But everything should've been cleaned up again.

azoitl:
Ok goo to know. Next time I will not do it with out.

azoitl:
> Ok goo to know. Next time I will not do it with out.

I forgot to mention that Github gave me the sqash commits option here. I think that that the history would have been fine. We can try that out the next time.

Issue #146, opened 2023-03-20
Upgrade to Java 17


azoitl:
As announced by the Eclipse Platform team the Eclipse Platform will migrate to Java 17 with the upcoming 2023-06 release I would like to follow with GEF Classic rather soon. Therefore depending on the feedback of the community I would either do it as part of our 2023-06 (GEF Classic 3.16) or 2023-09 (GEF Classic 3.17) release. 

What is the opinion of the community on this?

\cc @pcdavid @Phillipus @Destrolaric @vogella

Phillipus:
Hi @azoitl,

does this simply mean changing these entries: `Bundle-RequiredExecutionEnvironment: JavaSE-17`? If so, is there any harm in leaving them at `JavaSE-11` in case GEF and Draw2d are being used outside of the Eclipse platform?

azoitl:
> Hi @azoitl,
> 
> does this simply mean changing these entries: `Bundle-RequiredExecutionEnvironment: JavaSE-17`? If so, is there any harm in leaving them at `JavaSE-11` in case GEF and Draw2d are being used outside of the Eclipse platform?

@Phillipus I was pretty sure about GEF but not about Draw2D and ZEST, therefore I checked. All three have dependencies to several Eclipse Platform plugins. While Draw2D only references eclipse.ui for the others it is a much deeper integration. So I don't see a use case for GEF Classic outside of  the Eclipse Platform. The only reason we could have is that we want GEF Classic 3.16 be also be used by non Eclipse 2023-06 users (e.g., Archi). For my main use-case of GEF Classic, 4diac  IDE, I have no concerns, I just submitted the Java 17 patches and we upgrade to latest Eclipse Platform every 3 months. 

On the other hand I guess I see it a bit like @vogella if we are on Simrel the Simrel versions should match and users can then decide which Eclipse Platform release with the according GEF Classic release they would like to use. As both is part of the same update site it makes their life easier.

Phillipus:
@azoitl Agreed on all your points. Archi is on Java 17 now anyway. :-)

azoitl:
@Phillipus I'm very happy to have you on board with a second RCP for feedback. I'm also looking so forward to Java 17. We where hold back by Xtext.

Destrolaric:
[...]

PR #189, opened 2023-04-24
Update compiler compliance of org.eclipse.draw2d to Java 17

ptziegler:
The properties 'org.eclipse.jdt.core.compiler.codegen.targetPlatform' and 'org.eclipse.jdt.core.compiler.source' are still set to Java 11. As a result, language features such as records are not recognized within the IDE.

See #146


azoitl:
Strange that I missed that. Thx for finding and fixing.

PR #195, opened 2023-04-28
Reproducer for issue 194

ptziegler:


ptziegler:
Ignore this one. I didn't noticed that a new rectangle is created for the second part of the test.

Issue #196, opened 2023-04-28
Shrink method of rectangles may result in negative width/height.


ptziegler:
If I shrink a rectangle by values larger than its current width or height, the new values may be negative.

I would expect each size of a rectangle to always be non-negative.

ptziegler:
Resolved with 6de0077525dae322d3d75e1f89a28a738a63ef27

PR #197, opened 2023-04-28
Adjust shrink method of rectangles to prevent negative width/height.

ptziegler:
This negative value occurs when the rectangle has a border, but its height or width is smaller than the height or width of its insets. The shrink method then tries to reduce the width/height of the rectangle by the width/height of the Insets which may then load to an error.


ptziegler:
Resolves #196

PR #198, opened 2023-04-29
Reworked Thumbnail to not use scaled graphics anymore addresses #106

azoitl:
The thumbnail is now using only the SWTGraphics and now scaled graphics anymore. Furthermore we stop now active updates of the thumbnail runable if during a repaint request it is running. 

@Phillipus I would be happy on your feedback. I've the slight feeling if the performance got worse with this change but not fully sure.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/parts/Thumbnail.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/parts/Thumbnail.java
@@ -23,7 +23,6 @@ import org.eclipse.draw2d.Figure;
 import org.eclipse.draw2d.Graphics;
 import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.SWTGraphics;
-import org.eclipse.draw2d.ScaledGraphics;
 import org.eclipse.draw2d.UpdateListener;
 import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.draw2d.geometry.Rectangle;
@@ -63,8 +62,7 @@ public class Thumbnail extends Figure implements UpdateListener {
 		private Dimension tileImageSize;
 		// GC and Graphics to let the source figure paint on the tile image
 		private GC tileGC;
-		private SWTGraphics tileGCGraphics;
-		private ScaledGraphics tileGraphics;
+		private SWTGraphics tileGraphics;
 		// GC used to copy from the tile image into the thumbnail image
 		private GC thumbnailGC;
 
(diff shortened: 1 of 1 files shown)


Phillipus:
I tested it on Mac and it works pretty much the same as before.

Did you notice a difference by adding this?

```java
	@Override
	public void repaint(int x, int y, int w, int h) {
		if (updater.isRunning()) {
			updater.stop();
		}
		super.repaint(x, y, w, h);
	}
```

azoitl:
> Did you notice a difference by adding this?
> 
> ```java
> 	@Override
> 	public void repaint(int x, int y, int w, int h) {
> 		if (updater.isRunning()) {
> 			updater.stop();
> 		}
> 		super.repaint(x, y, w, h);
> 	}

Originally I only had it in the notifyPainting method and especially on zooming I noticed quite some speedup, as it cancels unnecessary paints.

azoitl:
@Phillipus, I spent some time today looking into the performance issue of this and in 4diac IDE with a medium sized diagram it feels noticeable slower until the thumbnail is updated. Unfortunately I found no good way to reliably measure.

I have a strong feeling that the scale line in the code below is the problem:
```
org.eclipse.draw2d.geometry.Point p = getSourceRectangle().getLocation();
tileGraphics.translate(-p.x * getScaleX() - sx1, -p.y * getScaleY() - sy1);
tileGraphics.scale(getScaleX());
sourceFigure.paint(tileGraphics);
tileGraphics.popState();
```
I looked into the the code of both and the SWTGraphics seems to be more complicated.  So what I would try and I'm not sure if possible is to set the scale only once. What do you think?

azoitl:
I found a way to move the scale method to the tileGraphics creation. For 4diac IDE it didn't change a lot and I think it is a different problem we have in 4diac IDE (has to do with our incremental growing canvas).  But what I noticed is that the quality of the thumbnail is much better without the scaled graphics. So if @Phillipus agrees I think this PR can be submitted and it will lead to an improvement of GEF Classic.

azoitl:
@Phillipus did you find some time to read my comment and re-review my code?

Phillipus:
> @Phillipus did you find some time to read my comment and re-review my code?

Sorry, not yet. Will do it this week! :-)

Phillipus:
@azoitl I did a very quick check on this and, unfortunately, the thumbnail is rendering a diagram multiple times, just like it did on Mac before the Mac workaround was added.

For example, this diagram:



Is rendered like this:

azoitl:
@Phillipus this is intersting. Because I thought I kept your code and just changed it such that it uses the SWTGraphics instead of the Scaled graphics. Let me look again on the code to find out where I messed up. Unfortunately I have no mac for testing.

Phillipus:
> Unfortunately I have no mac for testing.

This was on Windows.

azoitl:
> > Unfortunately I have no mac for testing.
> 
> This was on Windows.

Ok. Strange.

azoitl:
@Phillipus could you try without my the last commit in my PR?

Phillipus:
Commit b90f8f65 is better. However, the thumbnail is initially blank and is only drawn when resized.

azoitl:
> Commit [b90f8f6](https://github.com/eclipse/gef-classic/commit/b90f8f6555483b07c64612247bcd953a16c8750e) is better. However, the thumbnail is initially blank and is only drawn when resized.

Ok back to concept. I fear I really have to setup a windows dev environment for GEF Classic in my VM.

Phillipus:
@azoitl I tested on Windows and Mac and it all seems to be working OK.

PR #199, opened 2023-04-29
Cleanup of interval record

azoitl:
@ptziegler I found two minor things reported by sonar and PDE tools that I fixed in your new interval record.


ptziegler:
The serialVersionUID is removed because it's 0 by default?

azoitl:
> The serialVersionUID is removed because it's 0 by default?

Yes this is what sonar complaint. Also I read in some oracle blog post that java treats serialized for records special and in their example nothing special was needed.

ptziegler:
That's what I read when I looked it up today as well. No complaints from my side if it's removed, then.

PR #203, opened 2023-05-17
Added hide-nodes feature

sebHollersbacher:
Added feature to hide nodes inside a graph.
The feature can be enabled in the Graph-Constructor.


azoitl:
+1

PR #207, opened 2023-06-05
Added Dialog to search nodes in graph

sebHollersbacher:


Phillipus:
I don't mean to sound negative, but I don't think a UI dialog belongs in the `Graph` class. This kind of functionality should be added outside of this class by consumers of the framework. Also, this would allow different key bindings to be set in Commands.

azoitl:
> I don't mean to sound negative, but I don't think a UI dialog belongs in the `Graph` class. This kind of functionality should be added outside of this class by consumers of the framework. Also, this would allow different key bindings to be set in Commands.

@Phillipus thanks for the feedback. You are totally right. This was a bit an overshot. The creation of the search dialog is now moved to a snippet. But the dialog is there for convenience. I think that is fine and I will merge this PR.

PR #218, opened 2023-06-15
Fixed zooming when scaled graphics is activated.

### LINKED ISSUE #216: Issue with zoom levels and drawing after update
Hello, after updating to the latest Eclipse and GEF 3.16, our GEF editor doesn't zoom properly anymore. If the zoom level is NOT 100%, things are not drawn where they should be, and no zooming happens.

For example, observe the blue grouping that is selected, but the resizing controls (which is where the selection bounds are, and where I can actually interact with the item) are below and slightly to the right.
image

(at 89% zoom).

---

Observe the Outline view on the right. The blue outline shown is where the viewport should be, but you see the actual editor is slightly above again (you can see the same blue grouping from above should be at the very bottom but it isn't).
image (1)

---
At 100% zoom things are where they should be,
image

But at different zoom levels they are drawn in the incorrect location (also the size is the same).
image

I have the Map item selected but I clicked in the middle of the editor to do it, even though it's drawn in the top left. At lower zooms it moves to the bottom right.
image

Again, here I clicked in the middle to select it (after spending a while to find it) even though it is somewhere else.

---

Here you can see strange behavior when zooming, the selection area is zoomed correctly and moves but the actual items are drawn as if they are 100% zoom (in the incorrect location).

https://github.com/eclipse/gef-classic/assets/9405706/7bbb1975-b408-49c1-870d-10845ac47d22

(as you can see it is very hard to find things).

---

Was something changed regarding this in the release? This release shouldn't have been breaking. Please advise as we would really like to use the latest Eclipse/GEF but this is a big issue.

Eclipse GEF	GEF Classic GEF (MVC)	3.14.0.202306071707	org.eclipse.gef
Eclipse GEF	GEF Classic Draw2d	3.13.0.202306071707	org.eclipse.draw2d

azoitl:
Fixes  #216.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableFreeformLayeredPane.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableFreeformLayeredPane.java
@@ -69,7 +69,7 @@ public class ScalableFreeformLayeredPane extends FreeformLayeredPane implements 
 			super.paintClientArea(graphics);
 		} else {
 			Graphics graphicsToUse = IScalablePaneHelper.prepareScaledGraphics(graphics, this);
-			paintChildren(graphics);
+			paintChildren(graphicsToUse);
 			IScalablePaneHelper.cleanupScaledGraphics(graphics, graphicsToUse);
 		}
 	}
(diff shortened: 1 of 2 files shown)


kysmith-csg:
I'm not familiar at all with the code, but makes sense. Any chance you can make a 3.16.1 release? 😄

kysmith-csg:
image
image


Built it locally and it looks great :D

azoitl:
> I'm not familiar at all with the code, but makes sense. Any chance you can make a 3.16.1 release? smile

I'll try. Its the first time I have to do this.

@kysmith-csg  If you want a work around you can also try it the new feature of not useing the scaled graphics but the swt graphics. For that you need to create the `ScalableFreeformRootEditPart` with false as constructor parameter. In my experience this should lead to nicer graphics.

ptziegler:
+1
The problem also showed up in our product after updating to the 3.16, but not on this branch.

PR #222, opened 2023-07-03
Quick clean-ups for flow example

azoitl:


azoitl:
As this commit is only cleaning an example, and all checks pass. I'm merginging.

PR #224, opened 2023-07-03
Cleaned text example and a bit of digraph2

azoitl:


azoitl:
Again as it is only a clean-up of an example and the tests pass I'm merging it.

PR #227, opened 2023-07-09
[WB] Add a new getTextExtent() method to the Graphics class

ptziegler:
This method allows clients to dynamically calculate the width and height of a string text. This may be useful in order to e.g. determine, whether a label has to be truncated.


azoitl:
Isn't this somehow duplicating the FigureUtilities and TextUtiltities functions?

azoitl:
Unfortunately they are rather hidden.

PR #231, opened 2023-07-26
Migrate Draw2D, GEF & Zest JUnit Tests from 3 to 4

ptziegler:
This change migrates all test cases to the latest JUnit version. With JUnit 4, annotations are used to control the execution flow. Given that the naming convention in JUnit 3 is rather fixed, a large chunk of this adaptation could be done automatically.

Normally, one would import statically import methods such as assertTrue(), assertEquals(), etc... However, those methods used to be provided by the TestCase class. In order to keep the code change to a minimum, we imitate this behavior by extending the Assert class instead.

The empty setUp() and tearDown() methods, as well as blank constructors have been removed.

Note that we only migrate to Junit4 instead of JUnit5, simply because Tycho is currently unable to execute those test suites. However, it should be much more straightforward to switch from 4 to 5 than from 3 to 4.

In short:
- Extend Assert instead of TestCase
- Annotate all tests with @Test
- Annotate disable tests/test classes with @Ignore
- Annotate setUp() method with @Before, make public
- Annotate afterEach() method with @After, make public
- Convert test suites into JUnit4 format

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AdvancedGraphicsTests.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AdvancedGraphicsTests.java
@@ -32,6 +32,10 @@ import org.eclipse.swt.graphics.Resource;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
 public class AdvancedGraphicsTests extends BaseTestCase {
 
 	static final int LINE[] = new int[] { 5, 5, 20, 20, 35, 5, 50, 5 };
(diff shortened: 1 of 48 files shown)


ptziegler:
Note that both ColorConstantTest and TextFlowWrapTest are both currently not executed. I expect the same amount of successful tests (305) and 2 skipped tests.

azoitl:
Thx for this work this is great!

Regarding Junit 5, I wanted to point out that I know that in 4diac IDE we are using Tycho with Junit 5. If you are interested this is our root pom file: https://git.eclipse.org/c/4diac/org.eclipse.4diac.ide.git/tree/pom.xml?h=develop

azoitl:
@ptziegler thx for the explanation. I must confess that I'm not the one who did your test infra. But explains some issues I had when upgrading dependencies.

PR #233, opened 2023-07-31
Cleaning of Draw2d examples

azoitl:
Using the new autoclean up and compiler setttings to find and fix more issues in the Draw2d examples


azoitl:
As it is only a clean-up of examples I'm directly merging without additional review.

PR #234, opened 2023-07-31
[#155] Typed node list

azoitl:
Using typed generics for NodeLists. This allowed quite some code clean-up across the whole draw2d graph package. Similar changes are in work for EdgeList and RankList.

@Destrolaric I would be happy if you could have a look on this PR.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundBreakCycles.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundBreakCycles.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2010 IBM Corporation and others.
+ * Copyright (c) 2003, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -16,7 +16,7 @@ package org.eclipse.draw2d.graph;
  * been modified to handle the presence of Subgraphs and compound cycles which
  * may result. This algorithm determines a set of edges which can be inverted
  * and result in a graph without compound cycles.
- * 
+ *
  * @author Daniel Lee
  * @author Randy Hudson
  * @since 2.1.2
@@ -29,38 +29,33 @@ class CompoundBreakCycles extends GraphVisitor {
 	 * from the list.
 	 */
 	private NodeList graphNodes;
-	private NodeList sL = new NodeList();
+	private final NodeList sL = new NodeList();
 
-	private boolean allFlagged(NodeList nodes) {
-		for (int i = 0; i < nodes.size(); i++) {
-			if (nodes.getNode(i).flag == false)
-				return false;
-		}
-		return true;
+	private static boolean allFlagged(NodeList nodes) {
+		return nodes.stream().allMatch(n -> n.flag);
 	}
 
-	private int buildNestingTreeIndices(NodeList nodes, int base) {
-		for (int i = 0; i < nodes.size(); i++) {
-			Node node = (Node) nodes.get(i);
-			if (node instanceof Subgraph) {
-				Subgraph s = (Subgraph) node;
+	private static int buildNestingTreeIndices(NodeList nodes, int base) {
+		for (Node node : nodes) {
+			if (node instanceof Subgraph s) {
 				s.nestingTreeMin = base;
 				base = buildNestingTreeIndices(s.members, base);
 			}
-			node.nestingIndex = base++;
+			node.nestingIndex = base;
+			base++;
 		}
 		return base++;
 	}
 
-	private boolean canBeRemoved(Node n) {
+	private static boolean canBeRemoved(Node n) {
 		return !n.flag && getChildCount(n) == 0;
 	}
 
-	private boolean changeInDegree(Node n, int delta) {
+	private static boolean changeInDegree(Node n, int delta) {
 		return (n.workingInts[1] += delta) == 0;
 	}
 
-	private boolean changeOutDegree(Node n, int delta) {
+	private static boolean changeOutDegree(Node n, int delta) {
 		return (n.workingInts[2] += delta) == 0;
 	}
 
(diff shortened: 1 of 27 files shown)


azoitl:
@Destrolaric @ptziegler  could you please have a look on this change?

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundBreakCycles.java
```diff
@@ -29,38 +29,34 @@ class CompoundBreakCycles extends GraphVisitor {
...
+	private final NodeList sL = new NodeList();
 
-	private boolean allFlagged(NodeList nodes) {
-		for (int i = 0; i < nodes.size(); i++) {
-			if (nodes.getNode(i).flag == false)
-				return false;
-		}
-		return true;
+	private static boolean allFlagged(NodeList nodes) {
+		return nodes.stream().noneMatch(n -> !n.flag);
```
ptziegler, line 35: I think it'd be more readable to use `allMatch(n -> n.flag)`, to avoid the double-negation

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/RankAssignmentSolver.java
```diff
@@ -273,32 +267,34 @@ void updateSubgraph(Node root) {
...
 			stack.push(node);
 			while (!stack.isEmpty()) {
-				node = (Node) stack.pop();
+				node = stack.pop();
 				tree.add(node);
-				Iterator neighbors = node.iteratorNeighbors();
+				Iterator<Node> neighbors = node.iteratorNeighbors();
 				while (neighbors.hasNext()) {
-					Node neighbor = (Node) neighbors.next();
+					Node neighbor = neighbors.next();
```
ptziegler, line 297: You can probably get rid of the iterator completely via `node.iteratorNeighbors().forEachRemaining(neighbor -> {...}`

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundPopulateRanks.java
```diff
@@ -16,21 +16,22 @@
...
 		 * Remove long containment edges at this point so they don't affect MinCross.
 		 */
-		Iterator containment = graph.containment.iterator();
+		Iterator<Edge> containment = graph.containment.iterator();
 		while (containment.hasNext()) {
-			Edge e = (Edge) containment.next();
+			Edge e = containment.next();
 			if (e.getSlack() > 0) {
 				graph.removeEdge(e);
 				containment.remove();
```
ptziegler, line 37: It'd also use a for-each loop here. Perhaps even something like `graph.containment.stream().filter(...).forEach(...)` to also get rid of the if() check.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/EdgeList.java
```diff
@@ -14,100 +14,93 @@
...
+	 *
 	 * @return <code>true</code> if completely flagged
 	 */
 	public boolean isCompletelyFlagged() {
-		for (int i = 0; i < size(); i++) {
-			if (!getEdge(i).flag)
-				return false;
-		}
-		return true;
+		return stream().noneMatch(e -> !e.flag);
```
ptziegler, line 80: As mentioned earlier, I think an allMatch(e -> e.flag) is more readable.

ptziegler:
It probably would've been a little easier to split this up into separate pull requests, but I aside from those minor adaptations, I think it's looking good.

azoitl:
> It probably would've been a little easier to split this up into separate pull requests, but I aside from those minor adaptations, I think it's looking good.

Yes it got bigger then originally intended. Thx for the feedback. This are very nice suggestions. Will rework it accordingly.

azoitl:
@ptziegler I just noticed that there is one more commit in this PR then I had intended. I hope I can get rid of it.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundBreakCycles.java
azoitl: Changed.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/RankAssignmentSolver.java
azoitl: This should be part of a separate pull request. Will fix it there.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundRankSorter.java
```diff
@@ -62,27 +64,33 @@ public int hashCode() {
...
 
+	@Override
 	void optimize(DirectedGraph g) {
 		CompoundDirectedGraph graph = (CompoundDirectedGraph) g;
 		Iterator containment = graph.containment.iterator();
-		while (containment.hasNext())
+		while (containment.hasNext()) {
 			graph.removeEdge((Edge) containment.next());
+		}
 		graph.containment.clear();
```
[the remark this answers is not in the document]
azoitl: This should be part of a separate pull request. Will fix it there.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundPopulateRanks.java
azoitl: This should be part of a separate pull request. Will fix it there.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/EdgeList.java
azoitl: This should be part of a separate pull request. Will fix it there.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundPopulateRanks.java
azoitl: I just noticed the `graph.removeEdge(e)` changes the container which is iterated over. Will the `stream().filter().foreEch()` work in this case?

ptziegler: Good catch. I also missed the call to containment.remove(); In that case, the iterator has to be used.

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundBreakCycles.java
```diff
@@ -77,58 +73,62 @@ private void cycleRemove(NodeList children) {
 			// outDegree - inDegree
 			Node max = findNodeWithMaxDegree(children);
 			if (max != null) {
-				for (int i = 0; i < children.size(); i++) {
-					Node child = (Node) children.get(i);
-					if (child.flag)
+				for (Node child2 : children) {
+					Node child = child2;
```
ptziegler, line 77: Is there any reason to create this local variable instead of using the loop parameter? I noticed something similar in buildNestingTreeIndices().

azoitl: This is sometimes created by the auto clean-up of Eclipse IDE. I tried to fix it manually. But it looks like I missed a few. Thx for noting.

azoitl: fxed

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/CompoundBreakCycles.java
```diff
@@ -29,38 +29,34 @@ class CompoundBreakCycles extends GraphVisitor {
...
 	}
 
-	private int buildNestingTreeIndices(NodeList nodes, int base) {
-		for (int i = 0; i < nodes.size(); i++) {
-			Node node = (Node) nodes.get(i);
-			if (node instanceof Subgraph) {
-				Subgraph s = (Subgraph) node;
+	private static int buildNestingTreeIndices(NodeList nodes, int base) {
+		for (Node node2 : nodes) {
+			Node node = node2;
```
ptziegler, line 40: My apologies that I wasn't clear enough. The auto clean-up messed up this method here as well :sweat_smile:

azoitl: I got that from your comment but I thought I fixed it in my latest commit. Now I'm confused. Let me check

azoitl: I'm a dumbass. Forgot to push :facepalm:

azoitl: Many thx for your time and patience.

ptziegler: No need to mention it :)

ptziegler (approved):

PR #235, opened 2023-08-01
Removed reveal button from Graph

sebHollersbacher:
Created new Zest example for hide-nodes and moved the reveal button from Graph to example


org.eclipse.zest.examples/src/org/eclipse/zest/examples/swt/GraphSnippet14.java
```diff
@@ -0,0 +1,80 @@
+/*******************************************************************************
+ * Copyright 2023
```
azoitl, line 2: please put your name here as you are the copyright holder.

azoitl (changes requested):

azoitl (approved):

PR #237, opened 2023-08-02
Add option to not use scaled graphics to PrintOperation

Phillipus:
Add support for optional scaled graphics in `PrintOperation` class.


azoitl:
@Phillipus I looked a bit deeper into the code. The only thing that I think will be backward compatible is to let PrintGraphics implement all methods of ScaledGraphics and switch in these between the normal or the ScaledGraphics version. What do you think?

PR #238, opened 2023-08-12
Typify dirtyRegion map in UpdateListener and UpdateManager

ptziegler:
This change adjusts the notifyPainting() method in UpdateListener and the firePainting() method in UpdateManager to expect a generic map of (IFigure, Rectangle) pairs, rather than a raw map.

This should not cause errors in client code, as calling or overwriting this method can still be done using a raw map (which will create a "unchecked or unsafe operations" warning).

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -46,7 +46,7 @@ public class DeferredUpdateManager extends UpdateManager {
 	}
 
 	private Rectangle damage;
-	private Map dirtyRegions = new HashMap();
+	private Map<IFigure, Rectangle> dirtyRegions = new HashMap<>();
 
 	private GraphicsSource graphicsSource;
 	private List invalidFigures = new ArrayList();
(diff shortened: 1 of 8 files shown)


ptziegler:
#155

org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java
```diff
@@ -266,15 +266,15 @@ protected void releaseGraphics(Graphics graphics) {
 	 * regions.
 	 */
 	protected void repairDamage() {
-		Iterator keys = dirtyRegions.keySet().iterator();
+		Iterator<IFigure> keys = dirtyRegions.keySet().iterator();
 		Rectangle contribution;
 		IFigure figure;
 		IFigure walker;
 
 		while (keys.hasNext()) {
```
azoitl, line 274: as we need both key and value wouldn't it be better if we directly iterate over the entry set? i.e. dirtyRegions.entrySet().forEach( ...

azoitl (commented):
I've tested your PR with GEF Classic's logic editor, 4diac IDE and GMF runtime. In general it looks good. Only one small question below.

org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java
ptziegler: I'm always hesitant to change more that what's absolutely necessary. But in principle, you're right. A lot of the iterators can be replaced with the for-each paradigm. You can make a similar argument for e.g. the changes to the Thumbnail class.
I'll go over the changes and see if I can simplify it a little bit more.

ptziegler: @azoitl Both the DeferredUpdateManager and the Thumbnail also use for-each loops now.

azoitl: The main reason I suggested the change is that I found it strange from a performance reason to iterate over the keys and then do a look up. But now with the forEach loop it looks so much clearer. Thx for the change. Again tested with GEF Classic Logic Editor, 4diac IDE and compiled GMF runtime. Looks good now.

PR #239, opened 2023-08-13
Typified command list on compound command. Cleaned-up compound command.

azoitl:
Addresses #155 

@Destrolaric @ptziegler as it is only changing GEF internals it shouldn't have any impact on users, or?

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/commands/CompoundCommand.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/commands/CompoundCommand.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -28,11 +28,11 @@ import java.util.List;
  */
 public class CompoundCommand extends Command {
 
-	private List commandList = new ArrayList();
+	private final List<Command> commandList = new ArrayList<>();
 
 	/**
 	 * Constructs an empty CompoundCommand
-	 * 
+	 *
 	 * @since 2.0
 	 */
 	public CompoundCommand() {
@@ -40,7 +40,7 @@ public class CompoundCommand extends Command {
 
 	/**
 	 * Constructs an empty CompoundCommand with the specified label.
-	 * 
+	 *
 	 * @param label the label for the Command
 	 */
 	public CompoundCommand(String label) {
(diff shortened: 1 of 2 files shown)


ptziegler:
> as it is only changing GEF internals it shouldn't have any impact on users, or?

The only part that might cause problems is the type returned by getCommands(). Currently, something like this is currently valid, but would cause errors after this change:
```
List<Object> commands = compoundCommand.getCommands();
```
Though I think that this is a rather theoretical example, because I don't see why one would want to store the value like this rather than a raw list or directly as a generic list of commands.

azoitl:
I missed that.  I have this change for some time in my backlog. The better solution is to return an < ? extends Command> this gives existing clients more flexibility. Will change it.

ptziegler:
> I missed that. I have this change for some time in my backlog. The better solution is to return an < ? extends Command> this gives existing clients more flexibility. Will change it.

I don't think that's a good idea. Then you can only store the returned value as a `List<? extends Command>`, but no longer as a `List<Command>`.

ptziegler:
> I don't think that's a good idea. Then you can only store the returned value as a `List<? extends Command>`, but no longer as a `List<Command>`.

Whereas with with `List<Command>`, you can store it as both.

org.eclipse.gef/src/org/eclipse/gef/commands/CompoundCommand.java
```diff
@@ -142,21 +150,23 @@ public Object[] getChildren() {
 	/**
 	 * @return the List of contained Commands
 	 */
-	public List getCommands() {
+	public List<? extends Command> getCommands() {
```
ptziegler, line 153: This now forces users to store the return value in a `List<? extends Command>`. If it just returns a `List<Command>`, then the user can store it as either one of those.

azoitl: The reason I'm typically using this pattern is that I've seen users doing in similar situations the following:

`List<MyCommand> lst = (List<MyCommand>) compound.getCommnds()`

Which is not possible if we return a `List<Command>`. But I'm not sure which of the both is more common and has more impact.

ptziegler: Hm, fair point. But that brings up another possibility: Users already can do something like:
`List<MyCommand> lst = compound.getCommands()`
Which now causes compile errors. With the <? extends Command>, the user has at least the ability to fix this compile error, so I would then say that's the preferable option.

ptziegler (approved):

azoitl:
Thx for the feedback. Will merge it now.

org.eclipse.gef/src/org/eclipse/gef/commands/CompoundCommand.java
ptziegler: > But I'm not sure which of the both is more common and has more impact.

As an FYI: I just had a look at how we use GEF in our other product and we always use one of the following patterns:
```
for (Object o :  compound.getCommands()) {
    Command command = (Command)o;
    ...
}
```
```
List<Command> commands = compound.getCommands();
```
At least on our side, the second one will require code adaptations once we update, but I'm not sure how representative a sample size of one is going to be 😅.
But I think your point still stands. With the <? extends Command>, it will always be possible to resolve the compile errors in user code with minimal effort.

azoitl: This nicely shows why it is so hard to clean the code. 

But not cleaning is IMHO also not an option. I already found one place where the documentation was not matching the implementation. With Java generics this is now clearer. I hope our users will see it the same way.

PR #240, opened 2023-08-14
[#155] Typed edge list

azoitl:
Using typed generics for EdgeLists. This allowed quite some code clean-up across the whole draw2d graph package. 

@ptziegler @Destrolaric again I would be happy for feedback. 

There is one patch in the stack that would then do something similar for the RankList.


ptziegler (approved):
LGTM

PR #241, opened 2023-08-14
[Clean-up] added missing @Override annotations

azoitl:
Using Eclipse IDE's quickfix all missing @Override annotations where added.

This reduced the warnings to < 3.5k.


azoitl:
As it is only adding @Override annotations and not changing any functionality I'm merging this patch.

PR #243, opened 2023-08-17
fix wrong color for grid line foreground

Destrolaric:
We found out that the wrong color was used for line foreground when grid is shown


azoitl:
Although I was a bit distracted with the "new" grid color you introduced with the color patch I found in the end it gave more contrast to our diagrams. However I also understand that it can be distracting. 

With your PR you are now setting it to just grey. Is this suitable for switching between light and dark theme?

Destrolaric:
@azoitl This impacts only the light theme. I accidentally changed it when adding the black piece. Dark color adds to much contrast on the diagrams.

azoitl:
> @azoitl This impacts only the light theme. I accidentally changed it when adding the black piece. Dark color adds to much contrast on the diagrams.

@Destrolaric ah sorry I missed that. Looks good to me.

PR #244, opened 2023-08-18
Code Cleanup - Use pattern matching for instanceof operator

ptziegler:
This replaces code of the form:
```
if (o instanceof <type>) {
  <type> t = (<type>) o;
  ...
}
```

with:
```
if (o instanceof <type> t) {
  ...
}
```

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/actions/CopyTemplateAction.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/actions/CopyTemplateAction.java
@@ -81,9 +81,8 @@ public class CopyTemplateAction extends WorkbenchPartAction implements ISelectio
 	@Override
 	public void selectionChanged(SelectionChangedEvent event) {
 		ISelection s = event.getSelection();
-		if (!(s instanceof IStructuredSelection))
+		if (!(s instanceof IStructuredSelection selection))
 			return;
-		IStructuredSelection selection = (IStructuredSelection) s;
 		template = null;
 		if (selection != null && selection.size() == 1) {
 			Object obj = selection.getFirstElement();
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/EntityConnectionData.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/EntityConnectionData.java
@@ -42,10 +42,9 @@ public final class EntityConnectionData {
 	 */
 	@Override
 	public boolean equals(Object obj) {
-		if (!(obj instanceof EntityConnectionData)) {
+		if (!(obj instanceof EntityConnectionData that)) {
 			return false;
 		}
-		EntityConnectionData that = (EntityConnectionData) obj;
 		return (this.source.equals(that.source) && this.dest.equals(that.dest));
 	}
 
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ChopboxAnchor.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ChopboxAnchor.java
@@ -107,8 +107,7 @@ public class ChopboxAnchor extends AbstractConnectionAnchor {
 	 */
 	@Override
 	public boolean equals(Object obj) {
-		if (obj instanceof ChopboxAnchor) {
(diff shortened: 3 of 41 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/EntityConnectionData.java
```diff
@@ -42,10 +42,9 @@ public EntityConnectionData(Object source, Object dest) {
 	 */
 	@Override
 	public boolean equals(Object obj) {
-		if (!(obj instanceof EntityConnectionData)) {
+		if (!(obj instanceof EntityConnectionData that)) {
```
azoitl, line 45: I thought that is only available inside of the if?

org.eclipse.gef/src/org/eclipse/gef/ui/actions/CopyTemplateAction.java
```diff
@@ -81,9 +81,8 @@ public void run() {
 	@Override
 	public void selectionChanged(SelectionChangedEvent event) {
 		ISelection s = event.getSelection();
-		if (!(s instanceof IStructuredSelection))
+		if (!(s instanceof IStructuredSelection selection))
```
azoitl, line 84: I thought selection is only valid inside of the if

org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/EntityConnectionData.java
ptziegler: According to their [documentation](https://docs.oracle.com/en/java/javase/16/language/pattern-matching-instanceof-operator.html#GUID-843060B5-240C-4F47-A7B0-95C42E5B08A7) that's a valid syntax. They even use it in one of their examples ;)
image

This is the resulting byte code. From what I understand, Java internally creates a "that" variable which is initialized in L2, i.e. the if condition and then lives until the end of the method..
```
public boolean equals(java.lang.Object arg0) { //(Ljava/lang/Object;)Z
	<localVar:index=0 , name=this , desc=Lorg/eclipse/zest/core/viewers/EntityConnectionData;, sig=null, start=L0, end=L5>
	<localVar:index=1 , name=obj , desc=Ljava/lang/Object;, sig=null, start=L0, end=L5>
	<localVar:index=2 , name=that , desc=Lorg/eclipse/zest/core/viewers/EntityConnectionData;, sig=null, start=L3, end=L5>

	L0 {
		aload 1
		instanceof org/eclipse/zest/core/viewers/EntityConnectionData
		ifne L2
	}
	L1 {
		iconst_0
		ireturn
	}
	L2 {
		f_new (Locals[2]: org/eclipse/zest/core/viewers/EntityConnectionData, java/lang/Object) (Stack[0]) 
		aload 1
		checkcast org/eclipse/zest/core/viewers/EntityConnectionData
		astore 2
	}
	L3 {
		aload 0 // reference to self
		getfield org/eclipse/zest/core/viewers/EntityConnectionData.source:java.lang.Object
		aload 2
		getfield org/eclipse/zest/core/viewers/EntityConnectionData.source:java.lang.Object
		invokevirtual java/lang/Object.equals(Ljava/lang/Object;)Z
		ifeq L4
		aload 0 // reference to self
		getfield org/eclipse/zest/core/viewers/EntityConnectionData.dest:java.lang.Object
		aload 2
		getfield org/eclipse/zest/core/viewers/EntityConnectionData.dest:java.lang.Object
		invokevirtual java/lang/Object.equals(Ljava/lang/Object;)Z
		ifeq L4
		iconst_1
		ireturn
	}
	L4 {
		f_new (Locals[3]: org/eclipse/zest/core/viewers/EntityConnectionData, java/lang/Object, org/eclipse/zest/core/viewers/EntityConnectionData) (Stack[0]) 
		iconst_0
		ireturn
	}
	L5 {
	}
}
```

ptziegler: Though I must admit, those patter matching rules still look like magic to me...

azoitl: I have to agree here. Thx for the clarification. Learned something new about Java today.

PR #246, opened 2023-08-18
Fix: Regressions caused by cleaning up BreakCycle & PopulateRanks

ptziegler:
The for-each loop inside the BreakCycle method used to iterate over all graph nodes. This was wrongfully changed to a local variable which is initialized as an empty list.

The for-each loop inside PopulateRanks iterates over all graph nodes and calls VirtualNodeCreation internally. The constructor of this class, however, may add nodes to the graph, leading to a
ConcurrentModifcationException. To avoid this problem, we iterate of a local copy of the graph nodes instead.

Amends 31a70127f763d40fd1f6eb6a56de9158cf9538ce

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/graph/test/DirectedGraphLayoutTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/graph/test/DirectedGraphLayoutTest.java
@@ -1,0 +1,102 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Patrick Ziegler and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.draw2d.graph.test;
+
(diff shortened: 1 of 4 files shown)


ptziegler:
@azoitl 

I just went over the changes from #234 again and in the future, those should really be separated into separate pull requests, simply because of how distracting some of the changes are...
At first glance there were:
- Excessive whitespace changes
- Override annotations and static modifiers being added
- Parenthesis being added
- Changes from Node.getNode() to Node.get()
- The for-loop changes

If all of those were separated, I think it'd have been much easier to figure out what went wrong. Just some food for thought :sweat_smile:

azoitl:
> @azoitl
> 
> I just went over the changes from #234 again and in the future, those should really be separated into separate pull requests, simply because of how distracting some of the changes are... At first glance there were:
> 
>     * Excessive whitespace changes
> 
>     * Override annotations and static modifiers being added
> 
>     * Parenthesis being added
> 
>     * Changes from Node.getNode() to Node.get()
> 
>     * The for-loop changes
> 
> 
> If all of those were separated, I think it'd have been much easier to figure out what went wrong. Just some food for thought 😅

@ptziegler I think you are totally right. My naive attempt of thinking to clean-up files that are touched automatically does not seem to work. At least for the Override annotation I found a better solution: I used a project wide clean-up feature from Eclipse IDE: PR #241. But for others I'm not yet sure. Happy for any ideas.

azoitl:
@ptziegler thx for the fast reaction, the fix, and a test for it!

PR #247, opened 2023-08-18
Typify internal list in DirectGraphLayout and convert to Deque

ptziegler:
By using a deque, we have access to both an iterator() and an descendingIterator() method. This allows us to iterate over the collection in both directions.

### DIFF
--- a/org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/algorithms/DirectedGraphLayoutAlgorithm.java
+++ b/org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/algorithms/DirectedGraphLayoutAlgorithm.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2005, CHISEL Group, University of Victoria, Victoria, BC, Canada.
+ * Copyright 2005, 2023 CHISEL Group, University of Victoria, Victoria, BC, Canada.
  * All rights reserved. This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License v1.0 which
  * accompanies this distribution, and is available at
@@ -10,9 +10,9 @@
 package org.eclipse.zest.layouts.algorithms;
 
 import java.lang.reflect.Field;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.List;
 
 import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.draw2d.graph.DirectedGraph;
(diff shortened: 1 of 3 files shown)


ptziegler:
#155 

Switching to a Deque should be safe, given that this field is a package-private class.

org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/algorithms/DirectedGraphLayoutAlgorithm.java
```diff
@@ -34,20 +34,14 @@ public void visit(DirectedGraph graph) {
 				field = DirectedGraphLayout.class.getDeclaredField("steps");
```
azoitl, line 34: should we add a protected getter to the steps so that we don't need reflection here?

ptziegler: I'm not sure if that would really work... "steps" stores objects of type GraphVisitor, which is a package-private class in draw2d. With a getter, this class would now be exposed.
As an example, the method call in DirectedGraphLayoutAlgorithm would still have to look like `Deque<?> parts = getParts()`, simply because GraphVisitor is not visible.

azoitl: I meant to add a protected getSteps method to DirectedGraph Layout so that Children like this ExtenddedDirectGraphLayout could access the steps. But I'm not sure if we want to expose it like that. As you say this has implications.

azoitl:
@ptziegler have you seen my latest comment? It would be great if we could merge that commit somewhere today or early tomorrow as we have M3 deadline.

org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/algorithms/DirectedGraphLayoutAlgorithm.java
ptziegler: Let me rephrase it: What should getSteps() return? It can't be Deque<GraphVisitor> because the GraphVisitor is package-private class. So the only other alternatives are Deque<?> or Deque<Object>, both of which I'm not really happy with...

azoitl: Ah now I get it. Yes a getter makes here then no sense at all.

azoitl:
No problem. Thanks for the feedback. Will merge it now.

Issue #248, opened 2023-08-21
Line width not set on Windows Hi-DPI


Phillipus:
I'm on Windows 11 using a hi-dpi monitor at 200% scaling and I notice that some of the figures in our RCP app are not being drawn properly.

Here's an example snippet:

```java
import org.eclipse.draw2d.Figure;
import org.eclipse.draw2d.FigureCanvas;
import org.eclipse.draw2d.Graphics;
import org.eclipse.draw2d.IFigure;
import org.eclipse.draw2d.geometry.Rectangle;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class LineWidth {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setLayout(new FillLayout());
        shell.setSize(300, 300);

        FigureCanvas fc = new FigureCanvas(shell);
        
        IFigure rectFigure = new Figure() {
            @Override
            protected void paintFigure(Graphics graphics) {
                graphics.setAdvanced(true);
                Rectangle rect = bounds.getCopy();
                
                rect.shrink(10, 10);
                graphics.setLineWidth(1);
                graphics.drawRectangle(rect);
                
                rect.shrink(20, 20);
                graphics.setLineWidth(2);
                graphics.drawRectangle(rect);
            }
        };
        
        rectFigure.setBounds(new Rectangle(10, 10, 250, 150));
        fc.getLightweightSystem().getRootFigure().add(rectFigure);

        shell.open();
        
        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }
    }
}
```

This is the result, both rectangles' line width are 1 pixel.

![Image 001](https://github.com/eclipse/gef-classic/assets/600504/cf6c9d41-5254-43c8-9440-3682f29e33b4)

Here's what it should look like:

![Image 002](https://github.com/eclipse/gef-classic/assets/600504/78643735-1745-4c34-bf7a-65e9ea44644d)

If you comment out the line `graphics.setAdvanced(true);` it is OK.

 I narrowed down where this is happening to:

https://github.com/eclipse/gef-classic/blob/ff0b9c31c348e9090cb8deea53b41b15879fe50e/org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java#L308-L318

[...]

PR #249, opened 2023-08-21
Fix wrong line width on Windows hi-dpi

Phillipus:
- On Windows hi-dpi calling GC#setLineAttributes will mean that lineattributes.width is increased by calling DPIUtil.autoScaleUp so we clone lineAttributes

- See https://github.com/eclipse/gef-classic/issues/248


azoitl:
It didn't change anything for me on non scaled displays. So I think it is fine to merge.

PR #250, opened 2023-08-21
Avoid excessive repaint operations when figure is not modified

ptziegler:
When updating the border or the background/foreground color of a figure,
it may happen what we whole figure is repainted, even though it hasn't
been modified.
Additional checks have been added to check for equality in the setter
methods, to limit the amount of times repaint() and revalidate() is
called.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/Draw2dTestSuite.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/Draw2dTestSuite.java
@@ -43,6 +43,7 @@ import org.junit.runners.Suite;
 	PrecisionPointTest.class,
 	PrecisionRectangleTest.class,
 	ThumbnailTest.class,
+	FigurePaintingTest.class,
 	FigureUtilitiesTest.class,
 	RectangleTest.class,
 	ColorConstantTest.class,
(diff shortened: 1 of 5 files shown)


ptziegler:
Why were none of the tests executed? :thinking:

ptziegler:
I'll try to create some tests today, to make sure that everything is still running as it should.

ptziegler:
@azoitl Your thoughts on this one?

I'm currently in the process of cleaning up the duplicate code in our own Figure implementation and noticed that we have some additional checks in our setBounds() and setForeground/BackgroundColor() methods that might be worth being merged into GEF itself.

azoitl:
@ptziegler during debugging GEF applications a always noticed lots of notifications on different levels. I mostly thought it is due to my lack of understanding or correctly using GEF. So bringing down repaints and notifications is something that seems desireable for me. If @Phillipus also agrees I would merge it.

Phillipus:
> If @Phillipus also agrees I would merge it.

It looks good to me, and there's a lot of work gone into creating new tests.

(In fact, thanks to @ptziegler it reminds me that I need to make more use of `Objects.equals()` in my own code.)

azoitl:
> > If @Phillipus also agrees I would merge it.
> 
> It looks good to me, and there's a lot of work gone into creating new tests.

Thx for the feedback. Yes I agree the tests are amazing.
> 
> (In fact, thanks to @ptziegler it reminds me that I need to make more use of `Objects.equals()` in my own code.)

Me too.

PR #251, opened 2023-08-23
Remove custom rebase action from workflow.

ptziegler:
This is now natively supported by GitHub.


ptziegler:
i.e. we now have this big button here when the branch is not up-to-date. I also don't think it was ever actively used to begin with...
image

PR #253, opened 2023-08-24
isHighlighted method for GraphItems

BauePhil:
Most of the functionality is already implemented, similar to the isVisible methods so I:
- added a isHighlighted method to the super class GraphItem, with a default implementation as to not create problems downstream
- added @Override notations to already existing methods in GraphNode and GraphConnection

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
@@ -448,6 +448,7 @@ public class GraphConnection extends GraphItem {
 	 * Returns true if this connection is highlighted, false otherwise
 	 * 
 	 * @return boolean state of highlight
+	 * @since 1.9
 	 */
 	@Override
 	public boolean isHighlighted() {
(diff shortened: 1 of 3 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
```diff
@@ -447,8 +447,9 @@ public void unhighlight() {
 	/**
 	 * Returns true if this connection is highlighted, false otherwise
 	 * 
-	 * @return
+	 * @return boolean state of highlight
 	 */
```
azoitl, line 451: when testing this on my setup, the API analysis tool requires now a since tag. Not sure why. I expect because it is now API. Can you please add one.

azoitl (changes requested):

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
BauePhil: done

azoitl: thx for fixing

azoitl (approved):

PR #257, opened 2023-08-24
added label initialization for graphconnection

BauePhil:
if the already exposed createFigure method in GraphConnection is overridden, the connectionLabel will never be initialized and thus lead to a null pointer in the doUpdateFigure method.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
@@ -583,7 +583,10 @@ public class GraphConnection extends GraphItem {
 		connectionShape.setLineStyle(getLineStyle());
 
 		if (this.getText() != null || this.getImage() != null) {
-			// Label l = new Label(this.getText(), this.getImage());
+			// label can be null if createFigure has been overridden
+			if (this.connectionLabel == null) {
+				this.connectionLabel = new Label();
+			}
 			if (this.getImage() != null) {
 				this.connectionLabel.setIcon(this.getImage());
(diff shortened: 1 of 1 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
```diff
@@ -583,7 +583,10 @@ private void doUpdateFigure(Connection connection) {
 		connectionShape.setLineStyle(getLineStyle());
 
 		if (this.getText() != null || this.getImage() != null) {
```
azoitl, line 585: Wouldn't it be more correct to add a this.connectionLabel != null into this if instead of just creating a label that is not shown and not used?

BauePhil: i believe the label being null here does say that its not used. Its private so it cannot be initialized by the overridden createFigure method, but could still be used when e.g. the text or the image are set (i.e., the code after the check).

azoitl: I did a second check where connectionLabel is used. And it is only used in the doCreateFigure and here in this method. With your change you are creating a label but never using it. Therefore you are just wasting memory. By simply adding a null check to the if the whole block (which could be moved to an one method). would not be executed. This solves the problem of connectionLabel beeing null and does not waste memory.

BauePhil: Ah I see now, you are right.

PR #258, opened 2023-08-25
Pull getLocation() method from Figure to IFigure interface.

ptziegler:
If a method only works on the IFigure interface, it currently has to either cast the object to a Figure, in order to be able to call getLocation() or instead go via getBounds().getLocation().

To avoid this issue, this method has been pulled to the implemented interface. A default implementation is used to for backwards compatibility. The implementation inside the Figure is not removed, in order to keep the 'final' modifier, but internally only calls the default implementation.

The interface already has a setter method for the location. Also providing a getter only seems natural.


azoitl:
This is a very nice fix.

PR #265, opened 2023-09-26
Use latest GEF Classic target platform for building GEF Classic

azoitl:
@ptziegler I think this is what you had in mind or?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -223,7 +223,7 @@
 							<includePackedArtifacts>false</includePackedArtifacts>
 							<!-- add target file content to target platform -->
 						    <target>
-								<file>../target-platform/2022-12.target</file>
+								<file>../target-platform/GEF_classic.target</file>
 							  </target>
 							<environments>
 								<environment>
@@ -260,7 +260,7 @@
 						<includePackedArtifacts>false</includePackedArtifacts>
 						<!-- add target file content to target platform -->
 						<target>
-							<file>../target-platform/2022-12.target</file>
+							<file>../target-platform/GEF_classic.target</file>
 						  </target>
 						<environments>
 							<environment>


ptziegler:
That's exactly what I meant

org.eclipse.draw2d-feature/feature.xml
```diff
@@ -15,7 +15,7 @@
       version="3.18.0.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
-      license-feature-version="1.0.1">
```
ptziegler, line 18: I assume now 2.x is used? Wouldn't this mean the features are now licensed under EPLv2?

azoitl: I must confess I'm not perfectly sure what was wrong here. I needed to change that to get it compiling and I looked at the Eclipse platform how they have it. 

Generally I anyhow think that we should ugprade GEF Classic to EPLv2. Maybe that would be a next step.

ptziegler: Found it... The license feature is declared [here](https://github.com/eclipse-cbi/epl-license-feature) and as suspected, 1.x is EPLv1 and 2.x is EPLv2.

> I'm not perfectly sure what was wrong here

My guess is that some dependencies in the SDK got updated, so that now a newer version of this feature (i.e. 2.x) is contributed via the target platform. Because GEF still expects 1.x, we get an error.

I assume the dirty solution would be to add the old license feature as an explicit dependency, but I haven't found a proper update site yet...

> Generally I anyhow think that we should ugprade GEF Classic to EPLv2.

I wonder... According to the bundles/features, GEF is still using EPLv1. But according to the [license](https://github.com/eclipse/gef-classic/blob/master/LICENSE), GEF is using EPLv2. So I'm not sure what to make of it?

In principle I agree, moving to the newer license should be the end goal. But that is some legal ground I don't really want to touch.  :/

azoitl: @ptziegler I checked with EMO and we can just upgrade if we want: https://www.eclipse.org/legal/epl-2.0/faq.php#h.tci84nlsqpgw

I would announce this on our mailing list and if I don't get negative feedback would create a dedicated issue in github and start the migration process in the next weeks.

ptziegler: Sounds like a solid plan :+1:

azoitl: I created #266 for tracking the migration to EPLv2. Should we merge this issue now or after we finished migration?

azoitl: @ptziegler with the upgrade to EPLv2 in PR #267  I think this PR can now also be merged or?

ptziegler: Absolutely! The build is looking good and I don't expect any other problems.

ptziegler (approved):

PR #267, opened 2023-10-15
Fixes #266  EPLv2 migration

### LINKED ISSUE #266: Move to EPLv2
Upgrade all license refernes to EPLv2
This inlcudes the 

 - copyright notices. Only the text may be changed the copright staments themselve must not be changed.
 - license files
 - about.htm
 - readme

azoitl:
All License files and copyright headers are now migrated to EPLv2.

@ptziegler @Phillipus thx in advance for cross checking.


org.eclipse.draw2d-feature/feature.xml
```diff
@@ -15,7 +17,7 @@
       version="3.18.0.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
-      license-feature-version="1.0.1">
+      license-feature-version="0.0.0">
```
ptziegler, line 20: Wouldn't it make sense to instead set the version to 2.0.1? Otherwise we'd automatically opt-in, in case the Eclipse ever decides to migrate to a newer version again.

azoitl: I was not sure. I checked what the Eclipse Platform has as value and did the same.

azoitl: This is what I thought so too. And if the platform does it we can not be that wrong.

ptziegler (approved):
That's all I wanted to clarify. The change itself is too much for a proper review, but I assume you effectively only changed the license header, so I think that's fine.

azoitl:
> That's all I wanted to clarify. The change itself is too much for a proper review, but I assume you effectively only changed the license header, so I think that's fine.

Yes this PR is a bit heavy for that. I mostly did search replace and double check results. We know have no mentioning of EPL V1.0 any more. Also I preserved all copyright dates, copyright holders and contributors. As I did that already once for Eclipse 4diac I think we are save here. Therefore I would merge and move on.

azoitl:
Fixes #266

PR #268, opened 2023-10-18
refactor: Java Security Ultimate Security Repo Scanner 2023

caytec:
Use this link to re-run the recipe: https://app.moderne.io/recipes/builder/TkgUEiqd7?organizationId=RWNsaXBzZSBGb3VuZGF0aW9u

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/graph/RankSorter.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/graph/RankSorter.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package org.eclipse.draw2d.graph;
 
+import java.security.SecureRandom;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Random;
@@ -22,7 +23,7 @@ import java.util.Random;
  */
 class RankSorter {
 
-	Random flipflop = new Random(3);
+	Random flipflop = new SecureRandom();
 	Node node;
 	double rankSize, prevRankSize, nextRankSize;
 	int currentRow;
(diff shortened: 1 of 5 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/graph/RankSorter.java
```diff
@@ -22,7 +23,7 @@
  */
 class RankSorter {
 
-	Random flipflop = new Random(3);
```
ptziegler, line 25: SecureRandom can be significantly slower than the plain old Random. Do we really need a cryptographically strong generator here?

ptziegler:
Thank you for your contribution! It seems like you first need to sign the [ECA](https://www.eclipse.org/legal/ECA.php) before we can accept this contribution, .

org.eclipse.draw2d/src/org/eclipse/draw2d/graph/RankSorter.java
Destrolaric: By the looks of it, this change is redundant because randomization here is not used for any security purposes and possibly slows the overall performance of the sorter.

caytec: Optimization cases which can risk in outages are also made in scope of understanding the data through the lenses of great mixture of AI and software engineering. That's the though worth taking to considerations, as sometimes he can be kinda silly and send either very serious stuff or very chill like that ;)

Destrolaric: I think only zip folder change can be viable here, could you update the pr to not include SecureRandom changes?

ptziegler:
@caytec Do you want to continue with this?

PR #269, opened 2023-10-22
Compile and test GEF against both Java 17 and Java 21.

ptziegler:
With Java 21 being the current LTS, we should ensure that GEF is forward compatible.


azoitl:
this is a very good idea.

ptziegler:
Given that this doesn't modify the source code in any way, I'll just merge it.

azoitl:
> Given that this doesn't modify the source code in any way, I'll just merge it.

Sure I wanted to merge, but got distracted in my day job.

PR #273, opened 2023-10-27
Changed shared cursors from bmps and masks to single pngs

### LINKED ISSUE #208: [GTK] Cursor created with GIF format images renders incorrectly
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=575448 reported a while ago on SWT by @Phillipus.

For example on the GMF Runtime'e GeoShape example, when using the "Rectangle Types" tool:

https://github.com/eclipse/gef-classic/assets/10608/85d21531-d118-4d5a-bfde-636c890cb416

A Sirius user reported this [recently](https://www.eclipse.org/forums/index.php?t=msg&th=1113080&goto=1859256&#msg_1859256) which finally triggered me to investigate.
I confirm that the workaround mentiond by @Phillipus in the bugzilla (using PNG versions of the cursor and mask) fixes the problem for me.

I replaced [`Tree_Add.gif`](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/internal/icons/Tree_Add.gif) and [`Tree_Add_Mask.gif`](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/internal/icons/Tree_Add_Mask.gif) with the PNG versions from the archive on the bugzilla, and this fixes the issue for me (for this particular cursor) on the example above.

Note that there is something fishy in how [`SharedCursors`](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/SharedCursors.java) creates its cursors: it calls `createCursor("icons/Tree_Add_Mask.gif", "icons/Tree_Add.gif")`, but `createCursor(String sourceName, String maskName)` expects the source first and the mask second.

azoitl:
Thx to Phillipus <p.beauvoir@dadabeatnik.com> for providing the cursor images.

This fixes #208.

@pcdavid @Phillipus thx for reviewing.


pcdavid:
I've not tested all the cursors (not even sure how to trigger most of them), but on the original scenario from the video on #208, this fixes the issue (Linux, X11, 4K screen).

lredor:
> I've not tested all the cursors (not even sure how to trigger most of them), but on the original scenario from the video on #208, this fixes the issue (Linux, X11, 4K screen).

I've tested the same scenario on Windows11: The behavior was OK before the fix, as the problem was only on Linux, and it is always OK with this PR.

azoitl:
Thanks for the feedback. I also tested all cursors under Debian/Linux with the Logic Editor example. So I think we are fine to merge.

PR #275, opened 2023-11-03
Typify GroupRequest and SimpleFactory classes

ptziegler:
For the GroupRequest class, we store the list of edit-parts as a List<? extends EditPart>, in order to support casting it to e.g. a List<CustomEditPart>.

For the SimpleFactory class, we replace the deprecated Class.newInstance() call with Constructor.newInstance(), to keep the compile-time exception checking (see JavaDoc for more details).

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/requests/SimpleFactory.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/requests/SimpleFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the 
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


org.eclipse.gef/src/org/eclipse/gef/requests/SimpleFactory.java
```diff
@@ -21,14 +21,14 @@
  */
 public class SimpleFactory implements CreationFactory {
```
azoitl, line 22: should we then turn the whole class into a generic? This should not change any client code and future code could better use it.

ptziegler: Good point... and done. I also thought about making the whole CreationFactory generic, but I think that'd require quite a lot of changes to resolve all warnings it would introduce...

ptziegler:
I've also just made the list that is created by `setEditParts(EditPart)` mutable again. Otherwise it might cause exceptions if users try to modify the list that is then returned by `getEditParts()`.

org.eclipse.gef/src/org/eclipse/gef/requests/SimpleFactory.java
azoitl: Maybe that is then something for another PR. I general I like these kinds of warnings as they point us to where we can now improve GEF Classic.

PR #274, opened 2023-11-02
Cleaned Zest Algorithms

azoitl:
In order to reduce our warning count and increase typed generic use (Issue #155)  this PR typed all generics in the Zest Algorithms package. This reduced the warning count by ap. 200.

PR #277, opened 2023-11-03
Remove trailing whitespaces via automatic code cleanup.

ptziegler:
The same operation is done whenever a file is edited and saved, leading to countless white-space changes, obscuring the actual change of a commit. The hope is that this'll make future commits much cleaner.


ptziegler:
You can look at e.g. the SpringLayoutAlgorithm change in #274, which makes it a lot more tedious to see the real change.

azoitl:
This is a very smart idea. Will merge as soon as the checks pass.

PR #279, opened 2023-11-04
[Modernize Code] Utilize new generic SimpleFactory

azoitl:
Based on the PR https://github.com/eclipse/gef-classic/pull/275 from @ptziegler all usages of SimpleFactory have been rechecked and if possible the typed version is used.

### DIFF
--- a/org.eclipse.gef.examples.shapes/src/org/eclipse/gef/examples/shapes/ShapesEditor.java
+++ b/org.eclipse.gef.examples.shapes/src/org/eclipse/gef/examples/shapes/ShapesEditor.java
@@ -163,7 +163,7 @@ public class ShapesEditor extends GraphicalEditorWithFlyoutPalette {
 		return new TemplateTransferDropTargetListener(getGraphicalViewer()) {
 			@Override
 			protected CreationFactory getFactory(Object template) {
-				return new SimpleFactory((Class) template);
+				return new SimpleFactory<>((Class<?>) template);
 			}
 		};
 	}
@@ -254,8 +254,9 @@ public class ShapesEditor extends GraphicalEditorWithFlyoutPalette {
 
 	@Override
 	public <T> T getAdapter(final Class<T> type) {
-		if (type == IContentOutlinePage.class)
+		if (type == IContentOutlinePage.class) {
 			return type.cast(new ShapesOutlinePage(new TreeViewer()));
+		}
 		return super.getAdapter(type);
 	}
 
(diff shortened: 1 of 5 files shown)


org.eclipse.gef.examples.shapes/src/org/eclipse/gef/examples/shapes/ShapesEditor.java
```diff
@@ -254,8 +254,9 @@ public void execute(final IProgressMonitor monitor) {
```
ptziegler, line 254: Sadly, GitHub doesn't let me add comments to unmodified code... But in line 166, SimpleFactory is still a raw type. Given that we work on objects there, I'd suggest something like:
```
return new SimpleFactory<>((Class<?>) template);
```

org.eclipse.gef/src/org/eclipse/gef/dnd/TemplateTransferDropTargetListener.java
```diff
@@ -73,13 +73,15 @@ protected final CreateRequest getCreateRequest() {
...
-			return ((CreationFactory) template);
-		} else if (template instanceof Class) {
-			return new SimpleFactory((Class) template);
-		} else
-			return null;
+		if (template instanceof CreationFactory creationFactory) {
+			return creationFactory;
+		}
+		if (template instanceof Class clazz) {
+			return new SimpleFactory(clazz);
```
ptziegler, line 82: Similar to the case in ShapeEditor, I'd use something like this to resolve the warning:
```
if (template instanceof Class<?> clazz) {
	return new SimpleFactory<>(clazz);
}
```

ptziegler (changes requested):

ptziegler:
My apologies for not checking the compiler warnings :sweat_smile: I completely forgot to check whether GEF is using them internally...

org.eclipse.gef.examples.shapes/src/org/eclipse/gef/examples/shapes/ShapesEditor.java
azoitl: I was not sure if Ii would do it. So thx for the feedback fixed it.

org.eclipse.gef/src/org/eclipse/gef/dnd/TemplateTransferDropTargetListener.java
azoitl: Done

azoitl:
> My apologies for not checking the compiler warnings 😅 I completely forgot to check whether GEF is using them internally...

No Problem at all. Was a very nice exercise for me to experiment with your change.

ptziegler (approved):

PR #280, opened 2023-11-04
[Modernize Code] typify the selection list of EditPart viewer

azoitl:
Supports #155.

Tested with:
  - GMF Runtime: no changes necessary
  - 4diac IDE: has ugly casts that need to be fixed

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
@@ -51,7 +51,6 @@ import org.eclipse.gef.MouseWheelHandler;
 import org.eclipse.gef.RequestConstants;
 import org.eclipse.gef.Tool;
 import org.eclipse.gef.commands.Command;
-import org.eclipse.gef.commands.CommandStackEvent;
 import org.eclipse.gef.commands.CommandStackEventListener;
 import org.eclipse.gef.editparts.LayerManager;
 
@@ -110,24 +109,28 @@ public abstract class AbstractTool extends org.eclipse.gef.util.FlagSupport impl
 	 *
 	 * @deprecated Use {@link SWT#BUTTON1} instead.
 	 */
+	@Deprecated
 	protected static final int MOUSE_BUTTON1 = SWT.BUTTON1;
 	/**
 	 * constant used for mouse button 2.
 	 *
 	 * @deprecated Use {@link SWT#BUTTON2} instead.
 	 */
+	@Deprecated
 	protected static final int MOUSE_BUTTON2 = SWT.BUTTON2;
 	/**
 	 * constant used for mouse button 3.
 	 *
 	 * @deprecated Use {@link SWT#BUTTON3} instead.
 	 */
+	@Deprecated
 	protected static final int MOUSE_BUTTON3 = SWT.BUTTON3;
 	/**
 	 * constant used to indicate any of the mouse buttons.
 	 *
 	 * @deprecated Use {@link SWT#BUTTON_MASK} instead.
 	 */
+	@Deprecated
 	protected static final int MOUSE_BUTTON_ANY = SWT.BUTTON_MASK;
 
 	/**
@@ -193,11 +196,9 @@ public abstract class AbstractTool extends org.eclipse.gef.util.FlagSupport impl
 	private int accessibleStep;
 	private Command command;
 
-	private CommandStackEventListener commandStackListener = new CommandStackEventListener() {
-		@Override
-		public void stackChanged(CommandStackEvent event) {
-			if (event.isPreChangeEvent())
-				handleCommandStackChanged();
+	private final CommandStackEventListener commandStackListener = event -> {
+		if (event.isPreChangeEvent()) {
+			handleCommandStackChanged();
 		}
 	};
 	private Input current;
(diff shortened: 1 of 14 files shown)


org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
```diff
@@ -1113,30 +1136,38 @@ public void mouseMove(MouseEvent me, EditPartViewer viewer) {
...
 				handleButtonUp(2);
-			if (b3)
+			}
+			if (b3) {
 				handleButtonUp(3);
-			if (b4)
+			}
+			if (b4) {
 				handleButtonUp(4);
-			if (b5)
```
ptziegler, line 1124: Do we really use an if-else chain for the buttons 1 to 5...?

It doesn't need to be part of this change, but perhaps it's worth adding a getMouseButton() to the Input class.

org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
```diff
@@ -363,16 +365,17 @@ public Map getEditPartRegistry() {
...
 				return getContents();
-			else
-				return getRootEditPart();
+			}
+			return getRootEditPart();
 		}
-		List selection = getSelectedEditParts();
-		return (EditPart) selection.get(selection.size() - 1);
+		List<? extends EditPart> selection = getSelectedEditParts();
+		return selection.get(selection.size() - 1);
```
ptziegler, line 378: I can't wait for Java 21 where we can simply call selection.getLast() :sweat_smile:

org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
```diff
@@ -181,16 +182,25 @@ protected void fireSelectionChanged() {
 		showSelectionInTree();
 	}
 
+	/**
+	 * @since 3.16
+	 */
+	@Override
+	public Tree getControl() {
+		return (Tree) super.getControl();
+	}
```
ptziegler, line 191: Perhaps it would be better to make the AbstractEditPartViewer generic with type \<T extends Control\>?

For the GraphicalViewer, T is a Canvas and for the TreeViewer, it's a Tree.

org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
```diff
@@ -223,31 +234,22 @@ public void widgetDefaultSelected(SelectionEvent e) {
...
-		TreeItem[] treeItems = new TreeItem[treeParts.size()];
-		for (int i = 0; i < treeParts.size(); i++) {
-			TreeEditPart part = (TreeEditPart) treeParts.get(i);
-			treeItems[i] = (TreeItem) part.getWidget();
 		}
-		tree.setSelection(treeItems);
+		TreeItem[] treeItems = getSelectedEditParts().stream().map(TreeEditPart.class::cast)
+				.filter(p -> p.getWidget() instanceof TreeItem).map(TreeEditPart::getWidget).map(TreeItem.class::cast)
+				.toArray(TreeItem[]::new);
+		getControl().setSelection(treeItems);
```
ptziegler, line 252: I'm not sure if this is really more readable... Perhaps splitting up the method calls into more lines makes it cleaner?

org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
azoitl: I saw that two and thought this is a target for a next PR.

org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
azoitl: me too

org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
azoitl: Not sure. We only have two three viewers and with that style it is rather explicit that this viewer will always have a Tree.

org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
azoitl: I'm now rather used to the stream style but I can understand the problems with this here. Let me try reformating it.

org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
ptziegler: Fair enough. It would probably also require  changes to the EditPartViewer interface, so it's probably safer to keep it like this.

org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
azoitl: Hmh it looks like Eclipse's format is condensing it and I didn't find any option on controlling formatting of stream. 

The reason I changed is is that it is more efficient and for me more maintainable then the previous code.  But I'm happy to change it back into something with loops if you think I overdid it.

ptziegler: > Hmh it looks like Eclipse's format is condensing it and I didn't find any option on controlling formatting of stream.

You can turn off the formatter by using these special comments (which needs to be enabled in the formatter profile):

> // @formatter:off
> ...
> // @formatter:on

Alternatively, you can just add empty comments at the end of a line:

> getSelectedEditParts() //
> 		.stream() //
>		.map(TreeEditPart.class::cast) //
>		.filter(p -> p.getWidget() instanceof TreeItem) //
> ...

> The reason I changed is is that it is more efficient and for me more maintainable then the previous code. But I'm happy to change it back into something with loops if you think I overdid it.

I'm fine with using Streams. It's just the formatting that's a little off-putting :sweat_smile:

azoitl: I know the formatter on/off thingy but it seems we have it not yet enabled. So I used the comment approach which I didn't know existed.

ptziegler (approved):

azoitl:
@ptziegler thx for reviewing.

PR #284, opened 2023-11-10
[Modernize Code] Type the return value of GraphicalEditPolicy getHost

azoitl:
Java now allows overriding return values. Using this for GraphicalEditPolicy.getHost() reduces cast requirements and makes code clearer.


azoitl:
As tomorrow is deadline for RC1 I would love to merge it. I would merge it if no one objects.

ptziegler (approved):
From what it looks like, those changes won't modify the API, so I think it's safe to commit them this late into development.

azoitl:
Then I'm merging it. Thx for the feedback.

PR #287, opened 2023-11-11
[Modernize Code] clean-ups in AbstractEditPartViewer

azoitl:
Addresses #155

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
@@ -51,6 +51,7 @@ import org.eclipse.gef.EditDomain;
 import org.eclipse.gef.EditPart;
 import org.eclipse.gef.EditPartFactory;
 import org.eclipse.gef.EditPartViewer;
+import org.eclipse.gef.GraphicalEditPart;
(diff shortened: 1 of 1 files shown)


azoitl:
As tomorrow is deadline for RC1 I would love to merge it. As its mostly internal clean-ups I would merge it if no one objects.

ptziegler (commented):
The change itself looks fine. _However_ keep in mind that this might cause compilation errors in downstream projects.

For the RC1, I had to [adjust](https://github.com/eclipse-windowbuilder/windowbuilder/commit/45df1ea373d7dc40aa7ff554e05f3e3edcb02711) the WindowBuilder code to resolve compilation errors due to https://github.com/eclipse/gef-classic/commit/0860312757105aedf20cbfa17b3aa4c219776971 

In order to avoid any unpleasant surprises directly after the release, this would now require downstream projects to build/test against the upcoming RC1, which I'm not sure everyone is doing.

Granted, primGetSelectedEditParts() is a protected method, so the chances of complications are probably low, but I think that's something we should keep in mind.

azoitl:
@ptziegler good point I'll better keep it for 3.19.0

azoitl:
As we are at the start of the new dev cycle I'll merge this early to get early feedback. Tested at least with 4diac IDE and GMF Runtime.

Issue #286, opened 2023-11-11
Minor memory leak in Cursors utility class


ptziegler:
A list of all global cursors is initialized statically. This technically works, as long as the Draw2D bundle is never deactivated, because then the resources are disposed once the application terminates.
But if the class happens to be unloaded prior to termination (e.g. if the Draw2D bundle is deactivated), then those resources remain allocated.

https://github.com/eclipse/gef-classic/blob/fb5f68dee5568acacb4bf0931cc89e6c43b5bcb9/org.eclipse.draw2d/src/org/eclipse/draw2d/Cursors.java#L174-L197

As suggested in the JavaDoc of the Cursor constructor, we can probably replace those invocations with Display.getSystemCursor(...)

Because we pass `null` to the constructor, SWT should attempt to retrieve the required Device instance via `Display.getCurrent()` if in the UI thread, otherwise by `Display.getDefault()`. Note that cursors are only supported by displays.

We can probably reuse the same logic that is used to create the system colors, where we simply replace the call to `getSystemColor()` with `getSystemCursor()`:
```
class SystemCursorFactory {
	public static Cursor getCursor(final int which) {
		Display display = Display.getCurrent();
		if (display != null)
			return display.getSystemCursor(which);
		display = Display.getDefault();
		final Cursor result[] = new Cursor[1];
		display.syncExec(new Runnable() {
			@Override
			public void run() {
				synchronized (result) {
					result[0] = Display.getCurrent().getSystemCursor(which);
				}
			}
		});
		synchronized (result) {
			return result[0];
		}
	}
}
```

azoitl:
That is a good point it also makes no sense to allocate new cursors at all when we anyhow want system cursors. I only wanted to point out that the Display class has now a `syncCall()` method which takes care of the return value handling and you can reduce the code here a lot.

PR #288, opened 2023-11-11
Create system cursors using Display getSystemCursor method #286

ptziegler:
Creating the static standard platform cursors may lead to a potential memory leak, if the Cursors class ever happens to be unloaded, before the application is terminated (e.g. when the bundle is stopped), because this process won't free up the allocated resources.

To avoid this issue, we use Display.getSystemCursor, similar to what is done to get the standard platform colors.

Given that those factory methods are technically API, they have also been documented, with a large part taken from the getSystemCursor() and getSystemColor() JavaDoc.


ptziegler:
I've thought about creating a shared factory class for the system cursors and colors. But given how short those methods are, I don't think it's really worth the effort...

azoitl:
> I've thought about creating a shared factory class for the system cursors and colors. But given how short those methods are, I don't think it's really worth the effort...

Totally agree here.

PR #291, opened 2023-11-12
Cleaned all Draw2d Tests

azoitl:
Fixed most warnings that resulted from stricter compiler rules.


azoitl:
Merging as it only cleans tests

Issue #290, opened 2023-11-12
SonarCloud.io repository for GEF Classic


azoitl:
It is possible to integrate a SonarQube analysis in Eclipse Jipp maven builds: https://wiki.eclipse.org/SonarQube. In another projects we use that and I find the dashboard and the overview on potential issues extremely helpful for maintenance tasks. 
 
Therefore I took the liberty and requested a SonarCloud.io repository for GEF Classic: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3938

I'll keep you updated here on the progress.

azoitl:
I was able to final get a build into a stage where sonar analysis is running. However I get the following error:

`[ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.9.1.2184:sonar (default-cli) on project org.eclipse.gef.releng: Two modules have the same id: 'org.eclipse.gef_root:org.eclipse.gef'. Each module must have a unique id. -> [Help 1]]`

I'm a bit lost what is wrong here. Is it that we have org.eclipse.gef and org.eclipse.gef_root?

Has anyone an idea where I took the wrong turn?

ptziegler:
The way I understand this tool is that `sonar.projectKey` needs to be unique for each module.

But in the Jenkins job, all modules use `gef-classic`

> sonar:sonar
> -Dsonar.projectKey=gef-classic
> -Dsonar.organization=eclipse
> -Dsonar.host.url=${SONAR_HOST_URL}
> -Dsonar.login=${SONARCLOUD_TOKEN}
> -Dmaven.test.skip=true

Which them seems to result in the message you described above.

> Two modules have the same id: 'org.eclipse.gef_root:org.eclipse.gef'. Each module must have a unique id. -> [Help 1]]

So I would define it as a property in the parent pom, with value `${groupId}.${artifactId}` or something like that.

ptziegler:
This is the relevant [code](https://github.com/SonarSource/sonar-scanner-maven/blob/b113d0e01d8029eeb30c6f413c8c1b226b940576/src/main/java/org/sonarsource/scanner/maven/bootstrap/MavenProjectConverter.java#L270C1-L284C4) that calculates the module id.
With `ScanProperties.PROJECT_KEY` being `sonar.projectKey`, I assume.

```
  private static void defineModuleKey(MavenProject pom, Map<String, String> props, @Nullable String specifiedProjectKey) {
    String key;
[...]

PR #298, opened 2023-11-14
Changed name of maven root module to org.eclipse.gef.root

azoitl:
Sonar compliant about to maven modules with the same name.

Addresses #290.


azoitl:
will merge for tests with sonar jenkins job

Issue #282, opened 2023-11-07
Figure.setBounds() fireFigureMoved called when only resized


azoitl:
Over the last days I'm tracing down performance issues in our GEF Classic based editors. The symptom that we is that while redrawing a large amount of garbage is created. When you are already operating on the memory limit this can be problematic. What I noticed is that the bend points of our connections constantly got updated. This happened even when just moving the mouse across the canvas or when scrolling. This seemed for me very odd. Debugging deeper I noticed that at the and of Figure's `setBounds()` there is this piece of code:
```
if (translate || resize) {
	if (resize)
		invalidate();
	fireFigureMoved();
	repaint();
}
```
The problem with this is that `fireFigureMoved()` is called in both cases is the figure is translated or if the figure is resized. Especially if this is one of the root figures (e.g., root figure of FigureCanvas) then the AncestorListeners will invalidate per default all connections issue at least one other repaint and issue an invocation of the Router for all connections. I don't think this should happen. 

Therefore I experimented and changed the above code to:
```
if (resize) {
	invalidate();
}
if (translate) {
	fireFigureMoved();
}
if (translate || resize) {
	repaint();
}
```
For the editors in 4diac IDE this works and I could get rid of the unnecessary updates mentioned above. 

However as you can imagine doing such a change in one of our most important classes makes me quite nervous. Therefore I'm calling out here to other Draw2d and GEF Classic experts (e.g., @Phillipus @BauePhil @pcdavid @ptziegler @kjsmita6 @Destrolaric @vogella @VincentLorenzo @wimjongman @kysmith-csg) if I missed something obvious.

ptziegler:
@azoitl 
That's why I get when I simply adapt your changes:

[FigureBounds.webm](https://github.com/eclipse/gef-classic/assets/70652978/fd1001de-ad9d-44fe-89f0-58e4562bec89)

I can have a closer look at which listener exactly is responsible for updating the border. But if this breaks our project, it probably also breaks other users projects...

azoitl:
@ptziegler I feared that this change would have much more impact then I originally hoped. You pointed me to an area that I haven't initially thought of. All SelectionFeedbackHandles are operating with an AncestorListener and will not get updated when figures are resized. While this is handled deep inside GEF and could maybe be changed I'm not sure this should be done. 

[...]

PR #300, opened 2023-11-15
Rework Rectangle.intersects so that it does not need to create a rect

azoitl:
The intersects checking method of Rectangle used intersect with a copy of an rectangle. This generated a lot of garbage during drawing as intersect is used to check if a children should be drawn or not.

Addresses #282


ptziegler (approved):
I've tried to come up with a way where you don't need to copy parts of the intersect() method, but without success.
Those few lines should be fine.

azoitl:
> I've tried to come up with a way where you don't need to copy parts of the intersect() method, but without success. Those few lines should be fine.

I was also not happy about that but also didn't find a better solution.

PR #301, opened 2023-11-15
Reworked shrink to not create an addtional point

azoitl:
Addresses #282.

@ptziegler I think you added that check that I had to rework a bit.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
@@ -190,6 +190,7 @@ public class Rectangle implements Cloneable, java.io.Serializable, Translatable 
 	 * @since 2.0
 	 * @deprecated Use {@link #shrink(Insets)} instead.
 	 */
+	@Deprecated
 	public Rectangle crop(Insets insets) {
 		return shrink(insets);
 	}
@@ -221,8 +222,9 @@ public class Rectangle implements Cloneable, java.io.Serializable, Translatable 
 	 */
 	@Override
 	public boolean equals(Object o) {
-		if (this == o)
+		if (this == o) {
 			return true;
+		}
 		if (o instanceof Rectangle r) {
 			return (x == r.x()) && (y == r.y()) && (width == r.width()) && (height == r.height());
 		}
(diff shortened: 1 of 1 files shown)


ptziegler (approved):
No objections from my side. Out of curiosity, is there any noticeable difference between both implementations?
In general, I think having dedicated methods for getting the horizontal/vertical center is a good idea, given how often we seem to use stuff like x +width / 2.

azoitl:
In my application I use expand, which invokes shrink, for drawing borders. Moving the mouse once across the canvas without any interaction with an object resulted in more then 18.000 Point creations by the getCenter() method. Which is in the end just garbage. In my setting where we are operating at the memory limit this is problematic (i.e., issues #282). Not as bad as the fix in PR #300, which creates at the same time 86.000 rectangles for checking if two rectangles have an intersection. So if you want to have a look on that one as well I would be happy. And after that the really big (but maybe dangerous one) SWTGraphics some 110.000 rectangles are created at the same time. This all sums up.

ptziegler:
So you're reaching the limit of the garbage collector, got it. Thanks for the clarification.

azoitl:
Yes. However after digging deeper I think my fixes are just a drop in the ocean. The really hard to crack nut is the memory allocated for maintaining the clipping in the SWT Graphics. This is triggered with every push and pop state and if you have like I a view 1000 Figures in one drawing this gets to quite some amount of memory. Not sure how I solve that.

vogella:
This reminds me of the "SWT cursor movement" memory allocation. If you open the heap monitor in the IDE and move your mouse all the time, the used heap goes up and up until gc is running.

azoitl:
> This reminds me of the "SWT cursor movement" memory allocation. If you open the heap monitor in the IDE and move your mouse all the time, the used heap goes up and up until gc is running.

Yes this is exactly what I'm investigating. When you have a bigger graph like I have to improve moving the mouse cursor once across the canvas gives you app. 300- 600M of garbage. 

By now I know the root cause of the problem: each figure is setting its own clip rectangle, this is a rather heavy operation. On push and pop this clip rectangle is updated again. However I have no clue if this can be omitted or how it can be improved. The only solution that I currently have in mind is duplicating SWTGraphics and provide a version without the clipping. However this may break a lot of editors. So I would only do this for starting and as experiment in 4diac IDE and see where I'm getting with this.

PR #302, opened 2023-11-16
Simplify Rectangle.intersects

mx990:
Simplify and optimize the intersects checking method of Rectangle.

The following shows the individual steps taken:
```java
int x1 = Math.max(x, rect.x()); // (1)
int x2 = Math.min(x + width, rect.x() + rect.width()); // (2)
int y1 = Math.max(y, rect.y()); // (3)
int y2 = Math.min(y + height, rect.y() + rect.height()); // (4)
return (((x2 - x1) > 0) && ((y2 - y1) > 0)); // (5)
```
```java
return x2 > x1 && y2 > y1; // (6) simplify 5
```
```java
return x1 < x2 && y1 < y2; // (7) reverse operands in 6
```
```java
// (8) insert 1, 2, 3, 4 in 7
return Math.max(x, rect.x()) < Math.min(x + width, rect.x() + rect.width()) 
    && Math.max(y, rect.y()) < Math.min(y + height, rect.y() + rect.height());
```
```java
// (9) factor out min and max w.r.t inequalities in 8
return x < x + width && x < rect.x() + rect.width() && rect.x() < x + width && rect.x() < rect.x() + rect.width()
    && y < y + height && y < rect.y() + rect.height() && rect.y() < y + height && rect.y() < rect.y() + rect.height();
```
```java
// (10) simplyfy by noting that
// "x < x + width",
// "rect.x() < rect.x() + rect.width()",
// "y < y + height", 
// "rect.y() < rect.y() + rect.height()"
// are always true
return x < rect.x() + rect.width() && rect.x() < x + width
    && y < rect.y() + rect.height() && rect.y() < y + height;
```

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
@@ -725,7 +725,7 @@ public class Rectangle implements Cloneable, java.io.Serializable, Translatable 
 		int x2 = Math.min(x + width, rect.x() + rect.width());
 		int y1 = Math.max(y, rect.y());
 		int y2 = Math.min(y + height, rect.y() + rect.height());
-		if (((x2 - x1) < 0) || ((y2 - y1) < 0)) {
+		if (((x2 - x1) <= 0) || ((y2 - y1) <= 0)) {
 			return setBounds(0, 0, 0, 0); // no intersection
 		}
 		return setBounds(x1, y1, x2 - x1, y2 - y1);
@@ -739,12 +739,9 @@ public class Rectangle implements Cloneable, java.io.Serializable, Translatable 
 	 * @since 2.0
 	 */
 	public boolean intersects(Rectangle rect) {
-		int x1 = Math.max(x, rect.x());
-		int x2 = Math.min(x + width, rect.x() + rect.width());
-		int y1 = Math.max(y, rect.y());
-		int y2 = Math.min(y + height, rect.y() + rect.height());
-
-		return (((x2 - x1) > 0) && ((y2 - y1) > 0));
+		return width > 0 && height > 0 && rect.width() > 0 && rect.height() > 0 //
+				&& x < rect.x() + rect.width() && rect.x() < x + width //
+				&& y < rect.y() + rect.height() && rect.y() < y + height;
 	}
 
 	/**
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/RectangleTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/RectangleTest.java
@@ -260,6 +260,14 @@ public class RectangleTest extends BaseTestCase {
 		assertFalse(rectangle1.intersects(new Rectangle(-100, -100, 10, 10)));
 		//
 		assertFalse(rectangle1.intersects(new Rectangle(0, 0, 5, 10)));
+		//
+		assertFalse(rectangle1.intersects(new Rectangle(15, 20, 0, 10)));
+		assertFalse(rectangle1.intersects(new Rectangle(15, 20, 10, 0)));
+		//
+		Rectangle rectangle3 = new Rectangle(0, 30, 0, 40);
+		assertFalse(rectangle3.intersects(rectangle3));
+		//
+		assertFalse(rectangle3.intersects(new Rectangle(0, 30, 100, 10)));
 	}
 
 	@SuppressWarnings("static-method")


org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
```diff
@@ -739,12 +739,8 @@ public Rectangle intersect(Rectangle rect) {
 	 * @since 2.0
 	 */
 	public boolean intersects(Rectangle rect) {
-		int x1 = Math.max(x, rect.x());
-		int x2 = Math.min(x + width, rect.x() + rect.width());
-		int y1 = Math.max(y, rect.y());
-		int y2 = Math.min(y + height, rect.y() + rect.height());
-
-		return (((x2 - x1) > 0) && ((y2 - y1) > 0));
```
ptziegler, line 747: > ```java
> return x2 > x1 && y2 > y1; // (6) simplify 5
> ```

✔️ Inequalities are preserved under addition.

> ```java
> return x1 < x2 && y1 < y2; // (7) reverse operands in 6
> ```

✔️ Inequalities are symmetric.

> ```java
> // (8) insert 1, 2, 3, 4 in 7
> return Math.max(x, rect.x()) < Math.min(x + width, rect.x() + rect.width()) 
>     && Math.max(y, rect.y()) < Math.min(y + height, rect.y() + rect.height());
> ```

✔️

> ```java
> // (9) factor out min and max w.r.t inequalities in 8
> return x < x + width && x < rect.x() + rect.width() && rect.x() < x + width && rect.x() < rect.x() + rect.width()
>     && y < y + height && y < rect.y() + rect.height() && rect.y() < y + height && rect.y() < rect.y() + rect.height();
> ```

✔️ If `Math.max(a1, a2) < b`, then `a1 < b` and `a2 < b`. Analogously, if `a < Math.min(b1, b2)`, then `a < b1` and `a < b2` 

> // (10) simplyfy by noting that
> // "x < x + width",
> // "rect.x() < rect.x() + rect.width()",
> // "y < y + height", 
> // "rect.y() < rect.y() + rect.height()"
> // are always true
> return x < rect.x() + rect.width() && rect.x() < x + width
>     && y < rect.y() + rect.height() && rect.y() < y + height;


✖️ x < x + width is true, if and only if the width is positive.

Counterexample:
Consider r1 = (x=5, y=0, w=10, h=10), r2 = (x=10, y = 1, w=0, h=8) with r1.intersects(r2)

(I)
int x1 = Math.max(x, rect.x());
int x2 = Math.min(x + width, rect.x() + rect.width());
int y1 = Math.max(y, rect.y());
int y2 = Math.min(y + height, rect.y() + rect.height());

return (((x2 - x1) > 0) && ((y2 - y1) > 0));

=> false

(II)

return x < rect.x() + rect.width() && rect.x() < x + width //
				&& y < rect.y() + rect.height() && rect.y() < y + height;

=> true

ptziegler (changes requested):

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java
mx990: Now that you mention this, I noticed that there appears to be an inconsistency between this method and `intersect(Rectangle rect)`:
- `intersect(Rectangle rect)` seems to allow empty intersections:
    ```java
    if (((x2 - x1) < 0) || ((y2 - y1) < 0)) {
        return setBounds(0, 0, 0, 0); // no intersection
    }
    return setBounds(x1, y1, x2 - x1, y2 - y1);
    ```
- `intersects(Rectangle rect)` does not:
    ```java
    return (((x2 - x1) > 0) && ((y2 - y1) > 0));
    ```

The latter would also imply that an empty rectangle does not intersect with itself.

azoitl: I fear I messed up De Morgan here.

mx990: If we choose to align the current behavior with that of `intersect(Rectangle rect)`, I would propose:
```java
int x1 = Math.max(x, rect.x()); // (1)
int x2 = Math.min(x + width, rect.x() + rect.width()); // (2)
int y1 = Math.max(y, rect.y()); // (3)
int y2 = Math.min(y + height, rect.y() + rect.height()); // (4)
return (((x2 - x1) >= 0) && ((y2 - y1) >= 0)); // (5)
```
```java
return x2 >= x1 && y2 >= y1; // (6) simplify 5
```
```java
return x1 <= x2 && y1 <= y2; // (7) reverse operands in 6
```
```java
// (8) insert 1, 2, 3, 4 in 7
return Math.max(x, rect.x()) <= Math.min(x + width, rect.x() + rect.width()) 
    && Math.max(y, rect.y()) <= Math.min(y + height, rect.y() + rect.height());
```
```java
// (9) factor out min and max w.r.t inequalities in 8
return x <= x + width && x <= rect.x() + rect.width() && rect.x() <= x + width && rect.x() <= rect.x() + rect.width()
    && y <= y + height && y <= rect.y() + rect.height() && rect.y() <= y + height && rect.y() <= rect.y() + rect.height();
```
```java
// (10) reorder inequalities in 9
return x <= x + width && y <= y + height && rect.x() <= rect.x() + rect.width() && rect.y() <= rect.y() + rect.height()
    && x <= rect.x() + rect.width() && rect.x() <= x + width
    && y <= rect.y() + rect.height() && rect.y() <= y + height;
```
```java
// (11) simplyfy first line of 10
return width >= 0 && height >= 0 && rect.width() >= 0 && rect.height() >= 0
    && x <= rect.x() + rect.width() && rect.x() <= x + width
    && y <= rect.y() + rect.height() && rect.y() <= y + height;
```

I have rebased and amended the PR to reflect these changes and also added some additional tests.

ptziegler: Just to avoid any misunderstandings, by "disallowing empty intersections" you mean that intersects() should return false, if the intersected area is empty, correct?
I would consider this to be the expected behavior, which is why I'm leaning towards the first solution:
> align the behavior of intersect(Rectangle rect) with that of intersects(Rectangle rect) by disallowing empty intersections

mx990: > Just to avoid any misunderstandings, by "disallowing empty intersections" you mean that intersects() should return false, if the intersected area is empty, correct?

Yes, exactly.

> I would consider this to be the expected behavior, which is why I'm leaning towards the first solution:
> 
> > align the behavior of intersect(Rectangle rect) with that of intersects(Rectangle rect) by disallowing empty intersections

I agree that this is probably the most sensible behavior, given that this seems to be primarily used in the context of calculating drawable areas and clipping for which empty intersections make little sense.

ptziegler:
```
return width > 0 && height > 0 && rect.width() > 0 && rect.height() > 0 //
				&& x < rect.x() + rect.width() && rect.x() < x + width //
				&& y < rect.y() + rect.height() && rect.y() < y + height;
```

With the new implementation, I'm counting 19 arithmetic operations in total.

```
int x1 = Math.max(x, rect.x());
int x2 = Math.min(x + width, rect.x() + rect.width());
int y1 = Math.max(y, rect.y());
int y2 = Math.min(y + height, rect.y() + rect.height());

return (((x2 - x1) > 0) && ((y2 - y1) > 0));
```

With the old implementation, I'm counting only 13 arithmetic operations (assuming that the ternary operator in Math.max() and Math.min() only counts as one and assuming that the compiler will inline the local variables).

Is this really going to be more efficient? @azoitl brought up performance problems in the Figure class some time ago, so my concern is that this'll be counterproductive...

ptziegler: I also want to apologize for being this nit-picky :sweat_smile: It's just that this is one of the core classes, so I tend to always look a little closer when it comes to modifications here...

ptziegler: > Is this really going to be more efficient?

It probably is when checking whether two rectangles **don't** intersect, because Java only evaluates until the first expression that returns false. Whereas with the current implementation the check always takes the same amount of time.

azoitl: @ptziegler I'm always also nervous when it is about the core Draw2d classes. Therefore I'm very happy about the feedback. Performance wise I don't think both versions have much difference. The big problem was the allocation we had before. But in the end I think the new version is a bit more explicit, which I like. So if you agree I would merge it to have it for 3.18.0.

ptziegler (approved):
> So if you agree I would merge it to have it for 3.18.0.

That's fine by me. Both methods should behave the same, which was my primary concern.

azoitl:
> That's fine by me. Both methods should behave the same, which was my primary concern.

Mine too. Then I'll merge and start the RC1 process.

ptziegler:
@mx990 Thank you for your contribution!

torokati44:
@ptziegler:
> Both methods should behave the same, which was my primary concern.

This is false. You explicitly mention above (in the review notes that are hidden by default) that `intersect` changed how it handles empty rectangles.
Now it always returns `(0, 0, 0, 0)`, whereas this was not the case before: If the empty rectangle had a `width` and/or `height` of `0`, its position and non-zero dimension (if any) was unaffected.

This was caused by the following change: https://github.com/eclipse-gef/gef-classic/commit/70c2969be53e51b54a7d2b95ed7c51b46c26b755#diff-daacfe6bcb876cf7c5ea92e6ba6acc0fbbbb257d6e1c1223f3277e2780c50f08L728-R728

PR #304, opened 2023-11-22
[releng] fixed since tags in GraphContainer & GraphConnection

azoitl:


azoitl:
Had it allready longer on my todo list but forgot since today.

Issue #116, opened 2022-09-21
Removal of com.ibm.icu has changed behaviour of BreakIterator


Phillipus:
With https://github.com/eclipse/gef-classic/issues/92 we now have changed behaviour in line breaks in TextFlow Figures.

`java.text.BreakIterator` is rendering different line break behaviour in `draw2d.text.FlowUtilities` than `com.ibm.icu.text.BreakIterator`.

This means that line breaks will be different in diagram figures that use draw2d `TextFlow` components.

See https://github.com/archimatetool/archi/issues/871

Should we revert #92?

vogella:
Can we fix the code without ICU? Platform also plans to drop it and this library has only little maintenance

Phillipus:
> Can we fix the code without ICU? Platform also plans to drop it and this library has only little maintenance

I don't think that would be easy if it is even possible. The internal implementations of `BreakIterator` sub-classes are quite different.

Phillipus:
Run the following code with `import com.ibm.icu.text.BreakIterator;` and again with `import java.text.BreakIterator;`:

```java
String text = "FirstPart/LastPart";

BreakIterator breakIterator = BreakIterator.getLineInstance();
breakIterator.setText(text);
int start = breakIterator.first();

for(int end = breakIterator.next(); end != BreakIterator.DONE; start = end, end = breakIterator.next()) {
    System.out.println(text.substring(start, end));
}
```

With `com.ibm.icu.text.BreakIterator` the output is:

```
FirstPart/
LastPart
```

With  `java.text.BreakIterator` the output is:
```
FirstPart/LastPart
```

vogella:
If only draw2d.text.FlowUtilities is different, maybe we can copy the split logic into the class?

@azoitl WDYT?

Phillipus:
The word split is one example of the difference in line break logic. There are other differences (See https://github.com/archimatetool/archi/issues/871)

azoitl:
> If only draw2d.text.FlowUtilities is different, maybe we can copy the split logic into the class?
> 
> @azoitl WDYT?

I think it would be good if we can keep the ibm.icu out. Keeping the code in sounds reasonable. Can we just do that? But ICU BreakIterator has depends on at least three more classes.

I had also a look on java.text.BreakIterator. I noticed that there are more different breakiterators:
- Line instance
- Word instance
- Character instance

See below for what they do. Could it be that we are using wordInstance wrong? 

`Line Instance:
[...]

PR #307, opened 2023-12-08
Add ability to contribute custom BreakIterator/Bidi via SPI

### LINKED ISSUE #289: Broken TextFlowWrapTest due to com.ibm.icu.text.BreakIterator
The TextFlowWrapTest is currently marked as ignored, due to unstable failures in `testNestedInlineFlows()`, `testHardWrapping()` and `testInlineFlow()`. 

After further looking into it, I came to the conclusion that they are not unstable, but rather that they fail because the BreakIterator used in FlowUtilities doesn't behave the way as expected by the test.

To reproduce:
- Remove the Ignore annotation from the test class
- Run the tests, 3 test cases should fail
- Go to FlowUtilities
- Switch to javax.text.BreakIterator
- Run the tests again, all should pass now

Looking at the history, we switched to the javax BreakIterator with 6b6eba6093e4f62536bec7d2f198aa961765cfa9 but then reverted back to the com.ibm.icu BreakIterator with 9e560cf2b2d63e03d6721a9346895d4ce44471f2, due to #116 

Given that the test was written for the com.ibm.icu BreakIterator but now fails, even though we continue to use the same class, makes me believe that something in the com.ibm.icu implementation has changed in a way that was not intended by the original author.

In general, there are two options on how we can approach this issue:

1) Accept the new behavior as the "correct" state, meaning we have to adjust the unit test accordingly.
2) Switch back to the javax BreakIterator under the assumption that this is how the iterator is supposed to behave.
3) ~~Continue to ignore those failing test cases~~

Given that we explicitly want the current behavior, I'm leaning towards the first solution. What are your thoughts?

ptziegler:
With this change, the FlowUtilities class uses the BreakIteratorProvider
to create its BreakIterator instances. This way, downstream project can
contribute their own implementations, if they don't want to use the
native Java implementation.

Similarly, a BidiProvider class is created, to support a custom Bidi
implementation in the BidiProcessor class.

Resolves #289

### DIFF
--- a/org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.draw2d.tests
-Bundle-Version: 3.12.200.qualifier
+Bundle-Version: 3.13.0.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.2.0,4.0.0)",
--- a/org.eclipse.draw2d/META-INF/MANIFEST.MF
+++ b/org.eclipse.draw2d/META-INF/MANIFEST.MF
@@ -14,7 +14,6 @@ Export-Package: org.eclipse.draw2d,
  org.eclipse.draw2d.text,
  org.eclipse.draw2d.widgets,
  org.eclipse.draw2d.zoom
-Import-Package: com.ibm.icu.text
 Require-Bundle: org.eclipse.swt;bundle-version="[3.4.0,4.0.0)";visibility:=reexport
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-17
--- a/org.eclipse.draw2d.tests/META-INF/services/java.text.spi.BreakIteratorProvider
+++ b/org.eclipse.draw2d.tests/META-INF/services/java.text.spi.BreakIteratorProvider
@@ -1,0 +1,1 @@
+org.eclipse.draw2d.test.TextFlowWrapTest$BreakIteratorProviderMock
--- a/org.eclipse.draw2d.tests/META-INF/services/org.eclipse.draw2d.text.BidiProvider
+++ b/org.eclipse.draw2d.tests/META-INF/services/org.eclipse.draw2d.text.BidiProvider
@@ -1,0 +1,1 @@
+org.eclipse.draw2d.test.BidiProcessorTest$BidiProviderMock
--- a/org.eclipse.draw2d.tests/pom.xml
+++ b/org.eclipse.draw2d.tests/pom.xml
@@ -10,7 +10,7 @@
 	</parent>
 	<groupId>org.eclipse.draw2d.plugins</groupId>
 	<artifactId>org.eclipse.draw2d.tests</artifactId>
-	<version>3.12.200-SNAPSHOT</version>
+	<version>3.13.0-SNAPSHOT</version>
 	<packaging>eclipse-test-plugin</packaging>
 	<profiles>
 		<profile>
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/BaseTestCase.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/BaseTestCase.java
@@ -32,6 +32,7 @@ import org.junit.Assert;
 public abstract class BaseTestCase extends Assert {
 
 	protected static final Font TAHOMA = new Font(null, "Tahoma", 8, 0);//$NON-NLS-1$
+	protected static final Font SERIF = new Font(null, "Serif", 8, 0);//$NON-NLS-1$
 
 	public static void assertEquals(Image expected, Image actual) {
 		assertTrue("The given images did not match", //$NON-NLS-1$
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/BidiProcessorTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/BidiProcessorTest.java
@@ -1,0 +1,55 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.draw2d.test;
+
+import java.text.Bidi;
+import java.util.ServiceLoader;
+
+import org.eclipse.draw2d.text.BidiProvider;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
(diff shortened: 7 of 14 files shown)


ptziegler:
Open topic: The package `com.ibm.icu.text` is still used inside the [BidiProcessor](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/text/BidiProcessor.java).

ptziegler:
Weird how TextFlowWrapTest only fails on certain systems...
Please don't tell me Oracle made a platform-specific implementation for this crap.

azoitl:
> Weird how TextFlowWrapTest only fails on certain systems... Please don't tell me Oracle made a platform-specific implementation for this crap.

When I looked the last time into these tests I had the impression that the font used on the different systems influences the outcome. Because the BreakIterator is only providing suitable breaks, and the TextFlow is then deciding on the space requirements of the text on which of the suitable breaks it chooses. Different fonts have different character widths and with that we get platform specific tests.

ptziegler:
> When I looked the last time into these tests I had the impression that the font used on the different systems influences the outcome.

That's a good point. This also explains why this problem doesn't show up when executed locally, likely because I have different font installed than what's available on the test machine.

ptziegler:
> Different fonts have different character widths and with that we get platform specific tests.

I've had a look at it again and we consistently use the Tahoma font on all systems. But what seems to be causing problems is the dpi scaling.

Example:
The string "tester" has the following width:
- Windows: 29
- Linux: 32
- MacOS: 21

What a pain.

azoitl:
Are you sure Tahoma is available on all three systems?

ptziegler:
> Are you sure Tahoma is available on all three systems?

Hm... Apparently not, even though SWT doesn't complain

```
Available Fonts:
Font: DejaVu Math TeX Gyre
Font: DejaVu Sans
Font: DejaVu Sans Mono
Font: DejaVu Serif
Font: Lato
Font: Liberation Mono
Font: Liberation Sans
Font: Liberation Sans Narrow
Font: Liberation Serif
Font: Monospace
Font: Noto Color Emoji
Font: Sans
Font: Serif
Font: System-ui
```

ptziegler:
Using Serif seems to do the trick...

org.eclipse.draw2d/META-INF/MANIFEST.MF
```diff
@@ -14,7 +14,8 @@ Export-Package: org.eclipse.draw2d,
  org.eclipse.draw2d.text,
  org.eclipse.draw2d.widgets,
  org.eclipse.draw2d.zoom
-Import-Package: com.ibm.icu.text
+Import-Package: com.ibm.icu.text,
```
azoitl, line 17: I think this import package can be removed now or?

org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.draw2d.tests
-Bundle-Version: 3.12.200.qualifier
+Bundle-Version: 3.12.300.qualifier
```
azoitl, line 5: I already increased the maintenance version number when doing the 3.19 preparations. Is this then still needed?

As we have added the new factory in draw2d and the tests using it should we change the version number to 3.15 to show that the tests are aligned? Or at least change the minor version number?

org.eclipse.draw2d/META-INF/MANIFEST.MF
ptziegler: It's still used by the [BidiProcessor](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/text/BidiProcessor.java)

org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
ptziegler: > As we have added the new factory in draw2d and the tests using it should we change the version number to 3.15 to show that the tests are aligned? Or at least change the minor version number?

I don't think we should jump from 3.12.200 to 3.15, but at the very least to 3.13, because new tests have been "added" (i.e. reactivated).

azoitl: This sounds like a plan.

org.eclipse.draw2d/META-INF/MANIFEST.MF
azoitl: Oh missed that.

ptziegler: But I think I can use the same mechanism to get rid of that dependency as well. The only trick part being `Bidi.requiresBidi(...)` a static method call...

ptziegler:
A few notes:

In hindsight, it would've also been possible to simply add a setter to FlowUtilities and BidiProcessor.  The reason I stuck to the OSGi service is because it doesn't require an explicit initialization. Simply creating the component class is sufficient.

Services have to be interfaces. But because BreakIterator is an abstract class without such interface, I've gone with the factory pattern instead of implementing the methods directly.

Similarly, because we only call a static method in Bidi, the factory currently only specifies `requiresBidi()`.

azoitl:
I like the solution as it is very non-intrusive for Eclipse-based GEF Classic users. Where I'm not sure is if you have plain SWT and draw2d and no OSGI. In such cases you would get now dependency errors or?

Are there users with that use case out there?

ptziegler:
> Where I'm not sure is if you have plain SWT and draw2d and no OSGI. In such cases you would get now dependency errors or?

Yes, this change adds a hard dependency to OSGi. I suppose I could rewrite it to use SPI instead, but then you'll have a lot fun with the different classloaders...

I'll see if I can come up with a more general solution.

azoitl:
You also have no idea on our user base, or?

ptziegler:
> You also have no idea on our user base, or?

I don't. But given that you _can_ use SWT without OSGi, I don't think it's appropriate to require it for GEF.

ptziegler:
I went over this again and switched to SPI. This way it should work with and without OSGi.
The downside being that now you have to make sure that the service is available to the Draw2D classloader. The general workflow in OSGi would then be to:

a) Set the ContextClassLoader before calling FlowUtilities.
b) Create a Draw2D fragment, containing the service

Overall, I think this looks much cleaner now, especially because SPI allows me to also use abstract classes. And I can reuse the BreakIteratorProvider from the JDK.

azoitl:
I totally agree it is much cleaner now. Thx for spending so much time on this issue.

ptziegler:
I just remembered: Do we have a process of communicating this with downstream projects? Given that both BreakIterator uses their unique set of line-break rules, this change will likely cause issues like #116 to flare up again.

azoitl:
I also thought about this if we ever plan to remove deprecated API. The only thing that came to my mind is a mail to our mailing list and maybe a cross post on cross projects. But that seems a bit weak or?

ptziegler:
For the WindowBuilder, I've recently started creating a [RELEASE_NOTES](https://github.com/eclipse-windowbuilder/windowbuilder/blob/master/RELEASE_NOTES.md) file, where I keep track of any noteworthy changes. So people know what to expect, when updating to a newer version.

I'm not sure how worthwhile this is for GEF, though, given that we put a lot of effort into backwards compatibility...
Otherwise doing more than a general heads-up seems like overkill.

azoitl:
GEF has a [CHANGELOG](https://github.com/eclipse/gef-classic/blob/master/CHANGELOG.md) file with a similar purpose. Maybe putting such information there could help. Or adding migration section in the documentation. But I'm not sure how this is done.

ptziegler:
The changelog looks like a good place for a brief summary, but I'd put the migration in the documentation. Both seem straigthfoward. I'll try to have a look throughout the week.

Issue #309, opened 2023-12-12
API baseline errors


merks:
Are you guys seeing these kinds of errors?

![image](https://github.com/eclipse/gef-classic/assets/208716/34c4224b-c413-4877-ac54-0d36f0578a57)

I think minor increments are needed...

merks:
The PR eliminates the errors:

![image](https://github.com/eclipse/gef-classic/assets/208716/f0ee8f81-2dcb-4dd7-a200-aee5a147dd0b)

ptziegler:
Hm, I'm also getting those errors with the latest baseline. Though I'm not sure what _exactly_ has changed...

azoitl:
I'm not sure why these errors appear. We changed the Java version already two releases ago and AFAIK we definitely updated all minor versions then. So I'm have no idea what is causing this.

merks:
I’m not a big fan of api tools. It’s  always complaining about something and then one usually has no clue why. So here one could filter the error and hope it doesn’t come back or increase the minor version which does no harm.  But a clean new workspace with errors is a big turn off for new contributors.

ptziegler:
That's a fair point. The minor increment seem rather unproblematic.Merging your pull request sounds like the easiest solution, rather than chasing after ghosts in the api-tools implementation.On Dec 12, 2023 18:25, Ed Merks ***@***.***> wrote:
I’m not a big fan of api tools. It’s  always complaining about something and then one usually has no clue why. So here one could filter the error and hope it doesn’t come back or increase the minor version which does no harm.  But a clean new workspace with errors is a big turn off for new contributors.

ptziegler:
I'm closing this together with #310. As said, I think the work of figuring out what exactly triggered the api tools far outweigh simply updating the minor version.

azoitl:
@ptziegler I' m a bit late to the party but totally agree with you.

PR #310, opened 2023-12-12
BREE changes require a minor version increment

merks:
https://github.com/eclipse/gef-classic/issues/309


ptziegler:
Thank you for cleaning up the API errors.

PR #311, opened 2023-12-12
Only promote build results when on the master branch

ptziegler:
To avoid uploading the artifacts from old branches to the update site, we should make sure that we only do so, when on the master branch.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,21 +35,29 @@ pipeline {
 	 				)
 	 			}
 				failure {
-		            mail bcc: '', body: "<b>GEF Classic: ${env.JOB_NAME} <br>Build Number: ${env.BUILD_NUMBER} <br> URL de build: ${env.BUILD_URL}", cc: '', charset: 'UTF-8', from: '', mimeType: 'text/html', replyTo: '', subject: "ERROR CI: GEF Classic -> ${env.JOB_NAME}", to: "alois.zoitl@gmx.at";
+					script {
+						if (env.BRANCH_NAME == 'master') {
+							mail bcc: '', body: "<b>GEF Classic: ${env.JOB_NAME} <br>Build Number: ${env.BUILD_NUMBER} <br> URL de build: ${env.BUILD_URL}", cc: '', charset: 'UTF-8', from: '', mimeType: 'text/html', replyTo: '', subject: "ERROR CI: GEF Classic -> ${env.JOB_NAME}", to: "alois.zoitl@gmx.at";
+						}
+					}
 		        }
 			}
 
 		}
         stage('Deploy') {
             steps {
-                sshagent ( ['projects-storage.eclipse.org-bot-ssh']) {
-                    sh '''
-                        ssh -o BatchMode=yes genie.gef@projects-storage.eclipse.org rm -rf /home/data/httpd/download.eclipse.org/tools/gef/classic/latest
-                        ssh -o BatchMode=yes genie.gef@projects-storage.eclipse.org mkdir -p /home/data/httpd/download.eclipse.org/tools/gef/classic/latest
-                        scp -o BatchMode=yes -r org.eclipse.gef.repository/target/repository/* genie.gef@projects-storage.eclipse.org:/home/data/httpd/download.eclipse.org/tools/gef/classic/latest
-                    '''
-                    }
-                }
+				script {
+					if (env.BRANCH_NAME == 'master') {
+						sshagent ( ['projects-storage.eclipse.org-bot-ssh']) {
+							sh '''
+								ssh -o BatchMode=yes genie.gef@projects-storage.eclipse.org rm -rf /home/data/httpd/download.eclipse.org/tools/gef/classic/latest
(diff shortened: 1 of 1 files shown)


ptziegler:
@azoitl This is what we just talked about

ptziegler:
That looks a lot like this branch build wasn't uploaded :tada: 
I've also made it so that an email is only sent out when the master fails.
```
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Deploy)
[Pipeline] tool
[Pipeline] envVarsForTool
[Pipeline] tool
[Pipeline] envVarsForTool
[Pipeline] withEnv
[Pipeline] {
[Pipeline] script
[Pipeline] {
[Pipeline] }
[Pipeline] // script
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // timeout
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
```

azoitl:
looks good to me.

ptziegler:
And the lastest build from master has been uploaded. Very good~ :D

azoitl:
@ptziegler it seems that the publish state on https://download.eclipse.org/tools/gef/classic/latest/ is not the same as latest master. You can see this for example in `EditPartViewer` the ` getSelectedEditParts()` return the untyped list.

ptziegler:
You're right... the version also doesn't match. `org.eclipse.gef` is supposed to be in version 1.17.0, but the update site has 1.15.0.

It's because the artifacts were uploaded by https://ci.eclipse.org/gef/job/build-classic/job/R3_17_maintenance/, which I believe is using the Jenkinsfile from the same branch (i.e. the one without the "master" guard).

For the immediate future, I would simply run the "master" build again, to fix the state of the update site. I'll then create pull-requests for the release branches to also include this check.

ptziegler:
> Obtained Jenkinsfile from 21264727f39233bba4d65581bd92f193395910d1

That's exactly it :/ At least this solves the question why earlier release branches weren't detected by Jenkins.

azoitl:
Should we push then our jenkins file to the [R3_17_maintenance branch](https://ci.eclipse.org/gef/job/build-classic/job/R3_17_maintenance/)?

In the mean-time let me trigger the main branch for my tests.

ptziegler:
> Should we push then our jenkins file to the [R3_17_maintenance branch](https://ci.eclipse.org/gef/job/build-classic/job/R3_17_maintenance/)?

That's the plan. The commit ac03e35e38ff155d969a702d9f60edffa1b8ca83 needs to be merged into R3_16, R3_17 and R3_18.

azoitl:
Can do that from github ui? I guess not or.

ptziegler:
> Can do that from github ui? I guess not or.

I... don't think so. You probably have to cherry-pick the revision and then create individual pull-requests for each branch.

merks:
Do you folks actually produce or intend to ever produce service releases for the older branches?  I'm just wondering they there are built at all?

ptziegler:
> Do you folks actually produce or intend to ever produce service releases for the older branches? I'm just wondering they there are built at all?

Which such short release cycles, I don't think it's really worth publishing a bugfix releases, when the next milestone is only a few weeks away.

I'd say the primary reason we keep them is because "we always did it like this". But in the end it gives downstream projects the ability to get bugfixes for specific releases. The caveat being that those would then have to be built/hosted on their infrastructure.

merks:
I ask because of course I'm looking to change the Jenkinsfile and I would not want to see older branches promoting into the new structure.  I also just noticed that I don't have write access to the ci instance anymore, so I'll have to test further on a ci instance for which I do have write access...

azoitl:
Dumb question: If we are not pushing anything to these branches no build will run or?

merks:
Followup dumb question.  What caused this build to run?

https://ci.eclipse.org/gef/job/build-classic/job/R3_18_maintenance/

ptziegler:
> Dumb question: If we are not pushing anything to these branches no build will run or?

That should be the case.

> Followup dumb question. What caused this build to run?

Yesterday I set up the multi-branch pipeline in order to replace https://ci.eclipse.org/gef/job/gef-classic-latest/, primarily so that the Jenkins run shows up in the GitHub actions.

This job was then executed when Jenkins discovered all available branches.

ptziegler:
> I really really don't want ever for the old branches to publish results accidentally. Are steps being taken to ensure that can't happen?

The maintenance branches will use their branch-local copy of the Jenkinsfile for the build. For this, I need to merge ac03e35e38ff155d969a702d9f60edffa1b8ca83 into the relevant branches.

In short, it adds the following check around the deploy phase. I think you'll do something similar for the updated Jenkinsfile, where the "promote" profile is only activated on the "master" branch, correct?
```
if (env.BRANCH_NAME == 'master') {
   ...
}
```

Then both the old and upcoming release branches shouldn't be able to deploy anywhere.

merks:
Yes, the updated Jenkinsfile has this same test (though it's currently hacked to work of my clone's issue-313 branch).

PR #312, opened 2023-12-12
Bring back sonar build to jenkins file

azoitl:
@ptziegler in theory this should do it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,15 +15,20 @@ pipeline {
 		
 		stage('Build') {
 			steps {
-				wrap([$class: 'Xvnc', takeScreenshot: false, useXauthority: true]) {
-					sh '''
-					export GDK_BACKEND=x11
-					mvn clean verify -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-						-DapiBaselineTargetDirectory=${WORKSPACE} \
-						-Dgpg.passphrase="${KEYRING_PASSPHRASE}" \
-						-Dproject.build.sourceEncoding=UTF-8 \
-						-Peclipse-sign
-					'''
+				withCredentials([string(credentialsId: 'sonarcloud-token', variable: 'SONARCLOUD_TOKEN')]) {
+					withSonarQubeEnv(credentialsId: 'sonarcloud-token', installationName: 'My SonarQube Server') {
+						wrap([$class: 'Xvnc', takeScreenshot: false, useXauthority: true]) {
+							sh '''
+							export GDK_BACKEND=x11
+							mvn clean verify -Dmaven.repo.local=$WORKSPACE/.m2/repository \
+								-DapiBaselineTargetDirectory=${WORKSPACE} \
+								-Dgpg.passphrase="${KEYRING_PASSPHRASE}" \
+								-Dproject.build.sourceEncoding=UTF-8 \
+								-Peclipse-sign  \
+								-B sonar:sonar -Dsonar.projectKey=gef-classic -Dsonar.organization=eclipse -Dsonar.host.url=${SONAR_HOST_URL} -Dsonar.login=${SONARCLOUD_TOKEN}
+							'''
+						}
+					}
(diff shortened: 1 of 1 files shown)


ptziegler:
I see that I'm not the only one suffering with the Jenkinsfile syntax :P

azoitl:
And as the jenkins build tells us we are not there yet. I asked fredrik if he has a tip for me.

ptziegler:
It looks like `installationName` is a mandatory argument of [SonarQube Scanner](https://www.jenkins.io/doc/pipeline/steps/sonar/).

But I think it can be chosen arbitrarily, so perhaps it's enough if you just set it as e.g. `installationName: 'GEF Classic'`?

azoitl:
Ah thx, changed that lets see.

azoitl:
We are a step further. Now we only need to fix how the secret is coming in. As I'm totally new to jenkins files and doing this trial and error based I would be happy for tips. if it helps I can provide a screenshot how it is done in a Jenkins ui setup.

Jenkinsfile
```diff
@@ -11,19 +11,25 @@ pipeline {
 		maven 'apache-maven-latest'
 		jdk 'openjdk-jdk17-latest'
 	}
+	environment {
+        SONARCLOUD_TOKEN     = credentials('SonarCloud token for gef')
```
ptziegler, line 15: I would suggest loading the credentials via [withCredentials()](https://www.jenkins.io/doc/pipeline/steps/credentials-binding/) in the `Build` stage

i.e.
```
steps {
  withCredentials([string(credentialsId: 'sonarcloud-token', variable: 'SONARCLOUD_TOKEN')]) {
    withSonarQubeEnv(credentialsId: 'sonarcloud-token', installationName: 'GEF Classic') {
      ...
    }
  }
}
```

azoitl:
In Jenkins ui I need both:
image

ptziegler:
> In Jenkins ui I need both:

The Jenkins UI and the pipeline syntax can be a little weird sometimes :P
But I'm pretty sure the id is required here. Side note, don't store sensitive data in environment variables. 


image

azoitl:
Thx for all the explanation. I'l gave it a try lets see!

Jenkinsfile
azoitl: @ptziegler as the latest update did also not work should I remove the environment section?

ptziegler: @azoitl Yeah. Now that the variable is already defined in the stage, this entry seems obsolete.

ptziegler:
I think I completely misunderstood the purpose of the `installationName` field.

>  SonarQube installation defined in this job (GEF Classic) does not match any configured installation

Obviously, the error message doesn't  show _which_ installations are available...

ptziegler:
@azoitl Can you access https://ci.eclipse.org/gef/configureTools? There should normally be an entry of the installed SonarQube Scanner instance.

ptziegler:
Just grasping at straws here, but the other job includes something like "GEF_Classic_SonarQube.io". Can you try out that one as installation?

ptziegler:
And if that doesn't work, perhaps simply "SonarCloud.io"...

ptziegler:
And now I'm out of ideas :(
I'm afraid the name is something the IT guys have to look up for us.

azoitl:
> @azoitl Can you access https://ci.eclipse.org/gef/configureTools? There should normally be an entry of the installed SonarQube Scanner instance.

Unfortunately not. I tried now: GEF_Classic_SonarQube.io, SonarCloud.io, My SonarQube Server will ask in Eclipse helpdesk. Keep you updated.

ptziegler:
On the bright side, this seems to be the final hurdle. I already consider the fact that the other job works a small victory :smile:

azoitl:
@ptziegler @fredg02 was so kind and provided a patch fixing our problem in PR #318. So finally we have sonar analysis running.

PR #314, opened 2023-12-13
 Specify different group id for features to avoid collisions with bundles 

ptziegler:
Both the GEF feature and the GEF bundle have the artifact id "org.eclipse.gef".
If both artifact have the same group id, they are considered to be the "same" artifact, from a Maven perspective.

Overwrite the group id for features to be "org.eclipse.gef.features", to make sure the generated Maven coordinates are distinct.

See #290

### DIFF
--- a/org.eclipse.draw2d-feature/build.properties
+++ b/org.eclipse.draw2d-feature/build.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2007, 2009 IBM Corporation and others.
+# Copyright (c) 2007, 2023 IBM Corporation and others.
 #
 # This program and the accompanying materials are made available under the 
 # terms of the Eclipse Public License 2.0 which is available at
@@ -12,3 +12,4 @@
 ###############################################################################
 bin.includes = feature.xml,\
                feature.properties
+pom.model.groupId = org.eclipse.gef.feature
--- a/org.eclipse.draw2d.sdk-feature/build.properties
+++ b/org.eclipse.draw2d.sdk-feature/build.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2007, 2009 IBM Corporation and others.
+# Copyright (c) 2007, 2023 IBM Corporation and others.
 #
 # This program and the accompanying materials are made available under the 
 # terms of the Eclipse Public License 2.0 which is available at
@@ -12,4 +12,4 @@
 ###############################################################################
 bin.includes = feature.properties,\
                feature.xml
-               
+pom.model.groupId = org.eclipse.gef.feature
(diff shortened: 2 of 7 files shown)


ptziegler:
@azoitl If this is alright with you, I'd like to play around with https://ci.eclipse.org/gef/job/GEF_Classic_SonarQube.io/ later today, using this branch as reference.

The issue is clear to me now, I just need to find a good solution for it...

org.eclipse.draw2d-feature/build.properties
```diff
@@ -11,4 +11,5 @@
 #     IBM Corporation - initial API and implementation
 ###############################################################################
 bin.includes = feature.xml,\
-               feature.properties
\ No newline at end of file
+               feature.properties
+pom.model.groupId = org.eclipse.gef.feature
```
azoitl, line 15: shouldn't this by draw2.feature

org.eclipse.draw2d.sdk-feature/build.properties
```diff
@@ -12,4 +12,4 @@
 ###############################################################################
 bin.includes = feature.properties,\
                feature.xml
-               
\ No newline at end of file
+pom.model.groupId = org.eclipse.gef.feature
```
azoitl, line 15: also here draw2d?

org.eclipse.zest-feature/build.properties
```diff
@@ -12,3 +12,4 @@
 ###############################################################################
 bin.includes = feature.xml,\
                feature.properties
+pom.model.groupId = org.eclipse.gef.feature
```
azoitl, line 15: here zest?

org.eclipse.zest.sdk-feature/build.properties
```diff
@@ -12,4 +12,4 @@
 ###############################################################################
 bin.includes = feature.properties,\
                feature.xml
-               
\ No newline at end of file
+pom.model.groupId = org.eclipse.gef.feature
```
azoitl, line 15: here zest?

org.eclipse.draw2d-feature/build.properties
ptziegler: I suppose that's down to preference...
Because e.g. the draw2d bundle also uses `org.eclipse.gef_root` as group id.
I can change it later today.

azoitl: I just wanted to be sure that we don't run into problems. with the same groupID. I think I need to learn more maven now.

ptziegler: Everything is fine as long as the group id-artifact id pair is unique.

It won't cause issues if we keep it like this, because the artifact ids (i.e. the feature ids) of all features are already distinct.

So from my perspective, the only reason to give the Draw2D and Zest features their own group id is for the sake of readability. But as I said, this comes down to preference and I don't have a strong opinion on what's the "cleaner" solution here.

azoitl: I think that keeping it as part of the GEF feature as their parent ID seems to me a good idea. Will merge it later. I have a few builds running that depend on the current GEF Classic latest.

PR #315, opened 2023-12-13
Only promote build results when on the master branch (#311)

ptziegler:
To avoid uploading the artifacts from old branches to the update site, we should make sure that we only do so, when on the master branch.


ptziegler:
@azoitl Those pull requests should prevent the 3.16, 3.17 and 3.18 braches from uploading their build artifacts to the update site.

PR #318, opened 2023-12-13
Fix Sonar configuration

fredg02:


azoitl:
Thx for the patch. It finally works.

PR #323, opened 2023-12-15
tell sonar to not check Messages classes

azoitl:


azoitl:
@ptziegler this removes app 250 sonar messages which is nice.

ptziegler:
This suppresses the following error, correct?

`squid:S3008 - Static non-final field names should comply with a naming convention.`

Given that the string fields must be mutable, I think that makes sense. A quick suggestion: Can you perhaps add the suppressed rule ("Static non-final fields...") to the commit message?

Otherwise people have to always look up the error code, before they understand this change.

azoitl:
Good point. Extended the commit message that this information is not lost.

ptziegler (approved):

PR #324, opened 2023-12-17
Update and integrate Zest 2.0 development branch into GEF

ptziegler:
This change merges the latest of Zest 2.0 into GEF. The sources have been taken from the following repository:

https://github.com/fsteeg/zest

The history has been preserved and remains unmodified. Only the latest commit adapts any compilation errors.

The major version of all Zest bundles has been increased to 2 to indicated downwards incompatibility.


ptziegler:
A few side notes:

1) I deliberately chose not do to a rebase, because this would require me to modify the original commits. I think we have to make an exception here and do a "normal" merge.

2) There are still a few open API issues. Most noticably LayoutAlgorithm.applyLayout(), which exists twice.

3) Given the size of this commit, we have to consult the EMO team for a review.

ptziegler:
Alternatively, I could move the "old" Zest plugins to a separate folder, do a normal rebase and then merge them back...

ptziegler:
Still not sure what the best solution is here...

As the diff indicates, this change primarily consists of additions so I'm tempted to just squash it into a single commit.

azoitl:
I'm not sure if we shoul really upgrade Zest to the 2.x version. There are users of Zest 1.x out there. Also we have improvements on Zest 1 in our repo (e.g. from @sebHollersbacher or @BauePhil). I don't think we want to loose them or?

ptziegler:
Urgh... this'll be a LOT of work. After closer inspection, just the Zest 2.0 Layouts bundle alone looks completely different, with half the classes missing.

Just look at e.g. [their](https://github.com/fsteeg/zest/blob/master/org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/LayoutAlgorithm.java) and [our](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/LayoutAlgorithm.java) LayoutAlgorithm interface. Completely different.

I'm closing this request, because I simply don't have an immediate answer for this.

azoitl:
I hoped there where less changes. But it is still nice to have your PR as reference so that we can cherry pick things that sound interesting.

ptziegler:
> I hoped there where less changes. But it is still nice to have your PR as reference so that we can cherry pick things that sound interesting.

I thought about a little bit more and I believe we can do a backport of the cloudio bundle. It also just occurred to me that the `org.eclipse.zest.jface` plugin is effectively a 1-to-1 copy of the `org.eclipse.zest.viewer` package from `org.eclipse.zest.core`.

I want to create a separate PR to integrate those changes and then, as you suggested, we can always cherry-pick the remaining parts as needed.

Doing it this would lessen the work by quite a lot, if we decide to touch this topic again.

azoitl:
Cloudio would definitely be cool.

PR #325, opened 2023-12-17
Typify Draw2D Graph package and resolve raw-type warnings

ptziegler:
The signature of following public methods has been changed:

- ShortestPathRouter.solve(...) Returns a List<? extends Path> instead of raw List
- NodeCluster.adjustRank(...) Expects a collection of NodeCluster indeaf of a raw collection

Resolves ~250 compiler warnings.

#155


azoitl (approved):
Looks good and definitely a great improvement to Zest.

PR #326, opened 2023-12-18
Remove Language Toolkit preferences for Zest bundles

ptziegler:
This preference is responsible for creating and updating the refactorings.index/refactorings.history files in the Zest bundles.

The task of tracking file changes is already done by Git, so it seems redundant to also have a separate tool for this. Especially when it's not consistently used throughout the workspace.


azoitl:
As discussed with PR #322 . Thx for the patch.

Issue #313, opened 2023-12-12
Our p2 repository is in dire need of modernization


ptziegler:
Our current setup is quite rudimentary and prone to issues ~~such as acidentally deleting all releases~~.

I think we can do a better job. As far as I can tell, GEF fx is using [justj](https://eclipse.dev/justj/?page=tools) to host their artifacts. We should probably do the same.

I propose that we use https://download.eclipse.org/tools/gef/classic/updates as the new location for our artifacts.

@merks Given that you've already set this up for other projects, do you have any advice on what I have to take into consideration? Otherwise I'd give it a try within the next few days.

Because as far as I understand, the org.eclipse.justj.p2.manager already does most of the work for me, so I hope this is a straightforward configuration.

merks:
This time you are one step ahead.  😜

I can help get this prototyped, if I know that folks are receptive.

merks:
I have a prototype working locally. 

![image](https://github.com/eclipse/gef-classic/assets/208716/89d468a2-2279-4828-a736-974a5b1b18b5)

I feel it would be nicer if we just reused the `classic` folder directly:

https://download.eclipse.org/justj/?file=tools/gef/classic

Just like we use the `fx` folder directly:

https://download.eclipse.org/justj/?file=tools/gef/fx/

I think we could do that without breaking anything because if we move every that is under here:

https://download.eclipse.org/justj/?file=tools/gef/classic

to here instead:

https://archive.eclipse.org/justj/?file=tools/gef/classic

the download.eclipse.org servers 404 handler will just redirect any `https://download.eclipse.org/tools/gef/classic/releases/*` URL to `https://archive.eclipse.org/tools/gef/classic/releases/*` as if it were a mirror.

So no existing links would be broken.

As an added bonus. I believe all the existing releases can be copied into the new folder structure so that we can immediately have a complete structure with releases, milestones, and nightly builds.

Does that sound okay?

ptziegler:
> So no existing links would be broken.

That's the only thing we definitely need to make sure.

Being able to keep the artifacts in _gef/classic_, rather than _gef/classic/updates_ sounds like a much cleaner solution. Especially if we are also able to migrate the already existing releases.

merks:
[...]

PR #333, opened 2023-12-19
Update links to p2 repositories in ReadMe file

ptziegler:
I've generalized the URL to a specific release, so that we don't have to update the ReadMe for every release.

Primarily because we've forgotten about it since 3.14 :P


ptziegler:
#313

azoitl (approved):

PR #336, opened 2023-12-22
Add missing parenthesis using the Eclipse clean-up feature

azoitl:
@ptziegler as suggested a full clean-up of all missing parenthesis. I hope this greatly reduces the noise for future fixes.

### DIFF
--- a/org.eclipse.draw2d.examples/sandbox/AlphaImageData.java
+++ b/org.eclipse.draw2d.examples/sandbox/AlphaImageData.java
@@ -36,9 +36,11 @@ public class AlphaImageData {
 		shell.addListener(SWT.Paint, event -> event.gc.drawImage(image, 0, 0, 1, 1, 40, 40, 40, 40));
 
 		shell.open();
-		while (!shell.isDisposed())
-			if (!display.readAndDispatch())
+		while (!shell.isDisposed()) {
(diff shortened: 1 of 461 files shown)


ptziegler:
I skimmed over it and everything seems to be good. I find it a little weird how the clean-up feature also cleaned-up the imports and changed for-loops into for-each loops, but I suppose that's fine.

azoitl:
> I skimmed over it and everything seems to be good. I find it a little weird how the clean-up feature also cleaned-up the imports and changed for-loops into for-each loops, but I suppose that's fine.

Ok that is interesting I created a configuration that should only clean the parenthesis. But cleaned for loops and imports is definitely nothing bad.

ptziegler:
> Ok that is interesting I created a configuration that should only clean the parenthesis. But cleaned for loops and imports is definitely nothing bad.

The only problem is that for the raw lists, we now iterate over Objects. At some point it probably makes sense to search for all explicit casts and see where those can be removed.

azoitl:
> The only problem is that for the raw lists, we now iterate over Objects. At some point it probably makes sense to search for all explicit casts and see where those can be removed.

Yes that is a point. I normally try to search for users of methods that I turn to generic. But it can easily be missed. Maybe there is a compiler warning we could turn on for that later.

PR #340, opened 2024-01-06
[Modernize Code] Typify GraphicalEditor Action Lists

azoitl:
Remove the ActionIDList and instead clearly define that the Action lists for selection, stack and property Actions can only hold Strings.

Addresses #155 

This PR may break client code if the lists where used wrongly.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
@@ -14,7 +14,6 @@ package org.eclipse.gef.ui.parts;
 
 import java.util.ArrayList;
 import java.util.EventObject;
-import java.util.Iterator;
 import java.util.List;
 
 import org.eclipse.swt.widgets.Composite;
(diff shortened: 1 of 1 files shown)


ptziegler:
> This PR may break client code if the lists where used wrongly.

Wouldn't client code break already, if they use something other than strings?

azoitl:
> > This PR may break client code if the lists where used wrongly.
> 
> Wouldn't client code break already, if they use something other than strings?

The existing code would at least gracefully accept IActions and use their id for adding. People using this feature would now get compile errors. However it would be an easy fix.

ptziegler:
> The existing code would at least gracefully accept IActions and use their id for adding. People using this feature would now get compile errors. However it would be an easy fix.

Ah, I didn't see that we immediately catch the execption we've just thrown. Thanks for the clarification. 😄

ptziegler:
I suppose there is not much that can be done... If people ignored this error for years, then it had to break eventually. Which is good that we fix it for the M1, giving downstream projects time to adjust their code, if necessary.

azoitl:
I also think that with the change the API is cleaner and better communicates the intention how it should be used. Thx for your reassurance.

PR #341, opened 2024-01-06
Fix path list return value for shortest path router

azoitl:
The shortest path router return a list of the routed paths. The change to a wildcard extend return was over pessimistic as Path is a class not to be extended. Therefore the list is now a list of Paths.

Addresses #155.

This fixes also a compile issue in GMF runtime: @pcdavid @ptziegler

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ShortestPathConnectionRouter.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ShortestPathConnectionRouter.java
@@ -269,10 +269,8 @@ public final class ShortestPathConnectionRouter extends AbstractRouter {
 			ignoreInvalidate = true;
 			processStaleConnections();
 			isDirty = false;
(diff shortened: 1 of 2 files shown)


ptziegler:
Probably makes sense to merge it before the M1, then.

pcdavid:
Thanks. The compilation error was fixed in GMF by https://github.com/eclipse-gmf-runtime/gmf-runtime/commit/24da7d3d84905b9e0e87d8958f18f087b3be297e.
It (GMF) still compiles with the reverted version, not sure if I'll revert too or leave it as it is.

azoitl:
@pcdavid sorry for the inconvenience. I checked to late with GMF runtime. My intention was that you don't have to do changes. I'll try to be again more careful. All latest changes from the last week are checked with GMF runtime to no break anything.

pcdavid:
No worry. Thanks for checking.

Issue #21, opened 2022-04-02
Clean-up Deprecated API Usage


azoitl:
Several of the API elements marked as deprecated are used through out the GEF code base. As a first step for clean-up it would be good to replace this usage with the according replacement APIs.

azoitl:
This will be a longer going effort. Therfore I reopen it it so that we have it as  notification in the open issue list.

azoitl:
While working through the different deprecated fields and methods in GEF Classic I came across [Point::getDistance2](https://github.com/eclipse/gef-classic/blob/8667a600701fab771e737916c64a889318552ffe/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Point.java#L200). This method calculates the squared distance between to points. The deprecation comment their states that instead `getDistance()` should be used and if needed the results squared. 

I checked the usage of `getDistance2`and it is mainly used to sort things according to distance. For this the getDistance2 has a much higher performance (sqrt, which is used in `getDistance2`is in general slow) and also a higher resolution. Therefore I don't think this`getDistance2`should be deprecated. If no one objects I would remove the deprecation from `getDistance2`and only change the return value from int to long. This is has no impact on the API and increases the utility.

Did I miss anything? Any opinion from @ptziegler, @Destrolaric, or @Phillipus?

ptziegler:
I really don't like the method name, so because of that alone, I would keep it deprecated. What about creating a new method `getDistanceSquared()` and have `getDistance2()` simply delegate to it?

l3002:
Hey @azoitl, I would like to contribute to this issue, just wanted to run that by you once. There wouldn't be any problem if I submit a PR in near future, would it?

wiresketch:
Please see #383 for the breaking change caused by the last merge.

azoitl:
@ptziegler I think with all the work that you did on marking deprecated API and cleaning-up we have a rather good state. Therefore I think we could close this task, or did I miss anything?

ptziegler:
I think that makes sense. Especially for such a broad issue, you'll never get to a point where "all" deprecated API has been touched.

PR #347, opened 2024-01-08
Removed usage of deprecated functions in draw2d core package

azoitl:
Addresses #21

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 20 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
```
ptziegler, line 2: Just a reminder, we're in 2024 already 😉

org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
```diff
@@ -73,23 +72,22 @@ private int calculateConnectionLocation(Point loc, Point topLeft, Point center)
...
+		if (loc.y < center.y) {
+			return 3;
 		} else {
-			// Connection start point along top or bottom
-			if (loc.y < center.y) {
-				return 3;
-			} else {
-				return 1;
-			}
+			return 1;
```
ptziegler, line 90: Is there a reason to keep the other "else" branches?

azoitl: This was done by the auto clean-up. Now that I see it also the else in line 82 should be removed. Will fix this,.

org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
azoitl: It seems I'm still living in the past ;-) will fix it.

PR #348, opened 2024-01-08
Update documentation regarding ICU removal

ptziegler:
We switched from the Bidi and BreakIterator implementation contributed by com.ibm.icu to the native Java implementation.

Because those implementations are not 100% compatible, we should note this in the changelog and also inform users about how they may be able to contribute their own implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,10 @@
 ## Draw2d
  - Removal of dependency to ICU. If required, downstream projects may contribute
    their own BreakIterator and Bidi implementations using the
-   `BreakIteratorProvider` and `BidiProvider`. See the documentation for
-   examples.
-
+   `BreakIteratorProvider` and `BidiProvider`.
+ - The selection, stack and property `Actions` of the `GraphicalEditor` can only
+   hold Strings and no longer support objects of type IAction.
+ 
 # GEF Classic 3.15.0 (Eclipse 2022-12)
 
 ## General
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,25 @@
+# GEF Classic 3.19.0 (Eclipse 2024-03)
+
+## Draw2d
+ - Removal of dependency to ICU. If required, downstream projects may contribute
+   their own BreakIterator and Bidi implementations using the
+   `BreakIteratorProvider` and `BidiProvider`. See the documentation for
+   examples.
+
 # GEF Classic 3.15.0 (Eclipse 2022-12)
 
-##General
+## General
  - Code clean-up and reduced deprecated code usage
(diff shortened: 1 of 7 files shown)


ptziegler:
I'm not good with words, so if anyone wants me to add more, feel free to tell me now 😅 

image

ptziegler:
I _love_ html, btw...

ptziegler:
I'm also a little bit confused by the release number. If we released the 3.15 with the 2022-12 and if we have four releases per year, shouldn't the 2024-12 be 3.20? Did we miss the release train at some point?

azoitl:
> I'm also a little bit confused by the release number. If we released the 3.15 with the 2022-12 and if we have four releases per year, shouldn't the 2024-12 be 3.20? Did we miss the release train at some point?

Yes we missed at least one. That was the reason I started with the check list issue for each release.

azoitl:
This is nice. Thanks. As we are on it should we also add a Draw2d Migration guide like the Eclipse does:
image

ptziegler:
> This is nice. Thanks. As we are on it should we also add a Draw2d Migration guide like the Eclipse does:

So you mean effectively writing the same again? That from 3.18 to 3.19, people have to write a service if they want to use the ICU implementation?

What else would I need to include? The only thing that comes to mind right now is 39b4b3232ddcb4c43c00787277c6f9c004ca0c27.

ptziegler:
Unrelated: I'd also like to migrate this stuff to AsciiDoc and then just generate the HTML on-the-fly, to make maintaining this whole stuff a little easier.

azoitl:
> > This is nice. Thanks. As we are on it should we also add a Draw2d Migration guide like the Eclipse does:
> 
> So you mean effectively writing the same again? That from 3.18 to 3.19, people have to write a service if they want to use the ICU implementation?
> 
> What else would I need to include? The only thing that comes to mind right now is [39b4b32](https://github.com/eclipse/gef-classic/commit/39b4b3232ddcb4c43c00787277c6f9c004ca0c27).

I think a shortend version. Something like draw2d has dropped ICU for the default Java implementation. If clients want to have different behavior they can and then link to your new documentation. I would see it more like an easy to find pointer that something has changed. 

As another topic I could see my PR #343.

azoitl:
> Unrelated: I'd also like to migrate this stuff to AsciiDoc and then just generate the HTML on-the-fly, to make maintaining this whole stuff a little easier.

that is a smart idea.

ptziegler:
I've added the removal of ICU and the changes to the GraphicalEditor action lists. Do you remember anything important I've missed?

azoitl:
I think for now we are fine. Thx a lot for the great work.

Issue #354, opened 2024-01-12
Giant cursors on Macbook with retina display


kysmith-csg:
Hello, I think this must be new as a result of the latest cursor updates, but the custom gef cursors are huge now on high-resolution displays (at least for Macs).

Normal cursor:
<img width="205" alt="normalcursor" src="https://github.com/eclipse/gef-classic/assets/136501322/df6aadfc-8e7b-4248-a4a9-664fea55452c">

Giant plug cursor:
<img width="215" alt="giantplug" src="https://github.com/eclipse/gef-classic/assets/136501322/18914298-b814-4a49-ba8c-5bb84743c7fc">

Giant add cursor:
<img width="296" alt="giantadd" src="https://github.com/eclipse/gef-classic/assets/136501322/0edf8d58-6f4a-4d4c-ac46-a5f263df32fc">

GEF is correctly loading the 2x versions of the cursors, but it seems that they are not being displayed in the correct size. I.e., a 2x version of a 16 bit icon should take up the same amount of _physical_ screen space on a retina display as a 1x 16 bit icon on a normal display, not just be twice as big.

I'm not very familiar with cursors. Are they designed to scale with the display resolution?

---

Eclipse GEF	GEF Classic Draw2d	3.14.100.202311222042	org.eclipse.draw2d
Eclipse GEF	GEF Classic GEF (MVC)	3.16.0.202311222042	org.eclipse.gef

Macbook Pro 2021
macOS Sonoma 14.2.1
16" Retina (resolution 3456 x 2234)
Apple M1 Max

Phillipus:
The problem is here:

https://github.com/eclipse/gef-classic/blob/586bead1bb52bba21210349ab6be90d4f11780aa/org.eclipse.gef/src/org/eclipse/gef/SharedCursors.java#L56-L74

On Mac the device zoom for cursors should always be 100.

This is how I implemented it in Archi:

https://github.com/archimatetool/archi/blob/master/com.archimatetool.editor/src/com/archimatetool/editor/diagram/tools/ExtCombinedTemplateCreationEntry.java

and

https://github.com/archimatetool/archi/blob/b5a2eef733d7c3a933d69f58d6452274ae0a5842/com.archimatetool.editor/src/com/archimatetool/editor/ui/ImageFactory.java#L72-L80

Phillipus:
So a check needs to added in `getDeviceZoom` to return 100 if on Mac

Phillipus:
[...]

PR #356, opened 2024-01-13
Fix cursors double-size on Mac and Linux 11

Phillipus:
See https://github.com/eclipse/gef-classic/issues/354

On Mac and Linux X11 device zoom should always be 100% when creating the ImageData for the cursor.


azoitl:
Thx for the fix @Phillipus. Didn't expect that this is necessary and I thought I checked with your original code first.

PR #362, opened 2024-01-14
Fixed all potential static method warnings in draw2d

azoitl:


ptziegler:
I get the feeling we're at a point where we simply suppress all "potentially static method" warnings...
Wouldn't it make more sense to set the severity to "ignore" instead?

azoitl:
> I get the feeling we're at a point where we simply suppress all "potentially static method" warnings... Wouldn't it make more sense to set the severity to "ignore" instead?

I know it now creates a bit of noise. However I already found some code where it made sense to turn it static. Also for future code it could lead to a bit more performance. So I have a tendency to clean that here now and keep it.

ptziegler:
> However I already found some code where it made sense to turn it static. Also for future code it could lead to a bit more performance. So I have a tendency to clean that here now and keep it.

That a fair point, I suppose.

PR #364, opened 2024-01-18
Make default hide delay for tooltip configurable

HeikoKlare:
The hide delay for tooltips in Draw2d is currently fixed to 5 seconds. The value can only be changed for a speific `ToolTipHelper`, which in turn is tied to a specific control. There is option to configure this delay globally.

With this change, the default value for the hide delay can globally be changed on the `ToolTipHelper` class, such that subsequent tooltip creations will use this value as default.

One could also think about more general configuration options, such as providing a system property, to make this delay configurable without a dependency to Draw2d. But since the demand for this configuration options does not seem to be high, I would expect the proposed lightweight realization to be sufficient (at least for us).


azoitl:
I like that kind of solution because clients can now use easily different measures to provide that value. E.g., a GEF based editor could offer an Eclipse property sheet.

HeikoKlare:
Thank you, @azoitl!

PR #366, opened 2024-01-23
Disable generation of JavaDoc as part of the incremental build

ptziegler:
This disables the Ant task (when inside the IDE) that is responsible for collecting the JavaDoc that is used by the online help. Reason being that its dependent "copy-resource" task isn't executed, leading to errors in a clean workspace.

Amends 60c6b5cda78241603ef99cc4ded08921f17cdaeb


ptziegler:
This is what the lifecycle mapping should look like, afterwards:

image

azoitl:
Thx for the quick fix. Tested with and without maven build. Works now perfectly.

PR #368, opened 2024-01-25
Added new Figure Container to serve as light weight drawing container

azoitl:
Using pure Figure just as layout container for other figures is very heavy weight as the graphics state is saved and restored very often. For pure layout containers this would not be necessary. This new figure can help to reduce memory footprint and drawing performance for these cases.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/Container.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/Container.java
@@ -1,0 +1,47 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Johannes Kepler University Linz
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alois Zoitl - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.draw2d;
+
+/**
+ * Lightweight Container which just draws the children according to the given
+ * layout.
+ *
+ * This Container does not maintain any graphics state and delegates that to its
+ * children. This can save memory and increase drawing performance for larger
+ * graphs with deep nested figures.
+ *
+ * A Container can not have any border and does not draw any background.
+ *
+ * This class is currently in development its API may
+ * change: @noreference, @noextend and @noinstantiate
+ *
+ * @since 3.16
+ */
+public class Container extends Figure {
+
+	public Container(LayoutManager manager) {
+		setLayoutManager(manager);
+	}
+
+	@Override
+	public final void paint(Graphics graphics) {
+		getChildren().stream().filter(IFigure::isVisible).forEach(child -> child.paint(graphics));
+	}
+
+	@Override
+	public final void setLayoutManager(LayoutManager manager) {
+		super.setLayoutManager(manager);
+	}
+
+}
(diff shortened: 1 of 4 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/Container.java
```diff
@@ -0,0 +1,39 @@
...
+ * graphs with deep nested figures.
+ *
+ * A Container can not have any border and does not draw any background.
+ *
+ * @since 3.15
+ */
+public class Container extends Figure {
+
+	public Container(LayoutManager manager) {
+		setLayoutManager(manager);
```
ptziegler, line 31: I'm not a huge fan of calling public methods inside the constructor, because subclasses could easily overwrite `setLayoutManager()` and cause a whole lot of undesired side-effects on a partially initialized object.

What do you think about making this method final as well?
i.e. by adding:

```
@Override
public final void setLayoutManager(LayoutManager manager) {
    super.setLayoutManager(manager);
}
```

azoitl: I also had planed to experiment using this class in different places in 4diac IDE and gradually refine it. Can/should I flag this class as experimental for the next release so that we still can fine tune its API?

ptziegler: > Originally I wanted to make the whole class final. However when searching in GEF Classic for use cases I noticed that there are use cases for subclassing. Therefore making this method final is a good idea.

I also like the idea a lot. I can already see it in the WindowBuilder project to represent e.g. the GridLayout or BorderLayout of a Composite.

> Can/should I flag this class as experimental for the next release so that we still can fine tune its API?

I don't know of a "proper" way to do this but the two options I'm aware of is to move this class to a separate package which is then exported as "provisional API", or to use things like the `@noreference` annotation.
See: https://wiki.eclipse.org/PDE/API_Tools/Restrictions

But in its current state, I think it's safe to keep it like that. You can generally overwrite or add new methods without breaking binary compatibility.

ptziegler (approved):

org.eclipse.draw2d/src/org/eclipse/draw2d/Container.java
azoitl: This sounds a bit over complicated. I only feared that we may want to make more methods final. But we have 4 more weeks to test and I plan to start using it in 4diac IDE as soon as it is merged. So I guess you are right we should be save.

ptziegler: Alternatively, you can always just keep it in 4diac for the time being and merge it into GEF, once you've refined the class.

ptziegler:
Wouldn't it be possible to mark classes as unstable using the `@noreference`, `@noextend` and `@noinstantiate` JavaDoc tags? This way we should be able to modify the class later on without breaking API (as clients use this class at their own risk).

https://wiki.eclipse.org/PDE/API_Tools/Architecture

azoitl:
> Wouldn't it be possible to mark classes as unstable using the `@noreference`, `@noextend` and `@noinstantiate` JavaDoc tags? This way we should be able to modify the class later on without breaking API (as clients use this class at their own risk).
> 
> https://wiki.eclipse.org/PDE/API_Tools/Architecture

This sounds interesting. So we would this now. Clients like 4diac IDE could still test it, but would get a warning. And when we think it is stable API either in 3.20 or 3.21 we would remove it and declare it as usable API. Am I right?

ptziegler:
> Clients like 4diac IDE could still test it, but would get a warning. And when we think it is stable API either in 3.20 or 3.21 we would remove it and declare it as usable API. Am I right?

That's how I understand it. Clients using this class would then do so at their own risk. Which is also why any changes won't be flagged by the API tools plugin.

azoitl:
done

ptziegler:
I checked out this branch together with the API tools and looks like the tags are properly recognized. Let's hope that they work the way I expect them to 😅

azoitl:
I hope that we can fix most of its API during this cycle.

PR #373, opened 2024-02-09
Wiki migration

vogella:
First commit moves the pages as is, second corrects the links I noticed that are wrong and removes a reference to an example I could not find anymore.


vogella:
cc @ptziegler @azoitl

vogella:
@azoitl or @ptziegler why can this PR not build?

merks:
It fails with:
```
[ERROR] Java heap space -> [Help 1]
```
I've seen that a lot latest on Platform PRs as well.

I've kicked it again so we can see if it fails the same way again or not...

merks:
It failed the same way.  I'm replaying it without the sonar thing.

merks:
The build finished really fast without the sonar thing.  It might be that the master build will start failing the same way as the PR...

vogella:
So fine to merge? I don't think these mark-down files can be the reason for the build failure.

merks:
Let's see what happens to the master build...

PR #374, opened 2024-02-09
Remove nested doSetUp() invocations due to SonarQube heap space error

ptziegler:


merks:
Still failing:

Exception in thread "OkHttp TaskRunner" java.lang.OutOfMemoryError: Java heap space

I tried to increase the heap space but my test build was aborted by another commit from Lars...

This still failed but failed differently:

image

I was going to try -Xmx3500m but it's not helpful when builds are aborted by commits...

ptziegler:
> I'm glad you are looking at this annoying problem!

Not much that can be done when the build doesn't work. 😅
Hm... But the wiki doesn't seem to be the source of the problem, which seems to be an OutOfMemoryError. I wonder if it is related to the Jenkins update a few weeks ago?

merks:
It's hard to tell what's changed.  I can only try to increase memory and hope, but I need a stable build machine.  I go out for a while.  Maybe you can try the heap space thing...

ptziegler:
> It's hard to tell what's changed. I can only try to increase memory and hope, but I need a stable build machine. I go out for a while. Maybe you can try the heap space thing...

I still believe that SonarQube chokes at one of our bundles. If I exclude everything, the scan (unsurprisingly) passes. I wonder what happens if I now start gradually including them again?

ptziegler:
> 2147483647 paths have been skipped in function org.eclipse.draw2d.test.ShortestPathRoutingTest#doSetUp(Lorg/eclipse/draw2d/geometry/Point;Lorg/eclipse/draw2d/geometry/Point;Lorg/eclipse/draw2d/geometry/Point;Lorg/eclipse/draw2d/geometry/Point;Lorg/eclipse/draw2d/geometry/Point;Lorg/eclipse/draw2d/geometry/Point;Lorg/eclipse/draw2d/geometry/Rectangle;)V.

I'd also run out of memory with that many paths...

merks:
What a nightmare.  😱

ptziegler:
If I understand the log correctly, then somehow SonarQube is no longer able to handle the Draw2D test suite... Which means that it might work if I simply exclude the project. 🤞

azoitl:
Unfortunately I also have no idea what is going on here. The only thing that I noticed is that yesterday sonar applied a new quality profile, which is finding much more things as you can see here: https://sonarcloud.io/project/activity?id=gef-classic

ptziegler:
> Unfortunately I also have no idea what is going on here. The only thing that I noticed is that yesterday sonar applied a new quality profile, which is finding much more things as you can see here: https://sonarcloud.io/project/activity?id=gef-classic

By now I'm almost certain that this is a bug within SonarQube itself... I believe that it somehow chokes at the doSetUp() -> doSetUp() -> doSetUp() chain which then leads to an infinite loop. Which is also why simply increasing memory does nothing to fix this issue.

azoitl:
> 
> By now I'm almost certain that this is a bug within SonarQube itself... I believe that it somehow chokes at the doSetUp() -> doSetUp() -> doSetUp() chain which then leads to an infinite loop. Which is also why simply increasing memory does nothing to fix this issue.

I think that in this case we should exclude the draw2d tests for now. I don't see that as our most important quality improvement target.

ptziegler:
> I think that in this case we should exclude the draw2d tests for now. I don't see that as our most important quality improvement target.

It's sufficient to only exclude ShortestPathRoutingTest, but that would also be my suggestion. I'll let this test run finish and then adjust the scanner.

ptziegler:
Inlining the method calls works, btw...

ptziegler:
I've adjusted the test to only have a single doSetUp() method. Unless there are any objections, I would suggest to go with that, because I don't really like the idea of making an exception for one specific file.

Is there any way to file a bug for this on the SonarQube issue tracker? As far as I understand, this requires a user account, meaning it would have to go through the EF if we wanted to pursue this any further...

azoitl:
> I've adjusted the test to only have a single doSetUp() method. Unless there are any objections, I would suggest to go with that, because I don't really like the idea of making an exception for one specific file.
> 

Thx for digging into it and fixing this.

> Is there any way to file a bug for this on the SonarQube issue tracker? As far as I understand, this requires a user account, meaning it would have to go through the EF if we wanted to pursue this any further...

I'm able to log into SonarQube with my Github account. Is this what you need?

ptziegler:
> I'm able to log into SonarQube with my Github account. Is this what you need?

I believe this is their issue tracker, for which you need to be a "SonarSourcer", whatever that's supposed to mean:
https://sonarsource.atlassian.net/jira/software/c/projects/SONAR/issues

And to file bugs, you apparently need to create threads on their forum:
https://community.sonarsource.com/c/sq/10

Urh, that already sounds like a lot of hoops to jump through...

Actually, we might not be the first ones with this problem:
https://community.sonarsource.com/t/project-analysis-error-details-java-lang-outofmemoryerror-java-heap-of-space/108925

PR #379, opened 2024-03-08
Use optimized scrolling for all three scroll operations of FigureCanvas

azoitl:
FigureCanvas' scrollToX and scrollToY uses an optimized scrolling functionality which leads especially for views with many figures to a much smoother scrolling experience. This commit utilizes this optimized scrolling also for scrollTo. Furthermore the code for all three has been combined in such a way that all three use the same core functions.

This work is triggered by the analysis of @BauePhil noticing that the `scrollTo()` resulted in a worse user experience in larger Zest graphs compared to the two `scrollToX()` and `scrollToY()` functions.  Today I finally found the reason for this. `scrollTo()` triggers **twice** (x and y) the range model resulting in triggering the propertyChange method of the Viewport, which invalidated the figure.  Already for mid-sized figures the change results in a user-noticeable difference.

I think this can be interesting for @pcdavid @Phillipus @ptziegler

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -189,10 +189,8 @@ public class FigureCanvas extends Canvas {
 	 */
 	@Override
 	public org.eclipse.swt.graphics.Point computeSize(int wHint, int hHint, boolean changed) {
-		// TODO Still doesn't handle scrollbar cases, such as when a constrained
-		// width
-		// would require a horizontal scrollbar, and therefore additional
-		// height.
+		// TODO Still doesn't handle scrollbar cases, such as when a constrained width
+		// would require a horizontal scrollbar, and therefore additional height.
 		int borderSize = computeTrim(0, 0, 0, 0).x * -2;
 		if (wHint >= 0) {
 			wHint = Math.max(0, wHint - borderSize);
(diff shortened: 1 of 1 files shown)


ptziegler:
For the sake of understanding, the performance problem is caused by the final else branch, because the ViewPort fires a PropertyChangeEvent for both the horizontal and vertical scrolling?

https://github.com/eclipse/gef-classic/blob/64bd301075c0b5f62bed57a09d027af5d865a5e5/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java#L374-L384

azoitl:
> For the sake of understanding, the performance problem is caused by the final else branch, because the ViewPort fires a PropertyChangeEvent for both the horizontal and vertical offset?
> 
> https://github.com/eclipse/gef-classic/blob/64bd301075c0b5f62bed57a09d027af5d865a5e5/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java#L374-L384

Yes. And the the code handling the PropertyChagenEvent is then doing a complete invalidation of the children, which may result in more propagated changes. To my understanding this is handled a bit more efficient in the scrollX and scrollY methods. But I may be wrong.

ptziegler:
> To my understanding this is handled a bit more efficient in the scrollX and scrollY methods. But I may be wrong.

With the new approach, the range model is updated while `ignoreScroll` is set, which then supresses revalidation.

https://github.com/eclipse/gef-classic/blob/64bd301075c0b5f62bed57a09d027af5d865a5e5/org.eclipse.draw2d/src/org/eclipse/draw2d/Viewport.java#L186-L199

The current state looks good, though I'm curious how noticeable this improvement will be on larger graphs. I'll try to test it myself in the next few days.

azoitl:
I would be really curios about your feedback. Because sofar my attempts on performance investigations resulted in mixed results and I'm mostly more confused then understanding.

ptziegler:
I had a look at it during the week, but I didn't really find any noticeable performance difference. I can't say whether that was due to the example was too small or due to other aspects.
Regardless, I like the idea of reusing the same piece of code, so I think we can merge it and then see how it goes.

ptziegler:
As mentioned, I think the implementation looks cleaner now. Let's see whether there is any noticeable performance improvement as well. :)

azoitl:
I'm still not perfectly sure what causes our performance issues. One thing that I notice is that the fix I proposed here reduces the redraw notifications. This reduces the redraws of a graphical outline. Which seems to be the main performance issue on our larger diagrams.

Issue #380, opened 2024-03-10
GEF 3.20 Release for Eclipse 2024-06


azoitl:
This issue is used to track and coordinate the work for the GEF Classic 3.20.0 release to be contributed to the Eclipse 2024-03 release train. It shall serve as checklist as well as template for the upcoming release.

- [x] Create 3.20 Milestone
- [x] Update version number and target platforms:
    - [x] Update the version number of all features to 3.20.0
    - [x] Bump maintenance version number of all plugins
    - [x] Update the target platform for the API baseline to GEF Classic 3.19.0
    - [x] Update the pom file target platform checking to GEF Classic 3.19.0
    - [x] Update the development target platform to Eclipse 2024-03
- [x] 10.04.2024: Release 3.20.0 M1 for 2024-06 M1
    - [x] create release record
- [x] 01.05.2024: Release 3.20.0 M2 for 2024-06 M2
- [x] 22.05.2024: Release 3.20.0 M3 for 2024-06 M3
- [x] 29.05.2024: Release 3.20.0 RC1 for 2024-06 RC1
- [ ] 05.06.2024: Release 3.20.0 for 2024-06 RC2
    - [x] create R3_20_maintenance branch
- [x] 12.06.2024: 2024-06 Release
    - [x] tag release commit and mark it as release in Github, create release notes

azoitl:
M1 Release is created and contributed to simrel:

- https://download.eclipse.org/tools/gef/classic/milestone/S202404101913
- https://github.com/eclipse-simrel/simrel.build/pull/312

azoitl:
M2 release is created and contributed to simrel:

- https://download.eclipse.org/tools/gef/classic/milestone/S202405011806/index.html
- https://github.com/eclipse-simrel/simrel.build/pull/330

azoitl:
M3 release is created and contributed to simrel:

- https://download.eclipse.org/tools/gef/classic/milestone/S202405222014/index.html
- https://github.com/eclipse-simrel/simrel.build/pull/360

azoitl:
Finally created the release record: https://projects.eclipse.org/projects/tools.gef/releases/gef-classic-3.19.0-2024-03

Feel free to extend and comment.

merks:
Sorry I did not catch this yesterday.  😱

azoitl:
@ptziegler I expected that when I saw the emails coming in this morning. However if I understand correctly we still have to wait for redoing M3 as there is no new certificate available yet.

[...]

PR #384, opened 2024-03-16
Updated target platform, baseline, and all versions for 3.20.0 developments

azoitl:
First step for #380.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.eclipse.gef_root</groupId>
 	<artifactId>org.eclipse.gef.releng</artifactId>
-	<version>3.19.0-SNAPSHOT</version>
+	<version>3.20.0-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<!-- this is the parent POM from which all modules inherit common settings -->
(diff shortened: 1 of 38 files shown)


pom.xml
```diff
@@ -5,20 +5,20 @@
...
+	<version>3.20.0-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<!-- this is the parent POM from which all modules inherit common settings -->
 	<properties>
 		<maven-antrun-plugin.version>1.7</maven-antrun-plugin.version>
 		<maven.compiler.source>17</maven.compiler.source>
 		<maven.compiler.target>17</maven.compiler.target>		
-		<tycho-version>4.0.4</tycho-version>
+		<tycho-version>4.0.6</tycho-version>
```
azoitl, line 16: @ptziegler I was not sure if this line could be removed as we have the tycho version anyhow in the maven.config

ptziegler: I think so. I'm not sure whether m2e is able to process the maven.config file but a quite test doesn't produce any workspace errors.
This would only be relevant for the documentation, as they are the only plugins using the Maven nature.

And in that case, I would even go one step further and also remove the `tycho.version` as well, and replace all references to it with `tycho-version`.

ptziegler:
Out of curiosity, is there a particular reason why you always bump the maintenance version of the plugins? I was always under the impression that this is only needed when there are actual code changes.

pom.xml
azoitl: removed it lets see what happens.

azoitl:
> Out of curiosity, is there a particular reason why you always bump the maintenance version of the plugins? I was always under the impression that this is only needed when there are actual code changes.

I thought the same when I started working on GEF Classic. However I then got PDE warnings and errors, and this seemed to be the best way. But I could be wrong here.

ptziegler:
> I thought the same when I started working on GEF Classic. However I then got PDE warnings and errors, and this seemed to be the best way. But I could be wrong here.

No, you're right... I'm already getting errors due to changes in the execution environment, which is evidently not the case. I'll just mark this down as the API tools being weird.
image

PR #385, opened 2024-03-16
Use PrecisionPoint instead of Point to handle deprecation warning

### LINKED ISSUE #383: Change from #347 break the FanRouter functionality
Change #347 that is seemingly only cosmetic, changes the code in FanRouter handleCollision method from:
```
		if (index % 2 == 0) {
			bendPoint = new Point(midPoint.x + (index / 2) * (-1 * ySeparation),
					midPoint.y + (index / 2) * xSeparation);
		} else {
			bendPoint = new Point(midPoint.x + (index / 2) * ySeparation,
					midPoint.y + (index / 2) * (-1 * xSeparation));
		}
		if (!bendPoint.equals(midPoint))
			points.insertPoint(bendPoint, 1);
```
to

```
		if (index % 2 == 0) {
			bendPoint = new Point(midPoint.x + (int) ((index / 2.0) * -1.0 * ySeparation),
					midPoint.y + (int) ((index / 2.0) * xSeparation));
		} else {
			bendPoint = new Point(midPoint.x + (int) ((index / 2.0) * ySeparation),
					midPoint.y + (int) ((index / 2.0) * -1.0 * xSeparation));
		}
		if (!bendPoint.equals(midPoint)) {
			points.insertPoint(bendPoint, 1);
		}
```

This breaks the connection drawing in the application that I'm developing. Here's a simple visual example.

Before:
before

After:
after

Clearly the new code change is not equivalent to the old one. The effect is that before the condition after the bendPoint calculation was false, but in the changed code it's calculated as true.

ptziegler:
This commit partially reverts fd75841fd7e0638c18827f203257e504536fb15e and uses PrecisionPoints instead of Points in the FanRouter class, to store the bend point.

Note that because the mid point is a plain point, the equality check is done using the integer coordinates.

Resolves #383

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FanRouter.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FanRouter.java
@@ -14,6 +14,7 @@ package org.eclipse.draw2d;
 
 import org.eclipse.draw2d.geometry.Point;
 import org.eclipse.draw2d.geometry.PointList;
+import org.eclipse.draw2d.geometry.PrecisionPoint;
 import org.eclipse.draw2d.geometry.Ray;
 
 /**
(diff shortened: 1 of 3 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/FanRouter.java
```diff
@@ -67,11 +68,11 @@ protected void handleCollision(PointList points, int index) {
 		Point bendPoint;
 
 		if (index % 2 == 0) {
-			bendPoint = new Point(midPoint.x + (int) ((index / 2.0) * -1.0 * ySeparation),
-					midPoint.y + (int) ((index / 2.0) * xSeparation));
+			bendPoint = new PrecisionPoint(midPoint.x + (index / 2) * (-1 * ySeparation),
```
azoitl, line 71: I think we should move (index/2) to a stack variable. Would make for me the code better readable, and the I wouldn't have made the wrong clean-up if it would have been that way.

ptziegler: Instead of having a separate variable, I can also just have it be part of the calculation for the `xSeparation` and `ySeparation`.

azoitl: Good idea.

azoitl (approved):

PR #388, opened 2024-03-29
Final Clean-up of shapes example. Mostly externalize strings

azoitl:


ptziegler:
Out of curiosity, is there any reason you prefer `MessageFormat.format(...)` over `Messages.bind(...)`?

azoitl:
> Out of curiosity, is there any reason you prefer `MessageFormat.format(...)` over `Messages.bind(...)`?

I guess the main reason is that I learned about i18n with using `MessageFormat.format(...)`. I personally don't have any preferences. Is `Messages.bind(...)` better?

ptziegler:
> > Out of curiosity, is there any reason you prefer `MessageFormat.format(...)` over `Messages.bind(...)`?
> 
> I guess the main reason is that I learned about i18n with using `MessageFormat.format(...)`. I personally don't have any preferences. Is `Messages.bind(...)` better?

They effectively do the same. The upside: `Messages.bind(...)` doesn't require an additional import. The downside: `Messages.bind(...)` doesn't support varargs. So if you have more than two parameters, you have to wrap them inside a `new Object[]{...}`.

azoitl:
> > > Out of curiosity, is there any reason you prefer `MessageFormat.format(...)` over `Messages.bind(...)`?
> > 
> > 
> > I guess the main reason is that I learned about i18n with using `MessageFormat.format(...)`. I personally don't have any preferences. Is `Messages.bind(...)` better?
> 
> They effectively do the same. The upside: `Messages.bind(...)` doesn't require an additional import. The downside: `Messages.bind(...)` doesn't support varargs. So if you have more than two parameters, you have to wrap them inside a `new Object[]{...}`.

Do you have any preferences? Personally I would be happy to try out `Messages.bind`

merks:
There's also org.eclipse.osgi.util.NLS.bind.  😱

azoitl:
More options, yeah! What is the Eclipse preferred way?

merks:
In the order of search matches:

image

image

image

I have no opinion.  😁

PR #389, opened 2024-03-29
Removed warnings from draw2d examples

azoitl:


org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/graph/GraphTests.java
```diff
@@ -660,22 +669,13 @@ public static DirectedGraph unstableGraph(int direction) {
 		nodes.add(a0 = new Node("node a0")); //$NON-NLS-1$
 		nodes.add(b0 = new Node("node b0")); //$NON-NLS-1$
 		nodes.add(c0 = new Node("node c0")); //$NON-NLS-1$
-		nodes.add(a1 = new Node("node a1")); //$NON-NLS-1$
-		nodes.add(b1 = new Node("node b1")); //$NON-NLS-1$
+		nodes.add(a1 = new Node(NODE_A1)); // $NON-NLS-1$
+		nodes.add(b1 = new Node(NODE_B1)); // $NON-NLS-1$
```
ptziegler, line 673: I'm a little bit confused. Earlier, you've moved the assignment to a separate line and now you keep it.

There are also some strings which are moved to static fields, while some are not. e.g "node a0". I assume that's only done for names that are used more than once?

The `// $NON-NLS-1$` comment can probably be deleted when there is no actual string in that line.

azoitl: I think I got lazy and overhelmed by that file. Let me get back to it on the weekend.

ptziegler: Hopefully this class doesn't need to be touched now, for a very long time 😅

azoitl: Yes I hoped so to. I'm not even sure if it is really used.  But as it builds up graphs for tests I didn't want to delete it.

Issue #391, opened 2024-04-02
Cannot release EditPolicies


pdeville2:
Hello,

In our tools Papyrus we have some memory leak issues, a part of this is that we are not able to release EditPolicies from the EditPart (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=583119).

I would like to add a method on org.eclipse.gef.editparts.AbstractEditPart to be able to release all EditPolicies. 

A pull request is coming soon for this.

pdeville2:
This seems not relevant since we should release EditPart instead.

PR #392, opened 2024-04-02
Open API to be able to remove all EditPolicies of an EditPart (#391)

pdeville2:
See issue #391

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/editparts/AbstractEditPart.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/editparts/AbstractEditPart.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2023, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -94,7 +94,7 @@ public abstract class AbstractEditPart implements EditPart, RequestConstants, IA
 	 * <code>null</code> values encountered.
 	 */
 	protected static class EditPolicyIterator {
-		private Object[] list;
+		private final Object[] list;
 		private int offset = 0;
 		private final int length;
 
@@ -109,6 +109,7 @@ public abstract class AbstractEditPart implements EditPart, RequestConstants, IA
 		 * @deprecated this constructor should not be used
 		 * @param list the list of policies.
 		 */
+		@Deprecated
 		public EditPolicyIterator(List list) {
 			this(list.toArray());
 		}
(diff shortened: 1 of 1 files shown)


ptziegler:
That sounds like a very specific use-case. What about a more generic approach using the Visitor pattern?

There was a similar problem with a different project and we ended up writing an [EditPartVisitor](https://github.com/eclipse-windowbuilder/windowbuilder/blob/master/org.eclipse.wb.core/src-gef/org/eclipse/wb/internal/gef/core/EditPartVisitor.java) 
```
public class EditPartVisitor {
	/**
	 * This method is invoked to check if given {@link EditPart} and its children should be visited.
	 */
	public boolean visit(EditPart editPart) {
		return true;
	}

	/**
	 * This method is invoked when all children of given {@link EditPart} were visited.
	 */
	public void endVisit(EditPart editPart) {
	}
}
```

And a matching `accept()` method in [EditPart](https://github.com/eclipse-windowbuilder/windowbuilder/blob/master/org.eclipse.wb.core/src-gef/org/eclipse/wb/gef/core/EditPart.java#L181)

```
public final void accept(EditPartVisitor visitor) {
	if (visitor.visit(this)) {
		for (EditPart childPart : getChildren()) {
			childPart.accept(visitor);
		}
		visitor.endVisit(this);
	}
}
```

In your situation, you could simply traverse through the edit parts and call `deactivate()` on all relevant edit parts. Via the return value, you can then control whether the visitor should also traverse through the children.

@pdeville2 @azoitl 
How does that sound? The code is already there, so creating a PR should be simple enough.

azoitl:
@ptziegler I personally like the `EditPartVisitor` approach more. As you see it seems to be a very specific use case.

On may way home from work I thought a lot on the problem. From my own experience digging into GEF Classic's memory consumption and  performance issues I know that this is hard and complicated. 

Although I don't know to much about the Papyrus' and GMF's usage of GEF Classic I strongly believe that the proposed solution is only curing the symptom but not the cause. EditPolicies are solely living as part of their assigned EditPart. If they are still around it means also their EditParts are still referenced by something. So I think @pdeville2 you first have to find out why the EditParts are kept around.

pdeville2:
Thanks for your quick answer.

I agree with @azoitl that the main problem is our living EditParts, I take the problem in a bad way since normally if EditParts are removed, EditPolicies should be Garbage collected.

I keep in mind the @ptziegler's proposal, but calling deactivate() is not enough because the EditPolicy will be still here.

I will close the bug.

azoitl:
@pdeville2 happy to help. I'm always interested in improving GEF Classic performance and memory consumption, as I have similar issues in 4diac IDE. There I have GEF Classic based editors showing thousands of Figures. So if you have any suggestions for or news for us I'm happy to hear about them.

PR #395, opened 2024-04-05
Handle implicit ClassCastException in Zest Core due to raw types

ptziegler:
The selection of an abstract viewer can either be of type GraphNode or GraphConnection. However, this list of items is explicitly casted to an array of GraphNodes. Use the common base class GraphItem instead.

Furthermore, the viewer attempts to remove the objects of type GraphConnection from the layout manager, whenever a node is removed. But the layout manager only accepts objects of type LayoutRelationship which is a member of the GraphConnection.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/internal/AbstractStructuredGraphViewer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2005, 2011 CHISEL Group, University of Victoria, Victoria, BC,
+ * Copyright 2005, 2024 CHISEL Group, University of Victoria, Victoria, BC,
  *                      Canada.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


ptziegler:
Makes me wonder is how this ever worked up until now...

azoitl (approved):

azoitl:
> Makes me wonder is how this ever worked up until now...

Me too...

ptziegler:
> > Makes me wonder is how this ever worked up until now...
> 
> Me too...

I have the impression two different design philosophies were implemented in the Zest component...

(I) Nodes and edges are contributed via a custom content provider. The actual nodes and edges are created outside the viewer.

II) The nodes are created via the non-API `createConnection` and `createNode` methods.

Approach (I) is what's described in the documentation and the only thing that's used in the tests/examples.
Approach (II) is addressed by this PR.

I _assume_ the intended way was for Zest to construct an internal representation of the graph using the nodes and connections. At least that's the impression I got when looking at the way it's handled in GEF FX. The reason it doesn't cause issues up until now is simply because this step was never implemented...

ptziegler:
That impression seems to have been wrong...
After looking deeper into it, the first ClassCastException doesn't occur, because `GraphViewer` overwrites the flawed `setSelectionToWidget()` from `AbstractStructuredGraphViewer`.
The second ClassCastException doesn't occur, because it's not possible to remove a node from the viewer. When the input is changed, a new graph is constructed from scratch, so there are never individual nodes being removed.

PR #400, opened 2024-04-07
Disable execution of tycho-packaging-plugin in Eclipse IDE

ptziegler:
The package-plugin mojo can't be executed within the Eclipse. Attempting to do only creates an error marker in the isv bundles.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
 			</plugin>
 			
 			<plugin>
+				<?m2e ignore?>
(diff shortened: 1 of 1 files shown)


ptziegler:
This is the error I frequently get in my workspace:

image

This is the updated lifecycle mapping with this fix:

image

azoitl:
Thx for finding a solution for this issue. This really also bothered me.

azoitl (approved):

PR #404, opened 2024-04-08
Added github-action checking to dependabot

azoitl:


ptziegler (approved):

ptziegler:
I finally figured out where those dependabot logs are stored:

Insights -> Dependency Graph
https://github.com/eclipse/gef-classic/network/updates

Unsurprisingly, it doesn't understand the Tycho pomless builds...
image

ptziegler:
And looking at my inbox... it works 🫠

azoitl:
> I finally figured out where those dependabot logs are stored:
> 
> Insights -> Dependency Graph https://github.com/eclipse/gef-classic/network/updates
> 
> Unsurprisingly, it doesn't understand the Tycho pomless builds... image

But shouldn't it at least find the root pom?

ptziegler:
> But shouldn't it at least find the root pom?

It does, which is why it finds around 30 pom dependencies. But when it tries to process the modules, an error is thrown.

azoitl:
> > But shouldn't it at least find the root pom?
> 
> It does, which is why it finds around 30 pom dependencies. But when it tries to process the modules, an error is thrown.

Ah, thx for the clarification.

PR #410, opened 2024-04-09
Reintegration of Zest2.0 fork

ptziegler:
At the end of the day, I don't believe (or rather I think it's harmful) to enforce a linear history for a fork that has evolved independently of our current Zest implementation for the past 15 years.

If I did so, most of the commits would look completely different to what the original authors had in mind, which I don't think is appropriate and the main reason why I keep both histories separate.

The current implementation only adds the new interfaces, but is not yet used internally.


ptziegler:
I want to create a proper issue tracker later today to keep track of all remaining tasks. The goal is to be done by 2024-09.

PR #413, opened 2024-04-09
Migrate Directed Graph Example by Anthony Hunter to Markdown

ptziegler:
Those examples were previously available via the following webpages.

http://www.eclipse.org/gef/reference/digraph1.php
http://www.eclipse.org/gef/reference/digraph2.php

The example has been converted to Markdown and moved to our GitHub repository.


azoitl:
Great idea to migrate that. Should we move that to the wiki or to draw2d docs?

ptziegler:
In hindsight, it probably makes more sense to move it directly to the wiki.

ptziegler:
Done:
https://github.com/eclipse/gef-classic/wiki/Directed-Graph-Example-(Digraph1)
https://github.com/eclipse/gef-classic/wiki/Directed-Graph-Example-(Digraph2)

PR #415, opened 2024-04-10
Scroll redraw on Windows lead to drawing artifacts

azoitl:
The reworked scrolling of the FigureCanvas didn't correctly calculated the exposure areas which need to be redrawn after scrolling. This led to drawing artifacts.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
@@ -411,24 +411,20 @@ public class FigureCanvas extends Canvas {
 		int dy = -vOffset + getViewport().getViewLocation().y;
 		Rectangle clientArea = getViewport().getBounds().getShrinked(getViewport().getInsets());
 		Rectangle blit = clientArea.getResized(-Math.abs(dx), -Math.abs(dy));
-		Rectangle expose = clientArea.getCopy();
 		Point dest = clientArea.getTopLeft();
 
-		expose.width = Math.abs(dx);
+		blit.width = Math.max(blit.width, 0);
+		blit.height = Math.max(blit.height, 0);
+
 		if (dx < 0) { // Moving left?
 			blit.translate(-dx, 0); // Move blit area to the right
-			expose.x = dest.x + blit.width;
-		} else {
-			// Moving right
+		} else { // Moving right
 			dest.x += dx; // Move expose area to the right
 		}
 
-		expose.height = Math.abs(dy);
 		if (dy < 0) { // Moving up?
 			blit.translate(0, -dy); // Move blit area down
-			expose.y = dest.y + blit.height; // Move expose area down
-		} else {
-			// Moving down
+		} else { // Moving down
 			dest.y += dy;
 		}
 
@@ -438,7 +434,7 @@ public class FigureCanvas extends Canvas {
 		getViewport().setHorizontalLocation(hOffset);
 		getViewport().setVerticalLocation(vOffset);
 		getViewport().setIgnoreScroll(false);
-		redraw(expose.x, expose.y, expose.width, expose.height, true);
+		scrollRedraw(dx, dy, clientArea, blit);
 	}
 
 	private void scrollChildren(int dx, int dy, Rectangle blit, Point dest) {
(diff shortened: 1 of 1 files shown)


BauePhil:
This has fixed the artifacting, that I could previously reproduce reliably with a big zest graph, caused by quick scroll direction changes.

azoitl:
@Phillipus I guess this could be interesting for you.

I would love to have this in the M1 that is due today, if possible.

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
```diff
@@ -462,6 +463,26 @@ private void scrollChildren(int dx, int dy, Rectangle blit, Point dest) {
 		}
 	}
 
+	private void scrollRedraw(int dx, int dy, Rectangle clientArea, Rectangle blit) {
+		Rectangle expose = clientArea.getCopy();
+		if (dx != 0) {
+			expose.width = Math.abs(dx);
+			if (dx < 0) { // Moving left?
+				expose.x += blit.width; // Move expose area right
```
ptziegler, line 471: This used to be `expose.x = dest.x + blit.width`, but given that `dest = clientArea.getTopLeft()` and `expose = clientArea.copy()`, I assume this section is identical to the current implementation?

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
```diff
@@ -411,24 +411,25 @@ private void performScroll(int hOffset, int vOffset) {
...
-		Rectangle expose = clientArea.getCopy();
 		Point dest = clientArea.getTopLeft();
 
-		expose.width = Math.abs(dx);
+		if (blit.width < 0) {
+			blit.width = 0;
+		}
+
+		if (blit.height < 0) {
+			blit.height = 0;
```
ptziegler, line 421: Wouldn't it be cleaner to write this as:
```
blit.width = Math.max(blit.width, 0);
blit.height = Math.max(blit.height, 0);
```

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
```diff
@@ -462,6 +463,26 @@ private void scrollChildren(int dx, int dy, Rectangle blit, Point dest) {
...
+		}
+		if (dy != 0) {
+			expose.x = clientArea.x;
+			expose.width = clientArea.width;
+			expose.height = Math.abs(dy);
+			if (dy < 0) { // Moving up?
+				expose.y += blit.height; // Move expose area down
+			}
+			redraw(expose.x, expose.y, expose.width, expose.height, true);
+		}
```
ptziegler, line 483: And this block is now the new code? Which, as far as I understand, redraws the -now visible- area?

ptziegler (commented):
To be honest, I already have a hard time understand the current implementation 😅

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
azoitl: Yes the problem of the old code was that a way to small area was redrawn. This resulted under windows under extreme drawing artifacts.

azoitl:
> To be honest, I already have a hard time understand the current implementation 😅

I totally understand that. Spent today quite some time to understand the code and to find the reason for the problems.

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
ptziegler: I see... I can't extensively test but from what I've heard, it works. At worst, we paint more than is needed, so I  don't mind giving it a go.

ptziegler (approved):

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
azoitl: In the case when both x and y are scrolled then we draw the overlapping square to much. But I think calculating that is more effort and makes the code much more unreadable.

PR #419, opened 2024-04-12
Changed RulerLayout to directly inherit from AbstractLayout  #418

### LINKED ISSUE #418: Ruler guides are broken in  GEF 3.19.0.202402212051
When using ruler guides the following exception is constantly being logged and ruler guides do not work:

```
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "org.eclipse.gef.internal.ui.rulers.RulerLayout.getConstraint(org.eclipse.draw2d.IFigure)" is null
	at org.eclipse.gef.internal.ui.rulers.RulerLayout.layout(RulerLayout.java:55)
	at org.eclipse.draw2d.Figure.layout(Figure.java:1201)
	at org.eclipse.draw2d.Figure.validate(Figure.java:2060)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.eclipse.draw2d.Figure.validate(Figure.java:2061)
	at org.eclipse.draw2d.Viewport.validate(Viewport.java:378)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.eclipse.draw2d.Figure.validate(Figure.java:2061)
	at org.eclipse.draw2d.DeferredUpdateManager.performValidation(DeferredUpdateManager.java:225)
	at org.eclipse.draw2d.DeferredUpdateManager.paint(DeferredUpdateManager.java:167)
	at org.eclipse.draw2d.LightweightSystem.paint(LightweightSystem.java:202)
	at org.eclipse.draw2d.LightweightSystem.lambda$0(LightweightSystem.java:110)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1529)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1555)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1538)
	at org.eclipse.swt.widgets.Control.gtk_draw(Control.java:3927)
	at org.eclipse.swt.widgets.Scrollable.gtk_draw(Scrollable.java:365)
	at org.eclipse.swt.widgets.Composite.gtk_draw(Composite.java:506)
	at org.eclipse.swt.widgets.Canvas.gtk_draw(Canvas.java:174)
	at org.eclipse.swt.widgets.Widget.windowProc(Widget.java:2482)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:6883)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:6162)
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_do_event(Native Method)
	at org.eclipse.swt.widgets.Display.eventProc(Display.java:1598)
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_iteration_do(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4514)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
```

The issue is introduced in commit 4635985b7045db7bd80758387c7c360aa7eda305 from Dec 30, 2023 which apparently only modernizes code, but breaks the ruler functionality. The problem is with the change in XYLayout where constraints are forced to be Rectangles. Since setConstraint method takes an object the change adds an instanceof check and silently rejects all other types of constraints:

```
	public void setConstraint(IFigure figure, Object newConstraint) {
		super.setConstraint(figure, newConstraint);
		if (newConstraint instanceof Rectangle rect) {
			constraints.put(figure, rect);
		}
	}
```

Previously this method looked like this:

```
	public void setConstraint(IFigure figure, Object newConstraint) {
		super.setConstraint(figure, newConstraint);
		if (newConstraint != null)
			constraints.put(figure, newConstraint);
	}
```

This change breaks RulerLayout class which is a subclass of XYLayout and which uses Integer constraints instead of Rectangle.

This issue is really critical since I couldn't not find a way to add a workaround in my app. The fix could either revert this change, or perhaps make RulerLayout inherit directly from AbstractLayout to remove the dependency on XYLayout.

I am really worried about all this changes lately. Code modernization is nice, but doing it just for the sake of it feels unnecessary to me. On the other hand this is the second breaking change I find, and this just in the code I use, so this trend really worries me.

azoitl:
`RulerLayout` inherited from `XYLayout`. The later assumed that children constraint are rectangles. As this is not valid for `RulerLayout`s the `RulerLayout` was changed to inherit from `AbstractLayout` and use its own constraint map with Integers.

Fixes #418

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/XYLayout.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/XYLayout.java
@@ -136,7 +136,11 @@ public class XYLayout extends AbstractLayout {
 	@Override
 	public void setConstraint(IFigure figure, Object newConstraint) {
 		super.setConstraint(figure, newConstraint);
-		if (newConstraint instanceof Rectangle rect) {
+		if (newConstraint != null) {
+			if (!(newConstraint instanceof Rectangle rect)) {
+				throw new IllegalArgumentException("XYLayout was given " + newConstraint.getClass().getName() //$NON-NLS-1$
+						+ " as constraint for Figure. Rectangle expected!"); //$NON-NLS-1$
+			}
 			constraints.put(figure, rect);
 		}
 	}
--- a/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerLayout.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerLayout.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2010 IBM Corporation and others.
+ * Copyright (c) 2003, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -12,8 +12,11 @@
  *******************************************************************************/
 package org.eclipse.gef.internal.ui.rulers;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.draw2d.AbstractLayout;
 import org.eclipse.draw2d.IFigure;
-import org.eclipse.draw2d.XYLayout;
 import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.draw2d.geometry.Rectangle;
 
@@ -25,7 +28,9 @@ import org.eclipse.draw2d.geometry.Rectangle;
  * @author Pratik Shah
  * @since 3.0
  */
-public class RulerLayout extends XYLayout {
+public class RulerLayout extends AbstractLayout {
+
+	private final Map<IFigure, Integer> constraints = new HashMap<>();
 
 	/**
 	 * @see org.eclipse.draw2d.AbstractLayout#calculatePreferredSize(org.eclipse.draw2d.IFigure,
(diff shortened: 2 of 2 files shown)


azoitl:
@wiresketch thx in advance for cross checking.

org.eclipse.draw2d/src/org/eclipse/draw2d/XYLayout.java
```diff
@@ -136,7 +136,11 @@ public void remove(IFigure figure) {
 	@Override
 	public void setConstraint(IFigure figure, Object newConstraint) {
 		super.setConstraint(figure, newConstraint);
-		if (newConstraint instanceof Rectangle rect) {
+		if (newConstraint != null) {
+			if (!(newConstraint instanceof Rectangle rect)) {
```
ptziegler, line 140: That nested if looks nasty. And wouldn't it be better to throw the exception before calling super.setConstraint()?

I.e.
```
if (newConstraint instanceof Rectangle) {
  ....
} else {
  throw new IllegalArgumentException()
}
```

azoitl: ` newConstraint` can be null and should then be ignored. At least this was how the code worked before.

ptziegler: Good point...

ptziegler (approved):

PR #421, opened 2024-04-13
[Modernize Code] Typify Class in DirectEditManager

azoitl:
Addresses: #155

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/tools/DirectEditManager.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/tools/DirectEditManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -23,6 +23,7 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 
+import org.eclipse.core.runtime.Assert;
 import org.eclipse.jface.viewers.CellEditor;
 import org.eclipse.jface.viewers.ICellEditorListener;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.gef/src/org/eclipse/gef/tools/DirectEditManager.java
```diff
@@ -146,15 +148,15 @@ protected void commit() {
...
 	 *
 	 * @param composite the composite to create the cell editor on
 	 * @return the newly created cell editor
 	 */
 	protected CellEditor createCellEditorOn(Composite composite) {
 		try {
-			Constructor constructor = editorType.getConstructor(Composite.class);
-			return (CellEditor) constructor.newInstance(composite);
+			Constructor<? extends CellEditor> constructor = editorType.getConstructor(Composite.class);
+			return constructor.newInstance(composite);
```
ptziegler, line 159: From what I understand, subclasses either contribute the cell editor via the constructor or by overwriting this method (in which case `null` is usually passed to the constructor).

If `null` is passed but this method is not overwritten (or super.createCellEditorOn()) is called), the resulting `NullPointerException` is suppressed.

This sounds like an illegal state to me. My expectation would be that an exception should be thrown, as this sounds like an oversight, rather than a proper use-case.

ptziegler (approved):
I tried to come up with a view ways

ptziegler:
No idea how to un-approve a commit. I forgot that I had it checked when I came up with a specific edge-case...

org.eclipse.gef/src/org/eclipse/gef/tools/DirectEditManager.java
azoitl: Yes this is a good point. Should we just catch Reflection exceptions then?
Should we add an Assert not null?

ptziegler: > Should we just catch Reflection exceptions then?

Catching only the reflection exceptions sounds a lot better than all exceptions.

> Should we add an Assert not null?

Sounds good to me.

azoitl: Done!

ptziegler (approved):

PR #423, opened 2024-04-15
Add test cases for XYLayout and RulerLayout constraints

ptziegler:
Check whether Rectangle/Integer are valid constraints and whether an error is thrown if the contract is violated.

See https://github.com/eclipse/gef-classic/issues/418


ptziegler:
Doing a full integration tests with a Rule project seems quite excessive. But those tests should at least avoid the regressions described in that issue, in the future.

ptziegler:
I'll add tests for the `layout()` methods within the next few days...

ptziegler:
Now that I think about it... what about the other layouts? As far as I can tell, we don't validate the input for e.g. the BorderLayout.

azoitl (approved):

Issue #426, opened 2024-04-17
Make AbstractTool's accessibleStep mechanism accessible for subclasses


lredor:
Currently, it's not possible to use the methods "accStepReset()", "accStepIncrement()" and "accGetStep()" in AbstractTool subclasses. The goal of this issue is to make these methods protected to allow its use in a subclass.

The final goal is to use them in a subclass of Sirius for issue [359](https://github.com/eclipse-sirius/sirius-desktop/issues/359).

azoitl:
In general I'm fine with this change. However quickly reading over the Sirius bug report I wanted to ask if not more of this problem should be solved on GEF Classic side? 


Also please note for API changes you need to add documentation to the now API functions including the `@since 3.18` tag.

PR #427, opened 2024-04-17
Make AbstractTool's accessibleStep mechanism accessible for subclasses #426

lredor:
Make AbstractTool's accessibleStep mechanism accessible for subclasses #426

Fixes https://github.com/eclipse/gef-classic/issues/426


azoitl:
In general I'm fine with this change. However quickly reading over the Sirius bug report I wanted to ask if not more of this problem should be solved on GEF Classic side? 


Also please note for API changes you need to add documentation to the now API functions including the `@since 3.18` tag.

ptziegler:
If I understand this correctly, the Sirius issue contains of two components:

(I)

> At handleKeyDown call, org.eclipse.gef.tools.AbstractTool.accStepIncrement() computes the increment according to the duration of the key press (from 1 to 16 to have an acceleration of the move). This can explain a different behavior according the duration of the key press (first gap is from one to four).

(II)

> Globally, the current GEF/GMF implementation does not allow solving the problem. This implementation is based on a simulation of a mouse move when one of the arrow key is pressed. But this implementation mixes absolutes GMF coordinates and mouse cursor coordinates that uses "screen" coordinates (for example in org.eclipse.gef.tools.AbstractTool.placeMouseInViewer(Point) by using org.eclipse.swt.widgets.Control.toDisplay(Point)).

The second part is going to the fixed in the Sirius implementation, while the first part is handled by this PR. Did I get this right so far?

In general, I agree with @azoitl. I think it's fine to make those classes accessible, but given that they aren't used anywhere else, they are better off being moved to `DragEditPartsTracker`.

lredor:
> In general I'm fine with this change. However quickly reading over the Sirius bug report I wanted to ask if not more of this problem should be solved on GEF Classic side?

@azoitl Absolutely, you are right. It would be better to deal with this problem directly on the GEF side. However, there is already an overlay on the GMF side (done for bugzilla [111901](https://bugs.eclipse.org/bugs/show_bug.cgi?id=111901)). And Sirius is based on GMF.
I am not sure I can control all the impacts by making a change at the GEF level. I'm talking about the impacts on different GEF applications.
I therefore prefer to make these changes on the Sirius side. I have made progress on the development on the Sirius side. I could try to do an equivalent PR on the GEF side (but that would also involve a PR on the GMF side).

> Also please note for API changes you need to add documentation to the now API functions including the `@since 3.18` tag.

I'll do that.

lredor:
> I just saw that the functionality in question is only used by the `DragEditPartsTracker`. As we are cleaning up. Should we move the functions in questions to the `DragEditPartsTracker` as how understand it they only make sense there. Or did I misunderstand the situation?

No it's also used in `org.eclipse.gef.tools.SimpleDragTracker` (for example [here](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/tools/SimpleDragTracker.java#L188) for `accStepIncrement()`).

lredor:
> The second part is going to the fixed in the Sirius implementation, while the first part is handled by this PR. Did I get this right so far?

@ptziegler You understood well. In addition, there is not really a problem solved with this PR. This just allows me to reuse existing behavior on Sirius side.

lredor:
A new version of PR with the javadoc of the "new" API with @since 3.18 tag. Is there additional documentation like "Release Notes" to complete?

ptziegler (approved):
lgtm. @azoitl Anything from your side?

>  Is there additional documentation like "Release Notes" to complete?

If you might, you can add something along the lines of  "Clients can now configure the accessible steps of tools." to the [CHANGELOG.MD](https://github.com/eclipse/gef-classic/blob/master/CHANGELOG.md)

azoitl:
> lgtm. @azoitl Anything from your side?

looks also good to me. And sorry for not noticing that the common base class I thought we could move the methods is an interface.

azoitl:
> > If you might, you can add something along the lines of "Clients can now configure the accessible steps of tools." to the [CHANGELOG.MD](https://github.com/eclipse/gef-classic/blob/master/CHANGELOG.md)
> 
> I've added since `@since 3.18` in the javadoc of methods, but in changelog.md I saw that the 2024.03 version is 3.19. So I imagine that I should have written `@since 3.20`. No ? 

This is a bit complicated. The next GEF Classsic version for 2024.06 will be 3.20. However the individual plugins have different versions. The GEF plugin, as show in its manifest, is at 3.18. Therefore `since 3.18`is correct. The Eclipse PDE tools help you with that.

> And I also imagine that I have to create a new 3.20 section in the CHANGELOG.MD ?

Yes please.

azoitl (approved):

azoitl:
Thx @lredor for the update in [CHANGELOG.MD](https://github.com/eclipse/gef-classic/blob/master/CHANGELOG.md)

PR #428, opened 2024-04-17
Fix redraw performance issues in FigureCanvas when GTK overlay scrolling is enabled (eclipse#425)

wiresketch:
Fix for eclipse#425. Detect when overlay scrolling is enabled on GTK and it's the case then ignore scrollbar trims when calculating the viewport client area.

This fix tackles the immediate performance issue however it adds a platform dependency (which isn't the first in GEF), but also is not future proof because the upcoming GEF4 removes the GTK_OVERLAY_SCROLLING environment variable support which is being checked here. Since SWT doesn't provide a way to test for GTK version, a more complete fix would involve a better support for GTK_OVERLAY_SCROLLING in SWT's Scrollable.getScrollbarsMode() method.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
@@ -300,9 +300,17 @@ public class FigureCanvas extends Canvas {
 
 	private void layoutViewport() {
 		ScrollPaneSolver.Result result;
+		int vBarWidth = 0;
+		int hBarHeight = 0;
+
+		if ((getScrollbarsMode() & SWT.SCROLLBAR_OVERLAY) != 0) {
+			org.eclipse.swt.graphics.Rectangle trim = computeTrim(0, 0, 0, 0);
+			vBarWidth = trim.width;
+			hBarHeight = trim.height;
+		}
+
 		result = ScrollPaneSolver.solve(new Rectangle(getBounds()).setLocation(0, 0), getViewport(),
-				getHorizontalScrollBarVisibility(), getVerticalScrollBarVisibility(), computeTrim(0, 0, 0, 0).width,
-				computeTrim(0, 0, 0, 0).height);
+				getHorizontalScrollBarVisibility(), getVerticalScrollBarVisibility(), vBarWidth, hBarHeight);
 		getLightweightSystem().setIgnoreResize(true);
 		try {
 			if (getHorizontalBar().getVisible() != result.showH) {


ptziegler:
After trying it out, it seems like the bounds out the root figure also needs to be set depending on the value of `GTK_OVERLAY_SCROLLING`.

Meaning one probably has to use 
```
Rectangle r = result.viewportArea;
r.setLocation(0, 0);
getLightweightSystem().getRootFigure().setBounds(r);
```

instead of 
```
Rectangle r = new Rectangle(getClientArea());
r.setLocation(0, 0);
getLightweightSystem().getRootFigure().setBounds(r);
```

a few lines further down.

azoitl:
I just tested your PR with 4diac IDE. I didn't see any observable improvement on memory consumption or performance. But as said in the last months I have a hard time to trace this.

On the plus side: I also didn't observe any negative impact. Drawing and scrolling where not affected.

wiresketch:
> Rectangle r = result.viewportArea;
> r.setLocation(0, 0);
> getLightweightSystem().getRootFigure().setBounds(r);
> ```
> 
> instead of
> 
> ```
> Rectangle r = new Rectangle(getClientArea());
> r.setLocation(0, 0);
> getLightweightSystem().getRootFigure().setBounds(r);
> ```

This makes no difference from my understanding. The calculated ScrollPaneSolver.Result.viewportArea must always be the same as the native client area as returned by the Scrollable.getClientArea(). The bug manifests itself when this is not the case. As long as the ScrollPaneSolver.solve gives the correct result, things work correctly regardless of whether overlay scrollbars are being used or not.

wiresketch:
> I just tested your PR with 4diac IDE. I didn't see any observable improvement on memory consumption or performance. But as said in the last months I have a hard time to trace this.
> 
> On the plus side: I also didn't observe any negative impact. Drawing and scrolling where not affected.

To test this properly you need to ensure that scrollbars are visible. So you need to have some content that is larger than the viewport, by either having a large "diagram" or zooming in. Once scrollbars are visible try selecting unselecting objects to trigger the FigureCanvas.layoutViewport method.

On good way to see the result is to add a log to the console something to see when figures are repainted. Something like:
```
	public final void paint(Graphics graphics) {
		System.out.println("" + System.currentTimeMillis() + " Paint " + getClass().getSimpleName() + " "
				+ Objects.hashCode(this));
		[...]
	}
```

With the bug present you'll see all the figures visible in the viewport being repainted on each select/unselect action. With the bug fixed you should only see the selected/unselected figure being repainted.

wiresketch:
@azoitl Now that I am trying to log figure paint events myself I see another problem (unrelated to this fix) which still causes a lot of repainting being done. Now this is caused by the SWT.Paint listener added on FigureCanvas in LighweightSystem.addListeners() method. This Paint listener is triggered by the GTK multiple times (by as little as a mouse move) as overlay scrollbars are being animated in response to mouse moves. This means that the content content under the animated scrollbars needs to be repainted and the paint event is triggered. On its turn it causes DeferredUpdateManager.paint(GC) method to be called and the entire viewport is then repainted. 

This means that this fix is good but it's not enough to reduce the number of paint events.

If you really want to compare to performance if what it should be like and what it is with overlay scrollbars, your should try and disabled overlay scrollbars using GTK_OVERLAY_SCROLLING=0 environment variable.

ptziegler:
> This makes no difference from my understanding. The calculated ScrollPaneSolver.Result.viewportArea must always be the same as the native client area as returned by the Scrollable.getClientArea() (after that scrollbars visibility is updated in the code just above). 

That makes sense. I think I got confused trying to test both cases without changing the environment variable.
However, now I'm in this weird situation where the overlay scrollbars are disabled, even though `GTK_OVERLAY_SCROLLING` is set to true... 😱 

image

azoitl:
> > I just tested your PR with 4diac IDE. I didn't see any observable improvement on memory consumption or performance. But as said in the last months I have a hard time to trace this.
> > On the plus side: I also didn't observe any negative impact. Drawing and scrolling where not affected.
> 
> To test this properly you need to ensure that scrollbars are visible. So you need to have some content that is larger than the viewport, by either having a large "diagram" or zooming in. Once scrollbars are visible try selecting unselecting objects to trigger the FigureCanvas.layoutViewport method.

@wiresketch rest assured that when I test I test with something where I can notice any changes. I have diagrams with several thousand figures. Moving the mouse once accross the canvas generates nearly 1GB of garbage. So I see the difference just by moveing the mouse cursor and looking at the heap state. I exaggerate the behavior by having also a thumbnail outline which is triggered on each repaint (the default behavior).

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
```diff
@@ -300,9 +303,17 @@ private void unhookViewport() {
 
 	private void layoutViewport() {
 		ScrollPaneSolver.Result result;
+		int vBarWidth = 0;
+		int hBarHeight = 0;
+
+		if (!GTK_OVERLAY_SCROLLING_ENABLED) {
```
ptziegler, line 309: Now that Scrollable.getScrollbarsMode() respects this environment variable, it sounds like it's no longer necessary to check for it in Draw2d.

This has the added benefit that it'll likely also work if overlay scrolling is disabled via e.g. the [gtk-overlay-scrolling](https://docs.gtk.org/gtk3/property.Settings.gtk-overlay-scrolling.html) setting, rather than an environment variable.

ptziegler (changes requested):

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
wiresketch: I've changed the PR. Not sure if we should add a "gtk".equals(SWT.getPlatform()) test since this is the platform the bug was encountered on.

ptziegler: > Not sure if we should add a "gtk".equals(SWT.getPlatform()) test since this is the platform the bug was encountered on.

That's probably not necessary. Even if other platforms were to support overlay scrolling, my expectation is that the width and height of the scrollbar should then also be zero.

ptziegler (approved):

ptziegler:
@azoitl Do you have any open topics?

azoitl:
> @azoitl Do you have any open topics?

No looks good to me.

azoitl (approved):

org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
```diff
@@ -300,9 +300,17 @@ private void unhookViewport() {
 
 	private void layoutViewport() {
 		ScrollPaneSolver.Result result;
+		int vBarWidth = 0;
+		int hBarHeight = 0;
+
+		if ((getScrollbarsMode() & SWT.SCROLLBAR_OVERLAY) != 0) {
```
ptziegler, line 306: We got this check wrong... `computeTrim(...)` should be called when scrollbar overlay is disabled (i.e. == 0), not the other way around.

wiresketch: Oups. You are right. What's the good way to fix this? Do I need to create another pull request?

ptziegler: I've already created https://github.com/eclipse/gef-classic/pull/429

PR #429, opened 2024-04-17
[GTK] Add test case for GTK overlay scrolling in FigureCanvas

ptziegler:


ptziegler:
Hm, the test works locally, so I assume it's the xvnc that's messing around with the bounds of the shell :/

ptziegler:
> I haven't seen this so far and not sure what the implication is. Especially as it is an I-build do we need to regularly update it?

I initially only added the 2024-06 update site. But at that point, the job would build against the M1 version of SWT, which doesn't yet contain the `setScrollbarsMode()` method and would therefore lead to compilation errors.
This would've only been necessary up until the next milestone.

The M2 was contributed yesterday, so perhaps it makes more sense to switch to the staging repository? https://download.eclipse.org/staging/2024-06

ptziegler:
Alternatively, I can move the changes to FigureCanvas to a separate PR. The new method is only needed for the tests.

azoitl:
> > I haven't seen this so far and not sure what the implication is. Especially as it is an I-build do we need to regularly update it?
> 
> I initially only added the 2024-06 update site. But at that point, the job would build against the M1 version of SWT, which doesn't yet contain the `setScrollbarsMode()` method and would therefore lead to compilation errors. This would've only been necessary up until the next milestone.
> 
> The M2 was contributed yesterday, so perhaps it makes more sense to switch to the staging repository? https://download.eclipse.org/staging/2024-06

Staging repository sounds like a good idea for me.

ptziegler:
The staging repository seems to have worked.

azoitl:
Great. This has also the advantage that we are nearer to any platform changes.

PR #431, opened 2024-04-19
[Modernize Code] Cleaned all GEF Tools

azoitl:
Addresses: #155

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
@@ -15,7 +15,6 @@ package org.eclipse.gef.tools;
 import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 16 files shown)


org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
```diff
@@ -321,13 +326,9 @@ protected void applyProperty(Object key, Object value) {
 			}
 			if (property != null) {
 				Method setter = property.getWriteMethod();
-				// setter.setAccessible(true);
 				setter.invoke(this, value);
 			}
-		} catch (IntrospectionException ie) {
-		} catch (IllegalAccessException iae) {
```
ptziegler, line 328: I usually like to catch a ReflectiveOperationException, rather than an IllegalAccessException, InvocationTargetException and SecurityException. What do you think?

ptziegler:
@azoitl Can you have a look at the conflict?

org.eclipse.gef/src/org/eclipse/gef/tools/AbstractTool.java
azoitl: I'm normally not using reflection to much. Therefore I didn't check the exceptions. But yes your proposal is much cleaner. Changed it.

azoitl:
> @azoitl Can you have a look at the conflict?

Sure. Done!

ptziegler (approved):

Issue #430, opened 2024-04-18
Huge number of paint events triggered by overlay scrollbars on Linux/GTK3


wiresketch:
I'd like to file another issue that I've observed while working on #425. Trying to debug performance issues in my application on Linux/GTK3 I am noticing another problem with overlay scrollbars that negatively impacts performance. I'd like to open this issue to have a discussion for possible solutions and workarounds.

What happens is that when overlay scrollbars are visible for a control, the GTK system will show/hide them automatically on user activity, specifically on mouse moves. For example if the mouse stops moving over the scrolled area, scrollbars will fade away. If the mouse starts moving again - scrollbars will fade in. This animation for scrollbars triggers hundreds of SWT.Paint events that are then handled by LightweightSytem/DeferredUpdateManager.

At first I thought it would be logical for the content to be redrawn under the scrollbars, as long as the paint event is sent with the right clipping that only includes the region under the scrollbar. However I was surprised to find out that this is not the case. The entire client area is being forced to be repainted.

Since this behavior is actually the native behavior, it can be observed with a simple SWT snippet. I've attached it here. 
[ScrollbarsTest.java.txt](https://github.com/eclipse/gef-classic/files/15022787/ScrollbarsTest.java.txt)

Try moving the mouse and observe the number of logged paint events.

The same example can be launched with GTK_OVERLAY_SCROLLING=0 to disable overlay scrollbars, in which case a very limited number of paint events can be observed, and none while moving the mouse.

Since it looks like this is the native behavior on Linux/GTK3 not specific to Draw2d/GEF I would like to explore here possible ways to solve this performance issue.

One obvious option that I see is to force the Draw2d/GEF applications to be always used with overlay scrollbars disabled to achieve  optimum performance on Linux. This means using GTK_OVERLAY_SCROLLING=0 variable on GTK3 when launching the application. However this has the side effect of disabling overlay scrollbars for all views in the application, not only for FigureCanvas control. Another problem is that this is not future proof as this environment variable is no longer being used on GTK4.

[...]

PR #432, opened 2024-04-20
Add static method to configure overlay scrolling for FigureCanvas

ptziegler:
This allows the user to specify whether new instances of FigureCanvas should use overlay scrolling on not. This is disabled by default, due to performance issues on Linux/GTK3.

As a result, the scrollbar is shown as a separate entity and no longer part of the client area. This setting only affects operating systems supporting overlay scrolling. This method requires SWT 3.126 or newer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# GEF Classic 3.20.0 (Eclipse 2024-06)
+
+## Draw2d
+ - Clients can disable overlay scrolling by calling FigureCanvas.setScrollbarsMode()
+   prior to creating the object.
+
 # GEF Classic 3.19.0 (Eclipse 2024-03)
 
 ## Draw2d
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
@@ -13,6 +13,9 @@
 package org.eclipse.draw2d;
 
 import java.beans.PropertyChangeListener;
+import java.lang.reflect.Method;
+import java.util.logging.Level;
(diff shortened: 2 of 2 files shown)


azoitl:
I'm not sure if we should change the default behavior here. While I think that overlay scrollbars are very bad usability I still think we should by default have a consistent look and feel. 

Also, as a long term >10 years user of GEF Classic on Linux I only note the performance issue if this on very large diagrams (> 10k Figures, maybe even bigger). 

Therefore I think GEF Classic users where this is really a problem should be able to deactivate it but not the other way round.

ptziegler:
> I still think we should by default have a consistent look and feel

I see your point. The underlying idea was to disable overlay scrolling by default, to have the look and feel be the same across all platforms (as overlay scrolling isn't supported by e.g. Windows).

I've added `null` as the new default value, which corresponds to the default setting of the operating system. So unless the user explicitly calls `FigureCanvas.setScrollbarsMode()`, the behavior isn't changed.

azoitl:
> > I still think we should by default have a consistent look and feel
> 
> I see your point. The underlying idea was to disable overlay scrolling by default, to have the look and feel be the same across all platforms (as overlay scrolling isn't supported by e.g. Windows).

I fear this is the price we pay with Eclipse SWT trying to look as native as possible on the different platforms. 

I like this feature as it gives me a chance to debug the bigger performance issues that my Windows users have now also under GTK/Linux

ptziegler:
I now wonder whether this static method is really necessary. Now that SWT provides a `setScrollbarsMode()`, clients can control when overlay scrollbars should be used.

ptziegler:
If clients want to disable overlay-scrolling, they can do so by accessing the Canvas widget directly.

PR #434, opened 2024-04-25
Add flag to enable double-buffering when doing paint operations

ptziegler:
By using the `org.eclipse.draw2d.LightweightSystem.useDoubleBuffering`, clients can enable a special functionality, that first paints all UI updates to an internal buffer, before it is painted to the screen.

The hope of behavior is to increase the perceived performance of an application by detaching rendering from the display.


ptziegler:
The UpdateManager already does double buffering via the given GraphicsSource. So this PR is redundant.

PR #435, opened 2024-04-26
Make SeparatorBorder public API and move it from GEF to Draw2D

ptziegler:
Use Case: Border between figures directly adjacent to one another. See the UMLClass example, for reference.

For my current task, I've converted the property table in the bottom left corner to a proper GEF viewer. Each column in this table corresponds to an edit part, which is each backed by a simple figure.

I currently need to use a custom border, to make sure that a line is only drawn at the bottom of each figure. I think this could be a common use case and given that a similar class already exists (despite not being used), I would like to make it public.

image


azoitl:
LGTM. I'm only curios why the internal seperator border was never used and why you are implementing the properties view as GEF viewer. We made very nice experience implementing it as tabbed properties with the Forms infrastructure.

ptziegler:
> I'm only curios why the internal seperator border was never used and why you are implementing the properties view as GEF viewer. 

A couple of reasons. The biggest one being that this isn't a new implementation, but rather a refactoring of the current code. Right now, the table is painted directly on a Canvas. While this isn't a problem in itself, it makes automatic testing next to impossible.

By using GEF, I'm able to reuse most of the existing implementation (effectively painting on a FigureCanvas, rather than a Canvas), which means that I only really need to modify a single class, as the backend doesn't change significantly.

> We made very nice experience implementing it as tabbed properties with the Forms infrastructure.

That probably would've made sense when starting from scratch, but I'm afraid it would take a significant amount of time to migrate all the functionality to work with Forms and at the same time also preserve the overall look and feel. Sadly, I just don't have the amount of time available to make such an extensive change... 😕

azoitl:
> 
> > We made very nice experience implementing it as tabbed properties with the Forms infrastructure.
> 
> That probably would've made sense when starting from scratch, but I'm afraid it would take a significant amount of time to migrate all the functionality to work with Forms and at the same time also preserve the overall look and feel. Sadly, I just don't have the amount of time available to make such an extensive change... 😕

I can relate a  lot to that. In the other hand the screenshot looks pretty nice. So definitely not a bad decision.

Issue #436, opened 2024-05-02
"Error: SWT Resource was not properly disposed" in ScaledGraphics.zoomTextLayout()


cedric780:
Latest Eclipse RCP introduced a watchdog that reports all no disposed SWT resource that get garbage collected.
Since them I get a lot of traces in the console with this stack trace:

```
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:172)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:120)
	at org.eclipse.swt.graphics.TextLayout.<init>(TextLayout.java:299)
	at org.eclipse.draw2d.ScaledGraphics.zoomTextLayout(ScaledGraphics.java:980)
	at org.eclipse.draw2d.ScaledGraphics.drawTextLayout(ScaledGraphics.java:414)
	at org.eclipse.draw2d.Graphics.drawTextLayout(Graphics.java:457)
	...
```

I use draw2d 3.10.xxx but it seems the bug is still present in latest version, I think the problem originates here:

https://github.com/eclipse/gef-classic/blob/1dd4d1dccd667a36aaa35c61a62dc49dda28db6d/org.eclipse.draw2d/src/org/eclipse/draw2d/ScaledGraphics.java#L999

The `zoomed` TextLayout should be disposed when returning null.

PR #437, opened 2024-05-02
Fix [#436] "Error: SWT Resource was not properly disposed" in ScaledG…

cedric780:
[#436] Fix "Error: SWT Resource was not properly disposed" in ScaledGraphics.zoomTextLayout()

Resolves eclipse/gef-classic#436
Dispose the TextLayout when returning null.

Signed-off-by: Cedric Marin <cedric.dev.780@gmail.com>


ptziegler:
Thank you for the fix!

PR #438, opened 2024-05-06
Extract Constraint handling into AbstractConstraintLayout

azoitl:
XYLayout and RulerLayout share quite some common code. This is extracted in a common base class to be used also by other Layouts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
 # GEF Classic 3.20.0 
 
+## Draw2d
+ - Introduction of new base class `org.eclipse.draw2d.AbstractConstraintLayout` that provides a generic constraint map management for 
+   layout implementations. 
+   Currently used by `XYLayout`,  `DelegatingLayout`, and in GEF `RulerLayout`.   
+
 ## GEF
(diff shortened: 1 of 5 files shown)


ptziegler:
Shouldn't the DelegatingLayout and ScrollBarLayout also extend the new AbstractConstraintLayout? For the BorderLayout and GridLayout it's probably not as easy, as it would interfere with the AbstractHintLayout.

image

ptziegler:
Could you also add a small entry to the changelog? I fear that this addition might get lost, otherwise.

azoitl:
> Shouldn't the DelegatingLayout and ScrollBarLayout also extend the new AbstractConstraintLayout? For the BorderLayout and GridLayout it's probably not as easy, as it would interfere with the AbstractHintLayout.
> 

I totally missed DelegatingLayout. ScrollBarLayout and BorderLayout use the constraints differently. So it makes no sense. GridLayout could benefit. But I also don't know how to do that. 

> Could you also add a small entry to the changelog? I fear that this addition might get lost, otherwise.

Yes will do. Good point

ptziegler (approved):

PR #439, opened 2024-05-06
Implement lightweight logging facade for Draw2D

ptziegler:
Draw2D only having a dependency to SWT greatly limits us in the way we can track error messages. This change implements a generic logging interface, in order to not restrict ourselves to one specific framework.

This interface is implemented by GEF, using the Eclipse ILog as back-end. If no custom implementation is provided, the native Java logging framework is used as fallback.

By doing so, we are now able to, we are able to log errors in Draw2D in a way that is consistent with other Eclipse projects.

The API for this facade is heavily influenced by the Java logger, in order to keep the mechanism simple and familiar.

This facade is for internal use only.


ptziegler:
This is a prerequisite for replacing the exceptions thrown by the layout managers with simple log messages. 

Here an example of how it looks like in the IDE:
image

azoitl (approved):
I like this a lot. I was always missing a logging infrastructure in draw2d.

PR #440, opened 2024-05-06
XYLayout and RulerLayout shouldn't throw exception on invalid constraint

ptziegler:
Depends on #439

### DIFF
--- a/org.eclipse.gef.tests/src/org/eclipse/gef/test/RulerLayoutTests.java
+++ b/org.eclipse.gef.tests/src/org/eclipse/gef/test/RulerLayoutTests.java
@@ -15,6 +15,15 @@ package org.eclipse.gef.test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.runtime.ILog;
+import org.eclipse.core.runtime.ILogListener;
+import org.eclipse.core.runtime.Platform;
 
 import org.eclipse.draw2d.Figure;
 import org.eclipse.draw2d.geometry.Rectangle;
@@ -23,6 +32,7 @@ import org.eclipse.gef.internal.ui.rulers.RulerFigure;
 import org.eclipse.gef.internal.ui.rulers.RulerLayout;
 import org.eclipse.gef.rulers.RulerProvider;
 
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -33,6 +43,9 @@ public class RulerLayoutTests {
 	private Figure dummy;
 	private Figure container;
 	private int constraint;
+	protected ILog log;
+	protected ILogListener logListener;
+	protected List<String> logMessages;
 
 	@Before
 	public void setUp() {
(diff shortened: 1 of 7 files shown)


azoitl (approved):

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerLayout.java
```diff
@@ -80,8 +86,12 @@ public void layout(IFigure container) {
 	@Override
 	public void setConstraint(IFigure figure, Object newConstraint) {
 		if (newConstraint != null && !(newConstraint instanceof Integer)) {
-			throw new IllegalArgumentException("RulerLayout was given " + newConstraint.getClass().getName() //$NON-NLS-1$
-					+ " as constraint for Figure. Integer expected!"); //$NON-NLS-1$
+			if (!(newConstraint instanceof GridData)) {
```
azoitl: I think this is a copy paste error.

ptziegler:
One second, I think there's a copy-paste error in the RulerLayout...

org.eclipse.gef.tests/src/org/eclipse/gef/test/RulerLayoutTests.java
```diff
@@ -57,9 +77,12 @@ public void testConstraint() {
 	/**
 	 * Use Case: Constraint must be of type int -> error
 	 */
-	@Test(expected = IllegalArgumentException.class)
+	@Test
 	public void testInvalidConstraint() {
```
azoitl, line 81: Do we now need a test where we are giving an Integer and check that no message was logged?

azoitl (changes requested):

azoitl:
> One second, I think there's a copy-paste error in the RulerLayout...

I also just noticed.

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerLayout.java
```diff
@@ -80,8 +85,12 @@ public void layout(IFigure container) {
 	@Override
 	public void setConstraint(IFigure figure, Object newConstraint) {
 		if (newConstraint != null && !(newConstraint instanceof Integer)) {
-			throw new IllegalArgumentException("RulerLayout was given " + newConstraint.getClass().getName() //$NON-NLS-1$
-					+ " as constraint for Figure. Integer expected!"); //$NON-NLS-1$
+			if (!(newConstraint instanceof Integer)) {
```
azoitl: I think the if is not needed anymore. we have all the checks in line 87

azoitl (changes requested):

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerLayout.java
ptziegler: That was the copy-paste error I mentioned 😉

org.eclipse.gef.tests/src/org/eclipse/gef/test/RulerLayoutTests.java
ptziegler: I suppose that's easy enough to do... Done.

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerLayout.java
ptziegler: In any case, that should also be fixed now.

azoitl (approved):

ptziegler:
I've also added a null check to the GridLayout class. Otherwise the logger would potentially cause a NPE.

azoitl:
> > I've also added a null check to the GridLayout class. Otherwise the logger would potentially cause a NPE.
> 
> I really shouldn't rush this stuff... Argh.

I know how you feel. Been there done that.

ptziegler:
Alright... I've gone over everything again and I don't see anything off. If I missed something, then so be it.

azoitl (approved):

Issue #443, opened 2024-05-16
Some of the Zest examples are broken


ptziegler:
**GraphJFaceSnippet3:**

![image](https://github.com/eclipse/gef-classic/assets/70652978/f39f570e-1875-4e03-be25-2885b4cf5ba7)

I would expect a proper layout, instead of all nodes being on top of one another.

**GraphJFaceSnippet5:**

![image](https://github.com/eclipse/gef-classic/assets/70652978/91db4413-68d9-4a0f-83a7-224792d9c8ca)

Running the same example multiple times results in different node sizes. Sometimes the nodes are too small.

**GraphJFaceSnippet6:**
![image](https://github.com/eclipse/gef-classic/assets/70652978/810a207e-7922-4c89-8362-bb2fd1cb5821)

The nodes inside the nested graph have an empty size.

**GraphJFaceSnippet8:**

![image](https://github.com/eclipse/gef-classic/assets/70652978/99670782-2a0b-4675-9dd3-164196dd5b75)

Similar issue. In addition, the label provider is broken.

There are probably other examples affected as well, but I haven't checked them all, yet.

ptziegler:
Fixed with https://github.com/eclipse/gef-classic/commit/4cd9bc048f373d00eecaa0ae640e7faf56d8c89b, https://github.com/eclipse/gef-classic/commit/2288625a8c910773a4e3076691b39ad3c13a2cc8, https://github.com/eclipse/gef-classic/commit/f29e90d9d8f23657009b20ce7276303c95cc3490 and https://github.com/eclipse/gef-classic/commit/89280416c74a276d9bd6f563c20931e81604cc3c

PR #444, opened 2024-05-16
Fix: Graph nodes using the tree layout have an empty size.

ptziegler:
Can be reproduced using the GraphJFaceSnippet6. The problem occurs because a size() == 0 check has been wrongfully replaced with a !isEmpty().

See https://github.com/eclipse/gef-classic/issues/443 Amends 2e860acc0b2e1e8847408141e00f787ee3aee9c1


azoitl (approved):
Stupid me and copy paste

PR #446, opened 2024-05-18
Integrate Zest-based examples into our test suite

ptziegler:
With this change, we instantiate and verify the correctness of the Zest examples as part of our JUnit test suite. Special care needs to be taken, in order to make sure that the examples can be used in the test environment, but also still be executed as standalone applications.

This means that those snippets must not create a new Display instance, as one is already created by the test suite and multiple displays are forbidden. The snippet is then instantiated by reflectively calling the main method.

Because the snippets block until the shell is closed by the user, a little trick is needed to allow for the tests to be executed. The actual test is scheduled as an asynchronous task before the snippet is created, which is then executed by the call to "Display.readAndDispatch()".

This contribution is done in preparation for the Zest2.0 migration as a means to increase the test coverage and thus spot any changes that are introduced.

### DIFF
--- a/org.eclipse.zest.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.zest.core/META-INF/MANIFEST.MF
@@ -12,6 +12,6 @@ Bundle-ActivationPolicy: lazy
 Export-Package: org.eclipse.zest.core.viewers,
  org.eclipse.zest.core.viewers.internal;x-internal:=true,
  org.eclipse.zest.core.widgets,
- org.eclipse.zest.core.widgets.internal;x-internal:=true
+ org.eclipse.zest.core.widgets.internal;x-friends:="org.eclipse.zest.tests"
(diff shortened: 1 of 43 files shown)


ptziegler:
This is the code coverage achieved with these additional tests:
image

Compared to our current coverage:
image

ptziegler:
I think it's a good idea to also convert the examples for Draw2D and GEF into integration tests. Though for the time being, I really am fed up with writing tests...

azoitl (approved):

azoitl:
This is a great piece of work. Way beyond my Java knowledge.

ptziegler:
I tried to move as much complexity to the TestRule as possible, so I hope the tests are somewhat understandable. In the end, the focus was more on having all examples be run once, rather than testing any complex user input.

azoitl:
Yes you did a great job here. Hope to find time to learn from it.

PR #447, opened 2024-05-19
Fix: Adding a node to a graph container adds it to the graph instead

ptziegler:
When the constructors in the GraphNode class have been unified, we wrongfully pass "graphModel.getGraph()" as an argument instead of the current container. This causes all nodes to be added to the parent graph.

See the UML example for reference. One of the class diagrams should be contained by the UML container, but all three are part of the parent graph instead.

Amends a13541fd21b23c8fdfb5c64a99321f7b8be44706

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphNode.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphNode.java
@@ -101,7 +101,7 @@ public class GraphNode extends GraphItem {
 	}
 
 	public GraphNode(IContainer graphModel, int style, Object data) {
-		this(graphModel.getGraph(), style, "" /* text */, null /* image */, data);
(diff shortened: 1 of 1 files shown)


ptziegler:
In the UML example, we have the following code segment, which should create a new UML node and add it to the graph container.

```
Graph g = new Graph(shell, SWT.NONE);
g.setConnectionStyle(ZestStyles.CONNECTIONS_DIRECTED);
GraphContainer c = new GraphContainer(g, SWT.NONE);
c.setText("A UML Container");
UMLNode n = new UMLNode(c, SWT.NONE, createClassFigure1(classFont, classImage, publicField, privateField));
```

However, this currently looks like this:

image

And with this fix, it looks like this:

image

ptziegler:
https://github.com/eclipse/gef-classic/issues/443

azoitl (approved):

PR #451, opened 2024-05-24
Fix NullPointerException when closing GraphSnippet13

ptziegler:
This example has a special situation where the connection figure is the same object as the source and target container connection figure. Because the connection figure is disposed first, the parent of those other objects is already null, leading to an exception.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphConnection.java
@@ -1,5 +1,6 @@
 /*******************************************************************************
- * Copyright 2005, 2023, CHISEL Group, University of Victoria, Victoria, BC, Canada.
+ * Copyright 2005, 2024, CHISEL Group, University of Victoria, Victoria, BC,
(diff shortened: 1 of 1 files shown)


azoitl (approved):

ptziegler:
Now that I look at it again... I wonder if it would've been better to do

```
if (sourceContainerConnectionFigure != null) {
	if (sourceContainerConnectionFigure.getParent() != null) {
		sourceContainerConnectionFigure.getParent().remove(sourceContainerConnectionFigure);
	}
	sourceContainerConnectionFigure = null;
}
```

instead of 
```
if (sourceContainerConnectionFigure != null && sourceContainerConnectionFigure.getParent() != null) {
	sourceContainerConnectionFigure.getParent().remove(sourceContainerConnectionFigure);
	sourceContainerConnectionFigure = null;
}
```

azoitl:
oh yes missed that. If you do that I think it is also cleaner to have a parent local variable.

ptziegler:
Sure, I'll fix that with a new PR.

ptziegler:
https://github.com/eclipse/gef-classic/pull/452

PR #453, opened 2024-06-05
add auto sync action

pazi146:
syncs master branch

### DIFF
--- a/.github/workflows/autosync.yml
+++ b/.github/workflows/autosync.yml
@@ -1,0 +1,16 @@
+name: Auto-sync
+run-name: Auto-sync of upstream branches
+on:
+  workflow_dispatch:
+  schedule: 
+    - cron: "35 4-22 * * *"
+
+jobs:
+  auto-sync:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+    steps:
+      - name: Sync release
+        run: |
+          gh repo sync ${{github.repository}} --branch master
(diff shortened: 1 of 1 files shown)


azoitl (approved):

azoitl:
@ptziegler we have the same workflow for some time now in Eclipse 4diac and it is very convenient as it is automatically syncing your fork with the main repo. @pazi146 found this solution therefore I asked him to also contribute it to GEF classic. What do you think?

@merks I think this workflow could also be something for simrel. However it only works after people synced there fork once manually.

ptziegler:
@azoitl I'm fairly neutral towards this as I rarely -if ever- have to synchronize the master branch of the fork. In my usual workflow, I only ever have to rebase the development branches and there I simply use the master branch of this repo.

That said, the only feedback I have for this PR is to squash both commits into one.

merks:
Yes I never need to sync my fork. I clone the original repo and create a remote for my fork. I only push topic branches to my fork and use it to create a PR.  The main/master branch of my fork is just never used. 

https://github.com/orgs/eclipse-simrel/discussions/3

Of course other people do things differently and it’s really annoying to find out later that your fork is out of sync.

azoitl:
@ptziegler @merks thanks for the feedback. I'm always interested in improving my Github workflow. I somehow have my fork as default remote. The main reason is that I always fear to accidentally directly push to this repo. However I think your approach seems to be with less overhead and steps. Any tips to address my fear?

merks:
One thing that I do that prevents accidental push is that I clone with an https and I don't have credentials that support https, I only have ssh keys so direct push fails.

image

It's probably also possible to delete the push from the remote...

azoitl:
I tried that workflow now in all my github projects. Thx again this makes it much easier. Also I noticed that the workflow from @pazi146 sometimes has issues. Therefore I don't think anymore we should have it. Still thx @pazi146 for finding and providing it.

PR #457, opened 2024-06-13
Update feature and plugin version numbers for 3.21 release cycle

ptziegler:
@azoitl 

FYI, I looked into this obscure "The minor version should be incremented in version x.y.z, since execution environments have been changed since version x.y.z" error and I'm fairly confident that this is an issue within Equinox.

In short, the API tool is unable to load the execution environment of the baseline plugins because Equinox ignores the "Require-Capability" header (at least for loading the EE). And therefore, an empty string is compared to the execution environment that is set in the workspace (JavaSE-17), which is therefore marked as an error.

### DIFF
--- a/org.eclipse.draw2d-feature/feature.xml
+++ b/org.eclipse.draw2d-feature/feature.xml
@@ -14,7 +14,7 @@
 <feature
       id="org.eclipse.draw2d"
       label="%featureName"
-      version="3.20.0.qualifier"
+      version="3.21.0.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.draw2d.sdk-feature/feature.xml
+++ b/org.eclipse.draw2d.sdk-feature/feature.xml
@@ -14,7 +14,7 @@
 <feature
       id="org.eclipse.draw2d.sdk"
       label="%featureName"
-      version="3.20.0.qualifier"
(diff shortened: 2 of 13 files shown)


merks:
That’s very annoying!

azoitl:
I agree with @merks. 

My only question is how should we treat the other plugins (tests, examples,...)?

ptziegler:
> My only question is how should we treat the other plugins (tests, examples,...)?

I believe the only reason they don't produce errors is because the tool is not enabled for them. We could enable a "reduced" validation, where we exclude API breakage.

azoitl:
> > My only question is how should we treat the other plugins (tests, examples,...)?
> 
> I believe the only reason they don't produce errors is because the tool is not enabled for them. We could enable a "reduced" validation, where we exclude API breakage.

What validation would we get then?

ptziegler:
> What validation would we get then?

Error when annotations are missing, if the version hasn't been updated when e.g. a test was added, etc. The kind of stuff one usually forgets :')

ptziegler:
Or to be more specific, everything outside of "API Compatibility"

image

azoitl:
> > What validation would we get then?
> 
> Errors when annotations are missing, if the version hasn't been updated when e.g. a test was added, etc. The kind of stuff one usually forgets :')

Ah yes, as chaotic as I am, this is definitely something I would need.

azoitl (approved):

Issue #456, opened 2024-06-13
ScrollPaneSolver doesn't respect scrollbar when using setScrollbarsMode(SWT.NONE)


ptziegler:
This is something I just encountered while investigating https://github.com/eclipse-windowbuilder/windowbuilder/issues/795.

In short, I have a table where I want to disable overlay scrolling by calling `setScrollbarsMode(SWT.NONE)`. However, this flag is ignored when calculating the initial layout. As a result, a part of the table is obscured by the scrollbar:

![image](https://github.com/eclipse/gef-classic/assets/70652978/e5599efd-b8c5-40f2-a423-3eb3af5f83f1)

PR #459, opened 2024-06-19
Use rectangle from ScrollPaneSolver as client area for root figure

ptziegler:
The value returned by getClientArea() might not match the physical size of the canvas. Therefore it might happen that parts of the canvas are hidden behind a scrollbar or not painted at all.

Resolves https://github.com/eclipse/gef-classic/issues/456

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
@@ -319,7 +319,7 @@ public class FigureCanvas extends Canvas {
 			if (getVerticalBar().getVisible() != result.showV) {
 				getVerticalBar().setVisible(result.showV);
 			}
-			Rectangle r = new Rectangle(getClientArea());
+			Rectangle r = new Rectangle(result.viewportArea);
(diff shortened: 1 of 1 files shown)


ptziegler:
For illustration, this is how the problem currently manifests itself. The viewer of interest is the property table, in the bottom left corner of the editor.

This is the current behavior:
[before.webm](https://github.com/eclipse/gef-classic/assets/70652978/1eb28b04-0bfa-497a-a411-860f6e6aceaf)

This is what it looks like when this PR is applied:
[after.webm](https://github.com/eclipse/gef-classic/assets/70652978/30ba5518-ff69-4b83-b515-728d678e9544)

azoitl:
Tested in  addition with 4diac IDE. Didn't see any impact on our "special" editors.

azoitl (approved):

ptziegler:
Maybe for context: This line was introduced with 793c107460b6e81147280e673f02f5f77b674645. But I don't really see how this is related to the commit message, so I can't make any assumptions about the initial intention.

azoitl:
With all the history and changes in SWT it could likely be that some assumptions or boundary conditions that where valid back in 2006 are not here anymore.

PR #468, opened 2024-07-06
Added 3.20 Baseline Target Platform

azoitl:


ptziegler:
The version bumps are due to the new Tycho version, I assume? I vaguely remember Tycho switching to a newer version of the Eclipse compiler every now and then, which may result in different byte code, given the same source file.

azoitl:
> The version bumps are due to the new Tycho version, I assume? I vaguely remember Tycho switching to a newer version of the Eclipse compiler every now and then, which may result in different byte code, given the same source file.

I'm not sure. What I did is that in the main pom I changed the base-line version to check against to GEF Classic 3.20. And then the check we have demanded that at least the minor version is upgraded.

Issue #469, opened 2024-07-07
SWTException: Widget is disposed on palette tooltips after window is detached


Phillipus:
1. Have a GEF Editor with a Palette
2. Right-click on the Editor's tab and select "Detach" to detach the Editor Part
3. Hover the mouse cursor over a tool in the palette to see the tooltip
4. Re-attach the editor part to the main workbench by dragging and dropping it into the workbench editor area
5. Hover the mouse cursor over the same tool that you hovered over before

```
!ENTRY org.eclipse.ui 4 0 2024-07-07 15:06:51.253
!MESSAGE Unhandled event loop exception
!STACK 0
org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:853)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:632)
	at org.eclipse.swt.widgets.Widget.getStyle(Widget.java:1069)
	at org.eclipse.draw2d.LightweightSystem$RootFigure.isMirrored(LightweightSystem.java:347)
	at org.eclipse.draw2d.Figure.isMirrored(Figure.java:1104)
	at org.eclipse.draw2d.text.BlockFlow.getOrientation(BlockFlow.java:169)
	at org.eclipse.draw2d.text.BlockFlow.validate(BlockFlow.java:318)
	at org.eclipse.draw2d.text.FlowPage.validate(FlowPage.java:160)
	at org.eclipse.draw2d.text.FlowPage.getPreferredSize(FlowPage.java:91)
	at org.eclipse.gef.ui.palette.editparts.PaletteEditPart$2.getPreferredSize(PaletteEditPart.java:124)
	at org.eclipse.draw2d.StackLayout.calculatePreferredSize(StackLayout.java:75)
	at org.eclipse.draw2d.AbstractLayout.getPreferredSize(AbstractLayout.java:113)
	at org.eclipse.draw2d.AbstractHintLayout.getPreferredSize(AbstractHintLayout.java:92)
	at org.eclipse.draw2d.Figure.getPreferredSize(Figure.java:855)
	at org.eclipse.draw2d.Figure.getPreferredSize(Figure.java:844)
[...]

PR #470, opened 2024-07-08
Fix possible SWTException "Widget is disposed" exception when displaying a Tooltip

Phillipus:
If a workbench part is detached and a tooltip is initially displayed, when the part is re-attached the parent Shell is disposed which in turn disposes of the PopupHelper's Shell. This will lead to an SWT Error when the Shell is accessed again.

To fix this, create the PopupHelper Shell with the Control's Display instead of its Shell.

See https://github.com/eclipse/gef-classic/issues/469


ptziegler (approved):
I've tried it on my end and can confirm that this solves the issue.

Issue #464, opened 2024-07-04
IllegalArgumentException: Argument cannot be null when creating SharedCursor


Phillipus:
Cursors for the Palette are created in the [SharedCursors ](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/SharedCursors.java)class. The `ImageData` is returned from the `ImageDescriptor` at the current device zoom level. But if the zoom level is not 100% or 200% then `ImageData` will be null and an exception is thrown.

1. On Windows launch a GEF-based app with a Palette with the VM argument `-Dswt.autoScale=150`
2. Select a tool from the palette

This is caused by the `ImageDescriptor` not finding an image with a `@1.5x` suffix.

Perhaps a fallback is required so that if ImageData is null 100% zoom is used instead?

Phillipus:
Or another possibility is to get the `ImageData` from an `Image` rather than an `ImageDescriptor` as that will not be null:

```java
    private static Cursor createCursor(String sourceName) {
        Image img = new Image(null, Internal.class.getResourceAsStream(sourceName));
        ImageData id = img.getImageData(getDeviceZoom());
        img.dispose();
        return new Cursor(null, id, 0, 0);
    }
```

Phillipus:
There are other cursors created in `FlyoutPaletteComposite` here -  https://github.com/eclipse/gef-classic/blob/e1d163b08c72039ce96efd2692edbda1ff06d6f9/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java#L1540

Because these are created with `ImageDescriptor.getImageData()` with no zoom level they are too small on Windows at hi-dpi 200% scale. These would need to be changed as well.

ptziegler:
I agree that throwing an exception is definitely not the way to go. My initial approach would've been to fall back to the 100% zoom level, as you suggested. Alternatively, one could also make it so that it goes to the "closest" zoom level instead. e.g. from 125% to 100% and from 175% to 200%.

```
private static Cursor createCursor(String sourceName) {
	ImageDescriptor src = ImageDescriptor.createFromFile(Internal.class, sourceName);
	ImageData data = src.getImageData(getDeviceZoom());
	if (data == null) {
		// if no image for that zoom level exists use the one for the default zoom level
		data = src.getImageData(100);
	}
	return new Cursor(null, data, 0, 0);
}
```

[...]

PR #479, opened 2024-07-16
Fix errors in drag cursors when using high-DPI screens

ptziegler:
The cursors used in the flyout palette only exist for the default 100%. If a system is using a higher zoom level, the creation of those cursors fails with an IllegalArgumentException. Instead, we should scale the cursor image to match the current level.

Instead of reading the zoom level only once, it is now calculated dynamically, as the zoom may change depending on the display or because it has been changed. The methods for handling this case have also been moved from the SharedCursors class to the InternalGEFPlugin, so that it becomes available to the DragCursors class.

From there, we call asImageDataProvider() to convert the image descriptor to image data which internally scales the image via DPIUtil magic, if necessary.

See https://github.com/eclipse/gef-classic/issues/464

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
@@ -13,6 +13,11 @@
 
 package org.eclipse.gef.internal;
 
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.ImageData;
+
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.resource.ImageRegistry;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.plugin.AbstractUIPlugin;
@@ -67,4 +72,43 @@ public class InternalGEFPlugin extends AbstractUIPlugin {
 		super.stop(context);
 	}
 
+	/**
+	 * Convenience method for getting the current zoom level of the active device.If
+	 * on MacOS or Linux (x11 window system) or if the device zoom couldn't
+	 * otherwise be determined, this method returns {@code 100} as default value.
+	 */
+	public static int getOrDefaultDeviceZoom() {
+		// On Mac and Linux X11 ImageData for cursors should always be created with 100%
+		// device zoom
+		if (Platform.getOS().equals(Platform.OS_MACOSX) || (Platform.getOS().equals(Platform.OS_LINUX)
+				&& "x11".equalsIgnoreCase(System.getenv("XDG_SESSION_TYPE")))) { //$NON-NLS-1$//$NON-NLS-2$
+			return 100;
+		}
+		String deviceZoom = System.getProperty("org.eclipse.swt.internal.deviceZoom", "100"); //$NON-NLS-1$ //$NON-NLS-2$
+		try {
+			return Integer.parseInt(deviceZoom);
+		} catch (NumberFormatException e) {
+			return 100;
+		}
+	}
+
+	/**
+	 * Convenience method to get the image data for a given zoom level. If no image
+	 * for the requested zoom level exists, the image data may be an auto-scaled
+	 * version of the native image and may look blurred or mangled.
+	 */
+	public static ImageData scaledImageData(ImageDescriptor descriptor, int zoom) {
+		// Default case: Image in matching resolution has been found
+		ImageData data = descriptor.getImageData(zoom);
+		if (data != null) {
+			return data;
+		}
+		// Otherwise artifically scale the image
+		Image image = descriptor.createImage();
+		try {
+			return image.getImageData(zoom);
+		} finally {
+			image.dispose();
+		}
+	}
 }
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2010 IBM Corporation and others.
+ * Copyright (c) 2004, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -32,6 +32,7 @@ import org.eclipse.swt.events.MouseTrackListener;
 import org.eclipse.swt.graphics.Cursor;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.widgets.Canvas;
@@ -84,6 +85,7 @@ import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.gef.GraphicalViewer;
 import org.eclipse.gef.dnd.TemplateTransfer;
 import org.eclipse.gef.internal.GEFMessages;
+import org.eclipse.gef.internal.InternalGEFPlugin;
 import org.eclipse.gef.internal.InternalImages;
 import org.eclipse.gef.internal.ui.palette.PaletteColorUtil;
 import org.eclipse.gef.ui.views.palette.PaletteView;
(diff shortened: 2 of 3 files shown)


ptziegler:
I've created https://github.com/eclipse-platform/eclipse.platform.ui/pull/2084 to contribute 2x variants for those drag icons. Everything seems to work for 100% zoom already, but I haven't had the chance to test this on a high-DPI screen.

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
```diff
@@ -58,6 +60,26 @@ public static AbstractUIPlugin getDefault() {
 		return singleton;
 	}
 
+	public static int getDeviceZoom() {
```
fedejeanne, line 63: This implementation will probably suffer from the same problem that `DPIUtil` was suffering which is: the value will be obtained once and never updated again _i.e._ if the zoom of the monitor changes then this value will be outdated.

ptziegler: How was this solved in SWT?

fedejeanne: `DPIUtil.setDeviceZoom` is called on zoom changes, which keeps the `deviceZoom` always up-to-date _i.e._ calling `DPIUtil.getDeviceZoom()` should always give you the current zoom. But those methods are not accessible in this case, so this is just fyi.

In this specific case, you could simply avoid storing the zoom in the field `deviceZoom` and query the system property `org.eclipse.swt.internal.deviceZoom` every time. Since `DPIUtil.setDeviceZoom(...)` is **also** updating that property then the result should be that querying it always gives you the right zoom :-)

Long story short: remove the field `deviceZoom` and remove the condition `if (deviceZoom == -1)`.

Phillipus:
If PNG images are used for the cursors then separate source and mask are not needed as the PNG format has transparency.

ptziegler:
> If PNG images are used for the cursors then separate source and mask are not needed as the PNG format has transparency.

The new images still have a white background, to make sure that they look exactly the same.

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
ptziegler: That you for the deeper insight in how this is handled on the SWT side of things. This is pretty similar to what I had in mind.

Though there is a separate issue which makes me concerned about how much we can get out of it in the first place. If the zoom changes, we normally would also need to refresh the cursors, as they are still using the old image data.

But those cursors are API, meaning we can't simply dispose them and create new objects...

ptziegler: With the latest state, the device zoom is no longer stored in a static variable, but always read directly from the system property.

ptziegler:
At 100% zoom:
image

At 200% zoom:
image

At 300% zoom:
image

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
fedejeanne: > That you for the deeper insight in how this is handled on the SWT side of things. This is pretty similar to what I had in mind.
> 
> Though there is a separate issue which makes me concerned about how much we can get out of it in the first place. If the zoom changes, we normally would also need to refresh the cursors, as they are still using the old image data.
> 
> But those cursors are API, meaning we can't simply dispose them and create new objects...

Yes, that is true. The problems are that:
1. The cursors are **constants** in `SharedCursors` _i.e._ one can not simply recreate them when the zoom changes
2. `Image` and `ImageData` don't really know anything about zoom changes which means once you create a `Cursor` out of an `Image`/`ImageData`, you're stuck with the zoom level at which they were created.

Point nr. 2 means that one has to manually react to zoom level changes and recreate the cursors with an image/image data at the appropriate zoom level, but this breaks the contract established in point nr. 1.

Sadly, I don't see an easy way out of this problem, at least not at this level of abstraction (GEF is a consumer of SWT). Best case scenario, this kind of issues would be handled transparently by SWT and not by the consumer (GEF), but I know this is not trivial since our team has been working on the [HighDPI subject](https://github.com/orgs/vi-eclipse/projects/1/views/1?filterQuery=label%3AHiDPI) for quite a while now and this issue has also surfaced [somewhere else](https://github.com/vi-eclipse/Eclipse-Platform/issues/65).

I will take this to the team and we'll try to address it in https://github.com/vi-eclipse/Eclipse-Platform/issues/65 too.

azoitl:
> > If PNG images are used for the cursors then separate source and mask are not needed as the PNG format has transparency.
> 
> The new images still have a white background, to make sure that they look exactly the same.
> 
> Edit: Clients may also still run GEF with an older Eclipse version. By omitting the mask, we would break backwards compatibility.

@ptziegler we already have some cursors without mask since the last release. I hope I didn't break anything there. But the mask always made problems at least for these cursors. So I'm happy when it is gone.

ptziegler:
@azoitl I talked to Heiko over at the Eclipse Platform PR and my current proposal is that for the new images, I drop the white background for transparency. In order to remain backwards compatible, we still need to use the mask. But for future cursors, it would then no longer be necessary.

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
ptziegler: > Point nr. 2 means that one has to manually react to zoom level changes and recreate the cursors with an image/image data at the appropriate zoom level, but this breaks the contract established in point nr. 1.

That's what I'm afraid of. And it's not just affecting the SharedCursors class, but also the Draw2D Cursors class containing static instances of the native SWT cursors. In the long term, I believe the most sensible thing to do would be to deprecate those static fields and have something similar to the ImageDataProvider, except for cursors. 

GEF can then listen to SWT.ZoomChanged events and reload the cursors accordingly. But that is going way beyond the scope of this PR and something we have to consider at a later date. Thank you again for your input.

azoitl:
> @azoitl I talked to Heiko over at the Eclipse Platform PR and my current proposal is that for the new images, I drop the white background for transparency. In order to remain backwards compatible, we still need to use the mask. But for future cursors, it would then no longer be necessary.

Thx for checking and the feedback.

ptziegler:
As a quick update: With the 2024-09, the mask & source images are going to be marked as "for removal", being replaced by a new constants that combines both into one.

So instead of:

> createCursor(ISharedImages.IMG_OBJS_DND_INVALID_SOURCE, ISharedImages.IMG_OBJS_DND_INVALID_MASK);

One can simply do:

> createCursor(ISharedImages.IMG_OBJS_DND_INVALID);

ptziegler:
So in principle, we could delay this PR until the M2 and then refactor the code or merge it now and keep the "old" images. The first would break backwards compatibility and require clients to also update to the latest Eclipse version, when updating GEF.

For my perspective, I would prefer to keep using the old icons and wait out the 2-year deprecation period.

ptziegler:
I've updated the PR to also scale the shared images. While it can't handle dynamic DPI changes, it at least won't throw an IllegalArgumentException anymore.

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
```diff
@@ -67,4 +72,45 @@ public void stop(BundleContext context) throws Exception {
 		super.stop(context);
 	}
 
+	/**
+	 * Convenience method for getting the current zoom level of the active device.If
+	 * on MacOS or Linux (x11 window system) or if the device zoom couldn't
+	 * otherwise be determined, this method returns {@code 100} as default value.
+	 */
+	@SuppressWarnings("nls")
```
fedejeanne, line 80: Better add the proper `//$NON-NLS-X$` and remove this, otherwise you'll be possibly silencing valid warnings that may be introduced in future changes.

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
```diff
@@ -67,4 +72,45 @@ public void stop(BundleContext context) throws Exception {
...
+			return 100;
+		}
+	}
+
+	/**
+	 * Convenience class to get the image data for a given zoom level. If no image
+	 * for the requested zoom level exists, the image data may be an auto-scaled
+	 * version of the native image and may look more blurred or mangled than
+	 * expected.
+	 */
```
fedejeanne, line 101:
```suggestion
	/**
	 * Convenience method to get the image data for a given zoom level. If no image
	 * for the requested zoom level exists, the image data may be an auto-scaled
	 * version of the native image and may look blurred or mangled.
	 */
```

org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
```diff
@@ -1557,37 +1559,42 @@ private static class DragCursors {
...
 				}
 			}
 			return cursors[code];
 		}
 
+		/**
+		 * Creates a new cursor using the shared images as source and mask,
+		 * respectively. The cursors are created with respect to the current device
+		 * zoom, with the hotspot being the center of the image.
+		 */
```
fedejeanne, line 1586: Add the parameters to the javadoc and mention that they need to be names of shared cursors.

fedejeanne (commented):
Just some nitpicks, the rest LGTM.

Regarding breaking backwards compatibility or not (https://github.com/eclipse/gef-classic/pull/479#issuecomment-2241762468) I'd say you make it clear that something will break and give time for the clients to adapt. If the appropriate methods are being deprecated _for removal_ anyway then there's no chance this change is going to be forgotten in 2 years anyway (one can live with the technical debt as long as there's an end in sight).

One last thing: I'd like to test this too. Can you please add a short "how to test" section to the description?

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
ptziegler: > Which class is that? If it's using the native cursors (i.e. the constructor that takes a style as parameter) then they shouldn't be affected by this issue.

I'm talking about this [Cursors](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/Cursors.java) class, where we create static instances of the native cursors. Looking at the actual implementation, it looks like they don't actually work with ImageData instances, so you are indeed correct and they should be updated automatically.

org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
ptziegler: I've updated the PR to include the three things you've pointed out.

fedejeanne (approved):

azoitl (approved):

Issue #465, opened 2024-07-05
Convert gif images to png format


Phillipus:
GEF uses several gif images - 

https://github.com/eclipse/gef-classic/tree/master/org.eclipse.gef/icons
https://github.com/eclipse/gef-classic/tree/master/org.eclipse.gef/src/org/eclipse/gef/internal/icons

Some distributions of Linux might not have support to load gif files. See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1307

Perhaps these could be converted to png format in both standard and @2x sizes? (I don't have the artistic skill to do this)

azoitl:
I can give it a try. Especially some of the internal icons would also be available in the platform. Maybe we can take some of theirs and reduce ours.

ptziegler:
It's still wild to me that projects plan on dropping something as widespread as gifs...

Just as a note: From my experience, pngs tend to be much larger than gifs, unless you crank the compression level to the limit. 

I also remember it being possible to request funding from the EF. Perhaps it makes sense to commission a new set of icons instead? Especially since some icons (prominently the cursors) did not age gracefully.
Though I wouldn't even know where to start there...

azoitl:
The biggest problem I see with gifs is that AFAIK there is no alpha channel. PNG icons blend here much nicer in and this is a plus for dark mode.

merks:
Yes, there is one "color" that can be 100% transparent, but definitely no alpha.

azoitl:
In the last two days I started working on migrating the align icons to pngs and also added 2x versions. Not fully done yet but you can see where it is going here: https://github.com/eclipse/gef-classic/commit/5cf918ec254c839c9beb5dffac34a9a2063feb19

I think for most icons I will be able to produce nice versions for the 2024-09 release.

One thing I noticed where pngs are also better is that you can get nicer gradients as png allow full 8 bit rgba colors. This is especially helpful for making icons fitting to the Eclipse UI guidelines.

Phillipus:
@azoitl Thanks! They look good. :-)

ptziegler:
A quick search resulted in the following list of GIFs that still need to be converted. @azoitl What, if any, icons are you currently working on? Most of the icons belong to the GEF bundle and I'd like to avoid both of us working on the same files.

- org.eclipse.gef
	- org/eclipse/gef/internal/icons/
		- [ ] add.gif
		- [ ] add-disabled.gif
		- [ ] arrow16.gif
		- [ ] arrow24.gif
[...]

PR #480, opened 2024-07-18
Migrated all Match Icons to PNGs #465

azoitl:
As part of this migration the icons where polished, harmonized with Eclipse UI Guidelines and a 2x version is added.

https://github.com/eclipse/gef-classic/issues/465


ptziegler:
Just a minor thing, but it feels like some of the larger icons are a little blurry.

For example, this is your icon:
matchsize@2x

Where I would expect something closer to this, where the arrows are a little bit more crisp.
matchsize@2x

ptziegler:
Unrelated: Perhaps it also makes sense to look into converting those icons into SVGs, at some point. I believe that could save us a lot of work in the long run, when we want to create e.g. icons in 4x resolution.

azoitl:
The blurriness on the arrows comes from how I tried to keep the original design. The arrows have blue and orange halos. I didn't understand where this came from. With your feedback I'll try to get rid of this and get them crisper. 

Not sure if SVG will help to much. Normally at the sizes of our icons manual adjustments will still be needed. At least from my experience sofar. But happy to experiment a bit later.

ptziegler:
> The arrows have blue and orange halos. I didn't understand where this came from. With your feedback I'll try to get rid of this and get them crisper.

I've noticed that as well. But in my example I've reduced it to just a single pixel. Or generally, I get the impression that anti-aliasing doesn't really work well with such small icons.

> Not sure if SVG will help to much. Normally at the sizes of our icons manual adjustments will still be needed. At least from my experience sofar. But happy to experiment a bit later.

I'm currently cleaning up some WindowBuilder icons and so far, I had a very positive experience (positive as in the result, not the actual work of creating them... 😓). At least for geometric shapes, it think it's very helpful. But as I said: Completely unrelated.

ptziegler:
@azoitl At the end of the day, I don't mind either way. If you think the icons are good, go for it.

azoitl:
> @azoitl At the end of the day, I don't mind either way. If you think the icons are good, go for it.

Just polished them a bit more. I may give also SVGs a try. Always keen to improve my drawing skills.

ptziegler:
Thank you going above and beyond for those icons! Do you want to contribute the SVGs as part of this or a separate pull request?

azoitl:
> Thank you going above and beyond for those icons! Do you want to contribute the SVGs as part of this or a separate pull request?

Would try them for some new icons. So for a separate PR.

PR #481, opened 2024-07-19
Mark Draw2D Cursors class as deprecated and for-removal

ptziegler:
Using static cursors is fundamentally flawed when considering dynamic DPI changes. The static cursors created by this class are created with respect to the current display zoom. Those objects are not updated, should this value change at a later time. Clients need to listen to the SWT.ZoomChanged event and update their cursors via the CursorProvider interface.

### DIFF
--- a/org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/scrollpane/InternalFrame.java
+++ b/org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/scrollpane/InternalFrame.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2023 IBM Corporation and others.
+ * Copyright (c) 2005, 2024 IBM Corporation and others.
  *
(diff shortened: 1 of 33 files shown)


ptziegler:
As discussed in https://github.com/eclipse/gef-classic/pull/479: The static Cursor objects don't support dynamic DPI changes. Meaning those instances keep using the initial display zoom, even though the level has already been changed.

This is effectively just an experiment where I mark the Cursors class as deprecated and update all references with a "CursorProvider", i.e. an interface that is supposed to always return a cursor that matches the current display zoom.

But even now this is already becoming a mess, as it requires changes to both the Draw2D figures and GEF tools. @azoitl wdyt?

ptziegler:
Alternatively, we could also keep this CursorProvider purely internal and push the responsibility of updating the cursor to the client.

ptziegler:
A misunderstanding on my side... The native cursors are updated upon DPI changes, but not the shared cursors.

PR #482, opened 2024-07-20
Create SVG of Draw2D CheckBox icons and recreate them as png

ptziegler:
This adds 96 and 192 DPI variants for the "checked" and "unchecked" icons used in the CheckBox figure. A minor change has been done do the images, where the white pixels outside of the check box have been removed.

On Linux, the icons can be generated automatically via the build.sh script in the "icons" folder, which also moves them to the correct location in the bundle projects.


ptziegler:
A side-by-side view of all icons. The first column are the existing gif files, the other two the generated png files.

image

azoitl (approved):

ptziegler:
@azoitl A minor oversight which is why I've temporarily set it as draft again: SWT has no built-in mechanism for loading different versions of the same image, unlike you have with the JFace ImageDescriptors.

Which means Draw2D needs to provide its own logic via the new FileImageDataProvider.

azoitl (approved):

PR #486, opened 2024-07-22
Create high-DPI variants for more of the GEF icons

ptziegler:
This converts the following icons to PNG and adds a 192 DPI version for:
- folder_closed
- move_down_disabled
- move_up_disabled
- next_nav
- palette_layout
- palette_view
- prev_nam
- zoom_in
- zoom_out

The following icons are not used and have been removed:
- palette_left
- palette_right

Most of the SVG files have been taken from the Eclipse platform project, in order to remain consistent with the rest of the IDE.


azoitl:
@ptziegler thx for spending the effort on the svgs. However I have an issue with several of the icons. Most of them are also available in some form in the Eclipse Platform. There they look significantly different. Mostly because some gradients are missing. I still have to brush up my Inkscape know how to do this. 

But this also brings me to another question. Should we really duplicate Icons from the Eclipse Platform in GEF Classic?

ptziegler:
> However I have an issue with several of the icons. Most of them are also available in some form in the Eclipse Platform. There they look significantly different. Mostly because some gradients are missing. I still have to brush up my Inkscape know how to do this.

Most of the icons have been taken directly from the platform. So the source files should -in principle- be identical. But I believe the Eclipse platform is using the [org.eclipse.images.renderer](https://github.com/eclipse-platform/eclipse.platform.images/tree/master/org.eclipse.images.renderer) plugin for generating the png files, which internally is using Batik instead of Inkspace. But from my experience, Batik is really bad at creating smooth edges, which is why I try to avoid it.

[folder](https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.ui.ide/icons/full/obj16/folder.svg)
[next_nav](https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.ui.glance/icons/full/elcl16/next_nav.svg)
[prev_nav](https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.compare/icons/full/elcl16/prev_nav.svg)
[zoom_in](https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.ui.workbench.texteditor/icons/full/etool16/zoomIn.svg)
[zoom_out](https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.ui.workbench.texteditor/icons/full/etool16/zoomOut.svg)

> Should we really duplicate Icons from the Eclipse Platform in GEF Classic?

If those icons are accessible via some public API then yes. But I don't believe this is the case for any of the icons here and we can't really rely on the path to stay the same indefinitely. Just like clients can't rely on e.g. `org/eclipse/gef/internal/icons/zoom_in.gif` to be accessible in all GEF releases.

azoitl:
Oh sorry this is my fault. I didn't fully get the SVG concept yet and how it looks in Inkscape. Yes I would definitely use Inkscape for rendering the pngs. 

> If those icons are accessible via some public API then yes. But I don't believe this is the case for any of the icons here and we can't really rely on the path to stay the same indefinitely. Just like clients can't rely on e.g. org/eclipse/gef/internal/icons/zoom_in.gif to be accessible in all GEF releases.

Most of the icons have rather stable names. But they are in plugins which we are currently referencing. So I fear duplicating is the only chance we have here.

ptziegler:
Maybe it's better to approach this from a different direction... What is one of the icons that look different? Then I could try to play around and see what exactly is causing those deviations.

azoitl:
> Maybe it's better to approach this from a different direction... What is one of the icons that look different? Then I could try to play around and see what exactly is causing those deviations.

Sorry I was not clear enough. By now I think it is an optical illusion when I look at the SVG files. The pngs are really as they should and I was just confused. I would merge it now as is.

azoitl (approved):

ptziegler:
> By now I think it is an optical illusion when I look at the SVG files

Ah, I see what you mean. I believe Inkscape makes the colors slightly transparent, so that the background is shining through or something along the lines. It definitely it slightly different from what they look like when exported.

PR #488, opened 2024-07-23
Cleanup GEF cursors and remove some of the blockiness

ptziegler:
This creates SVG files for the plug/tree cursors and also the shared arrow icons, resulting in much smoother sides.

Following icons are affected:
- arrow16
- arrow24
- plug-cursor
- plugnot-cursor
- tree_add-cursor
- tree_move-cursor


ptziegler:
The cursors are a weird one... They are clearly based on the Windows cursors, given that the shaft is slightly tilted, making it point to a different position than the tip of the arrow.

image

And for the plug, I have no idea how to make it look prettier. I'm also not sure if we should even keep it like this. In modern UI, icons tend to be more simplistic, which causes this 3dish icon to look dated. Perhaps it'd be better to replace it with a flat image of a plug?

azoitl:
To add my five cents here: I never liked the connection cursors, and in 4diac IDE we are not using them for a very long time. We resorted to take the most suitable system cursor instead (e.g., crosshair for connection creation).  This gives a much more native look and feel. Because any cursor we design, even if done very nice and perfect, will be wrong on any at least 2 of the 3 supported OSes.

ptziegler:
I'd have to double check, but I'm fairly certain we do the same for our product. The cursors just don't look very good. Which is also why I'm hesitant to go above and beyond here.

I tried looking into the possibility of decorating the native cursors so that one could piece together the image data, but I sadly don't believe that's possible.

ptziegler:
I finally got around to looking at those cursors again. I think the original plug is just too far gone, so I redesigned it from scratch. I've also adapted the cursor size, so that the "tree" and "plug" cursors are now the same.

azoitl (approved):
Thx @ptziegler that you spend more work on that. Especially as Eclipse SWT will get SVG support with the next release we could then think of directly using the SVGs later.

ptziegler:
> Eclipse SWT will get SVG support with the next release we could then think of directly using the SVGs later.

I'm already working on a PR that I want to push later today for exactly that ;D We just have to be careful to remain backwards compatible.

PR #490, opened 2024-07-23
Remove deprecated GEF StackView and Debug Console

ptziegler:
The classes have been deprecated since almost 20 years and are not used anywhere.

@azoitl The classes have been deprecated for two decades, so the only difference between marking it for-removal now or in the second next release is us forgetting about it :smiling_face_with_tear:. Might as well do it now.


azoitl:
I was again not precise in my wording (should really work on that). What I meant is that we should start the process of removing this deprecated API. I thought this would be in 2 or 4 releases but I checked the [Eclipse Platform API removal policy ](https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/ApiRemovalProcess.md) and there it is two years after announcement. 

I guess we would follow a similar policy. So we also should ask the PMC about this or?

Finally as the API in question was marked deprecated in the Javadoc with version 3.1 should the deprecation annotation not use the same version?

ptziegler:
> I guess we would follow a similar policy. So we also should ask the PMC about this or?

As far as I understand, the PMC approval is only required for the Eclipse Platform, but not for any other Eclipse projects. Otherwise this sounds like a lot of work... But I'm happy to be corrected.

In any case, waiting two years before the actual removal sounds like a good idea to me.

> Finally as the API in question was marked deprecated in the Javadoc with version 3.1 should the deprecation annotation not use the same version?

That is true. I picked the next release because the removal annotation is added there, but 3.1 would indeed be correct.

merks:
You can have your own policy.  No one needs to approve it.  Given how long it's already been deprecated, I think it's reasonable to remove then sooner rather than later.  For newly deprecated things you might want to leave more runway, but such long, long deprecated things surely doing need a long runway.

ptziegler:
In that case I would simply remove those classes. GEF doesn't contribute them to the UI, so it's not like clients accidentally stumble upon them.

merks:
Yes, my sense is just to delete it now and be done with it.

azoitl:
Wow, the first thing that came to my mind on this PR is "well, that escalated quickly" ;-) But as I always say: Deleted code has no bugs! So I'm in strong favor of removing this and also cleaning up other deprecated code. This should really make our live easier.

I also heard that some of our users already removed any deprecated code parts (e.g., I have seen commits from @pcdavid in GMF runtime to handle this). Therefore if we start that slowly we should be fine.

pcdavid:
From a quick check it looks like neither GMF Runtime nor Sirius Desktop would not be impacted by the removals in this PR.

azoitl:
With all the feedback I would merge that now. So that we have it rather early in the development cycle.

PR #491, opened 2024-07-24
Mark deprecated elements in Draw2D Geometry component as "for Removal"

ptziegler:
This marks all fields, methods, constructors and classes which have been marked as deprecated as "for-removal", to be deleted or made private in a future release.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -22,19 +22,21 @@ public class PrecisionPoint extends Point {
 	/**
 	 * Double value for X
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseX(double)} and {@link #preciseX()} instead.
 	 *             This field will become private in future versions.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
 	public double preciseX;
 
 	/**
 	 * Double value for Y
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseY(double)} and {@link #preciseY()} instead.
 	 *             This field will become private in future versions.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
 	public double preciseY;
 
 	/**
@@ -301,12 +303,14 @@ public class PrecisionPoint extends Point {
 	/**
 	 * Updates the integer fields using the precise versions.
 	 *
+	 * @noreference This method is not intended to be referenced by clients.
 	 * @deprecated This method should not be accessed by clients any more (it will
 	 *             be made private in future releases). The update of integer and
 	 *             precision fields is performed automatically if {@link #preciseX}
 	 *             and {@link #preciseY} field values are not manipulated directly,
 	 *             but only via respective methods offered by this class.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
 	public final void updateInts() {
 		updateXInt();
 		updateYInt();
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2010 IBM Corporation and others.
+ * Copyright (c) 2003, 2024 IBM Corporation and others.
  *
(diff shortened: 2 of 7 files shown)


ptziegler:
Inspired by https://github.com/eclipse/gef-classic/pull/490. It probably makes sense to slowly nudge clients into migrating off of the deprecated API so that we can clean up some of the old code.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
```diff
@@ -26,38 +26,42 @@ public final class PrecisionRectangle extends Rectangle {
 	/**
 	 * Double value for height
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseHeight(double)} and
 	 *             {@link #preciseHeight()} instead. This field will become private
 	 *             in the future.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
```
azoitl, line 34: This field should not be removed but turned private!

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
```diff
@@ -26,38 +26,42 @@ public final class PrecisionRectangle extends Rectangle {
...
 
 	/**
 	 * Double value for width
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseWidth(double)} and {@link #preciseWidth()}
 	 *             instead. This field will become private in the future.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
```
azoitl, line 44: This field should not be removed but turned private!

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
```diff
@@ -26,38 +26,42 @@ public final class PrecisionRectangle extends Rectangle {
...
 
 	/**
 	 * Double value for X
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseX(double)} and {@link #preciseX()} instead.
 	 *             This field will become private in the future.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
```
azoitl, line 54: This field should not be removed but turned private!

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
```diff
@@ -26,38 +26,42 @@ public final class PrecisionRectangle extends Rectangle {
...
 
 	/**
 	 * Double value for Y
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseX(double)} and {@link #preciseY()} instead.
 	 *             This field will become private in the future.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
```
azoitl, line 64: This field should not be removed but turned private!

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
```diff
@@ -22,19 +22,21 @@ public class PrecisionPoint extends Point {
 	/**
 	 * Double value for X
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseX(double)} and {@link #preciseX()} instead.
 	 *             This field will become private in future versions.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
```
azoitl, line 29: This field should not be removed but turned private!

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
```diff
@@ -22,19 +22,21 @@ public class PrecisionPoint extends Point {
...
 
 	/**
 	 * Double value for Y
 	 *
-	 * @noreference
+	 * @noreference This field is not intended to be referenced by clients.
 	 * @deprecated Use {@link #setPreciseY(double)} and {@link #preciseY()} instead.
 	 *             This field will become private in future versions.
 	 */
+	@Deprecated(since = "3.7", forRemoval = true)
```
azoitl, line 39: This field should not be removed but turned private!

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
ptziegler: Of course it's not the field that's removed, but rather the access to said field. What I want to achieve is that users get a "removal" warning, rather than a "deprecation" warning because from experience people tend to treat a deprecation like a "suggestion" and will only adapt their code once it's actually removed.

Sadly, I don't believe the annotation allows for such a level of granularity so I went with this, just to really bring the point across that people should move to the getters and setters instead.

azoitl: I'm stupid the what will happen is anyhow in the Java-doc comment.

azoitl (approved):

Issue #492, opened 2024-07-28
ViewportAutoexposeHelper is very intensive on CPU


Phillipus:
The [ViewportAutoexposeHelper](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/editparts/ViewportAutoexposeHelper.java) consumes a lot of CPU if you hover a creation tool near the edge of the Viewport.

1. Add a debug line `System.out.println(where);` to the `public boolean step(Point where)` method so you can see when it's called
2. Select a creation tool from the Palette
3. Hover your mouse near the edge of the viewport
4. Watch many calls being made to `ViewportAutoexposeHelper#step` and the CPU usage rise in System monitor

I don't know if this could be optimised somehow to be less busy?

ptziegler:
Main culprit seems to be the the following methods, which mercilessly fires one update after the other:

https://github.com/eclipse/gef-classic/blob/a9f29fd0ad586397e45c0223908c6b0b72e9cb3f/org.eclipse.gef/src/org/eclipse/gef/tools/TargetingTool.java#L98-L108

https://github.com/eclipse/gef-classic/blob/a9f29fd0ad586397e45c0223908c6b0b72e9cb3f/org.eclipse.gef/src/org/eclipse/gef/tools/TargetingTool.java#L347-L354

There was a similar discussion [here](https://github.com/eclipse/gef-classic/issues/430) which is why I'd propose a similar approach: To copy the JFace [Throttler](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface/src/org/eclipse/jface/util/Throttler.java) class to Draw2D and then add a delay between the updates. 

The updates are already done asynchronously, so the only difference would be that instead of e.g. every 1ms, the methods are called every 100ms.

ptziegler:
I forgot that those classes are in the GEF bundle, so we can already use the Throttler class without adding any new dependencies. 🎉 

See https://github.com/eclipse/gef-classic/pull/494.

Phillipus:
That was quick! :-)

I tested it and it improves the polling rate. The only thing is that when the viewport scrolls into view it's a bit more jerky and not so smooth.

I wonder if there is another way to optimise this by not making the calls if the viewport is currently scrolled into view?

Phillipus:
I set the value to 1ms in the Throttler and the CPU usage is way down even though this is a low value. So perhaps the Throttler is doing something to pace things?

[...]

PR #494, opened 2024-07-28
Fix: ViewportAutoexposeHelper is very intensive on CPU

ptziegler:
The calls do doAutoexpose() are currently fired relentlessly one after the other, leading to high (and mostly unnecessary) CPU usage. By using the JFace Throttler class we add a small delay between those events.

The delay is an arbitrary but fixed value; Large enough to avoid the excessive CPU cost, but small enough to not be noticeable by the user.

Resolves https://github.com/eclipse/gef-classic/issues/492


ptziegler:
Not sure about the "delay" value. I've currently chosen 100ms, but perhaps something else is better? Perhaps it should even be configurable via a system property?

ptziegler:
I suppose the fix is not needed, in the end...

Issue #495, opened 2024-07-30
Figure::findMouseEventTargetInDescendantsAt() seems not reentrant


ddscharfe:
This is similar to https://bugs.eclipse.org/bugs/show_bug.cgi?id=81725 (fixed by 4a091db1f84922e59ee0f7d36586c8657e8c9e86), but for finding mouse event targets. There was already a ticket for this (https://bugs.eclipse.org/bugs/show_bug.cgi?id=226103) which for some reason was resolved as `WORKSFORME` later.

When a figure has multiple children where one child uses local coordinates, this can lead to wrong event target detection, see comments 1 and 2:

```Java
	protected IFigure findMouseEventTargetInDescendantsAt(int x, int y) {
		PRIVATE_POINT.setLocation(x, y);
		translateFromParent(PRIVATE_POINT);

		if (!getClientArea(Rectangle.SINGLETON).contains(PRIVATE_POINT)) {
			return null;
		}

		for (IFigure fig : getChildrenRevIterable()) {
			if (fig.isVisible() && fig.isEnabled() && fig.containsPoint(PRIVATE_POINT.x, PRIVATE_POINT.y)) {
                                // (1) This call will change PRIVATE_POINT if fig uses local coordinates
				fig = fig.findMouseEventTargetAt(PRIVATE_POINT.x, PRIVATE_POINT.y);
				if (fig != null) {
					return fig;
				}
                                // (2) subsequent iterations will use the translated coordinates
			}
		}
		return null;
	}
```

ptziegler:
>  There was already a ticket for this (https://bugs.eclipse.org/bugs/show_bug.cgi?id=226103) which for some reason was resolved as WORKSFORME later.

My assumption is that this is a misunderstanding regarding the example provided by Manuel. I can see the same issue with the latest version (clicking the button should NOT highlight the yellow rectangle), so the wrong behavior definitely still exists.

Storing the position in local coordinates (similar to bug [81725](https://bugs.eclipse.org/bugs/show_bug.cgi?id=81725))  solves the issue, but I'm still working on integrating the snippet in our test suite to avoid any regressions.

PR #498, opened 2024-08-06
Bug 226103 - findMouseEventTargetAt problem with local coordinate system

ptziegler:
A figure using local coordinates translates the given point for its children at the beginning of findMouseEventTargetInDescendantsAt(int, int). However, during iteration over its children, the translated point is stored in a  static variable PRIVATE_POINT. The value would be changed by a child, if the child also uses local coordinates. This affects the following queries to the rest of children, if no figure is found under that child.

Fixes https://github.com/eclipse/gef-classic/issues/495

### DIFF
--- a/org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
@@ -2,14 +2,17 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.draw2d.tests
-Bundle-Version: 3.13.100.qualifier
+Bundle-Version: 3.14.0.qualifier
+Import-Package: org.slf4j
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.core.resources;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.draw2d;bundle-version="[3.16.0,4.0.0)",
+ org.eclipse.draw2d.examples,
  org.junit,
- org.eclipse.jface;bundle-version="[3.2.0,4.0.0)"
+ org.eclipse.jface;bundle-version="[3.2.0,4.0.0)",
+ org.eclipse.swtbot.swt.finder;bundle-version="4.2.1"
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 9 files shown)


org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
```diff
@@ -2,14 +2,17 @@ Manifest-Version: 1.0
...
 Bundle-SymbolicName: org.eclipse.draw2d.tests
-Bundle-Version: 3.13.100.qualifier
+Bundle-Version: 3.14.0.qualifier
+Import-Package: org.slf4j
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.core.resources;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.draw2d;bundle-version="[3.16.0,4.0.0)",
+ org.eclipse.draw2d.examples;bundle-version="3.15.400",
```
azoitl, line 12: shouldn't this be 3.16.0 and should we really be so explicit?

org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
```diff
@@ -2,14 +2,17 @@ Manifest-Version: 1.0
...
  org.junit,
- org.eclipse.jface;bundle-version="[3.2.0,4.0.0)"
+ org.eclipse.jface;bundle-version="[3.2.0,4.0.0)",
+ org.eclipse.swtbot.swt.finder;bundle-version="4.2.1"
```
azoitl, line 15: I guess greater or equal to 4.2.1 would be better here or?

org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
ptziegler: > shouldn't this be 3.16.0 and should we really be so explicit?

I think I've raised the version number after adding it as a dependency. But you're right. For dependencies within the same project, the version constraint is probably not needed...

org.eclipse.draw2d.tests/META-INF/MANIFEST.MF
ptziegler: Isn't that exactly what's described here?

image

azoitl: I always thought for ranges you need some range qualifier. Just checked on the OSGI spec and learned I was wrong. Really need to deepen my OSGI knowledge. An OSGI book is already on my reading list. Recommendations for more are warmly welcome.

azoitl (approved):

PR #499, opened 2024-08-09
Cleaned Ruler classes #155

azoitl:
Added types to generics and fixed issues reported by Java compiler and SonarLint.

https://github.com/eclipse/gef-classic/issues/155

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/DragGuidePolicy.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/DragGuidePolicy.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2010 IBM Corporation and others.
+ * Copyright (c) 2003, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -12,9 +12,9 @@
  *******************************************************************************/
 package org.eclipse.gef.internal.ui.rulers;
 
-import java.util.Iterator;
 import java.util.List;
 
+import org.eclipse.draw2d.Cursors;
 import org.eclipse.draw2d.Figure;
 import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.geometry.Point;
@@ -23,7 +23,6 @@ import org.eclipse.draw2d.geometry.Rectangle;
 import org.eclipse.gef.EditPart;
 import org.eclipse.gef.GraphicalEditPart;
 import org.eclipse.gef.Request;
-import org.eclipse.gef.SharedCursors;
 import org.eclipse.gef.commands.Command;
 import org.eclipse.gef.commands.UnexecutableCommand;
 import org.eclipse.gef.editparts.ZoomManager;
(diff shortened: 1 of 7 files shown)


org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/DragGuidePolicy.java
```diff
@@ -35,11 +34,12 @@
...
 public class DragGuidePolicy extends GraphicalEditPolicy {
 
-	private List attachedEditParts = null;
-	private IFigure dummyGuideFigure, dummyLineFigure;
+	private List<? extends EditPart> attachedEditParts = null;
+	private IFigure dummyGuideFigure;
+	private IFigure dummyLineFigure;
 	private boolean dragInProgress = false;
 
-	protected IFigure createDummyLineFigure() {
```
ptziegler, line 42: I assume this method used to be marked as protected, to be overwritten by subclasses? As far as I can tell, this method is only referenced by `getDummyLineFigure()`. Wouldn't it make more sense to simply inline this method? Perhaps it would even make sense to do the same for `createDummyGuideFigure()`...

You already did the same in the RulerEditPartFactory, might as well apply it here, too.

org.eclipse.gef/src/org/eclipse/gef/rulers/RulerProvider.java
```diff
@@ -142,8 +144,9 @@ public void getAccGuideValue(AccessibleControlEvent e, Object guide) {
 	 * @param guide the guide to which the model parts are attached.
 	 * @return list of attached model objects
 	 */
+	@SuppressWarnings("static-method")
 	public List getAttachedModelObjects(Object guide) {
-		return Collections.EMPTY_LIST;
```
ptziegler, line 146: Is it intentional that this method still returns a raw list?

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/GuideEditPart.java
```diff
@@ -112,7 +112,7 @@ protected IFigure createFigure() {
 		return new GuideFigure(isHorizontal());
 	}
 
-	protected GuideLineFigure createGuideLineFigure() {
+	private static GuideLineFigure createGuideLineFigure() {
 		return new GuideLineFigure();
```
ptziegler, line 116: Same as in the DragGuidePolicy. I would just inline this method.

org.eclipse.gef/src/org/eclipse/gef/rulers/RulerProvider.java
azoitl: At the time I was not sure about the type. After doing the getModelChildren() change in PR #501 I now know better. Changed it.

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/DragGuidePolicy.java
azoitl: I guess this came from the general design pattern in GEF Classic that subclasses can with such methods easily provide their own figures. However as the class is internal and I don't see different policies in the near future soon I think it is better to inline them as you suggested.

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/GuideEditPart.java
azoitl: done as suggested.

ptziegler (approved):

PR #500, opened 2024-08-10
Added type information to getEditPartRegisry #155

azoitl:
As part of this clean-up a new helper method has been added for getting an EditPart for a model instance and for getting the LayerManager from a viewer.

@ptziegler this PR has still a problem, or better reveal a misuse of the EditPartRegistry by the palette. The `SliderPaletteEditPart` is registering an instance of `PaletteAnimator` at the EditPartRegsitry of the PaletteViewer. This `PaletteAnimator` is then later requested by `DrawerEditPart`s. The problem is that `PlaetteAnimator` is not an `EditPart`.  This totally violates the API description of `EditPartViewer`. Both `SliderPaletteEditPart` and `DrawerEditPart` are internal classes so we have a bit more freedom. Still I'm not sure how to best solve the issue. Currently I have two ideas on my mind:

1.  Move the PaletteAnimator to the PaletteViewer, add there a getter and a factory method. So that clients can provide own animators. This would for me be the cleanest solution. However clients may expect that PaletteAnimator to be in the EditPart Registry.
2. Therefore my other solution would be to let PaletteAnimator implement the EditPart interface. This would definitely be less clean but not break clients.  

So I'm unsure what to do.

### DIFF
--- a/org.eclipse.gef/.settings/.api_filters
+++ b/org.eclipse.gef/.settings/.api_filters
@@ -64,6 +64,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="src/org/eclipse/gef/EditPartViewer.java" type="org.eclipse.gef.EditPartViewer">
+        <filter id="403804204">
+            <message_arguments>
+                <message_argument value="org.eclipse.gef.EditPartViewer"/>
+                <message_argument value="getEditPartForModel(Object)"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="src/org/eclipse/gef/ui/actions/AlignmentRetargetAction.java" type="org.eclipse.gef.ui.actions.AlignmentRetargetAction">
         <filter id="571473929">
             <message_arguments>
--- a/org.eclipse.gef/src/org/eclipse/gef/EditPartViewer.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/EditPartViewer.java
@@ -281,7 +281,20 @@ public interface EditPartViewer extends org.eclipse.jface.viewers.ISelectionProv
 	 *
 	 * @return the registry map
 	 */
-	Map getEditPartRegistry();
+	Map<Object, EditPart> getEditPartRegistry();
+
+	/**
+	 * Convenience method to look up an edit part for a given model element in the
+	 * EditPart registry.
+	 *
+	 * See also {@link #getEditPartRegistry()} for details on the EditPart registry.
+	 *
+	 * @param model the model object for which an EditPart is looked up
+	 * @return the edit part or null if for the given model no EditPart is
+	 *         registered
+	 * @since 3.19
+	 */
+	EditPart getEditPartForModel(Object model);
 
 	/**
 	 * Returns the <i>focus</i> <code>EditPart</code>. Focus refers to keyboard
--- a/org.eclipse.gef/src/org/eclipse/gef/editparts/AbstractEditPart.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/editparts/AbstractEditPart.java
@@ -1190,7 +1190,7 @@ public abstract class AbstractEditPart implements EditPart, RequestConstants, IA
 	 * this method if they need to unregister this EditPart in additional ways.
 	 */
 	protected void unregisterModel() {
-		Map registry = getViewer().getEditPartRegistry();
+		Map<Object, EditPart> registry = getViewer().getEditPartRegistry();
 		if (registry.get(getModel()) == this) {
 			registry.remove(getModel());
 		}
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/edit/LogicEditPart.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/edit/LogicEditPart.java
@@ -23,7 +23,6 @@ import org.eclipse.draw2d.geometry.Rectangle;
 
 import org.eclipse.gef.AccessibleEditPart;
 import org.eclipse.gef.ConnectionEditPart;
-import org.eclipse.gef.EditPart;
 import org.eclipse.gef.EditPolicy;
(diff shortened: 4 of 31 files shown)


ptziegler:
Hm... that's indeed a tricky one.

> this PR has still a problem, or better reveal a misuse of the EditPartRegistry by the palette. The SliderPaletteEditPart is registering an instance of PaletteAnimator at the EditPartRegsitry of the PaletteViewer. This PaletteAnimator is then later requested by DrawerEditParts. The problem is that PlaetteAnimator is not an EditPart. 

We have a similar issue with the [LayerManager](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.gef/src/org/eclipse/gef/editparts/LayerManager.java), though it is not as critical here, as this interface is only implemented by edit parts.

https://github.com/eclipse/gef-classic/blob/11ae8bf359dec43733ca37ff03c7c218e0f32b0e/org.eclipse.gef/src/org/eclipse/gef/editparts/LayerManager.java#L59-L61

> Therefore my other solution would be to let PaletteAnimator implement the EditPart interface. This would definitely be less clean but not break clients.

I don't like this at all. If we add this interface now, we'll be stuck with it indefinitely. Outside of this workaround, I don't see why a PaletteAnimator should be an edit part, but e.g. the LayoutAnimator should not.

> Move the PaletteAnimator to the PaletteViewer, add there a getter and a factory method. So that clients can provide own animators. This would for me be the cleanest solution. However clients may expect that PaletteAnimator to be in the EditPart Registry.

As far as I can tell, there is no documented way for contributing a custom animator. So _technically_ we wouldn't break API if we kick it out of the edit part registry. Though you are absolutely right that we have no control over what clients may or may not do.

I assume how it should've been done initially is by adding the SliderPaletteEditPart to the registry and then access the PaletteAnimator via a getter. But we might as well do it right and expose it it via the PaletteViewer, as you described.

org.eclipse.gef/.settings/.api_filters
```diff
@@ -64,6 +64,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="src/org/eclipse/gef/EditPartViewer.java" type="org.eclipse.gef.EditPartViewer">
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.gef.EditPartViewer"/>
+                <message_argument value="getEditPartForModel(Object)"/>
```
ptziegler, line 71: Based on the documentation, the EditPartViewer interface shouldn't be implemented by clients. I wonder if one can instead add `@noimplement` and avoid this entry?

org.eclipse.gef/src/org/eclipse/gef/EditPartViewer.java
```diff
@@ -281,7 +281,22 @@ interface Conditional {
...
+	 *
+	 * See also {@link #getEditPartRegistry()} for details on the EditPart registry.
+	 *
+	 * @param model the model object for which an EditPart is looked up
+	 * @return the edit part or null if for the given model no EditPart is
+	 *         registered
+	 * @since 3.19
+	 */
+	default EditPart getEditPartForModel(Object model) {
+		return getEditPartRegistry().get(model);
```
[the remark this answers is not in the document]
azoitl: I get your point. However Java does not allow that default methods of interfaces are final. I can only add a @nooverride. Would this be sufficient.

org.eclipse.gef/.settings/.api_filters
azoitl: I fear still we need that. The reason for the problem is that there could be children that have a method with the same name and we would break these children.

org.eclipse.gef/src/org/eclipse/gef/EditPartViewer.java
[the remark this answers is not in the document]
ptziegler: Then I would only keep the signature in the interface and move the actual implementation to the AbstractEditPartViewer (where it can then be final). The documentation explicitly says that clients should never implement the interface directly.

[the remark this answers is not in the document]
azoitl: done

org.eclipse.gef/.settings/.api_filters
ptziegler: Fair point.

org.eclipse.gef/src/org/eclipse/gef/editparts/AbstractEditPart.java
```diff
@@ -844,7 +844,9 @@ protected final void registerAccessibility() {
 	 * this method if they need to register this EditPart in additional ways.
 	 */
 	protected void registerModel() {
-		getViewer().getEditPartRegistry().put(getModel(), this);
+		@SuppressWarnings("unchecked")
+		Map<Object, EditPart> registry = (Map<Object, EditPart>) getViewer().getEditPartRegistry();
```
ptziegler, line 848: I wonder if it's really a good idea for the `getEditPartRegistry()` to return a map with wildcards. All clients that register their own models now have to do the same cast in their edit parts...

azoitl: Thx for bringing it up. I noticed that also and was not sure. I was either thinking of non wildcard returns or an additional helper `registerEditPart(EditPart ep, Object model)`. What do you think is better?

ptziegler: I believe the cases where the registry only contains edit parts of a single type (i.e. where one would cast the map to something other than Map\<Object, EditPart\>) are mostly academical. So I would simply remove the wildcard, which would then hopefully minimize the amount of change required by downstream projects.

azoitl: You have a point here. Will change accordingly.

azoitl: done

ptziegler (approved):

PR #511, opened 2024-08-17
Reduced bluryness of italic and underline style icons #465

azoitl:


ptziegler (approved):
I personally like the bold text a little more because it's easier to read, but in the end, both are fine.

Issue #512, opened 2024-08-18
GEF, icon is missing: icon "palette_view.gif" (org.eclipse.gef/icons/palette_view.gif)


speckyspooky:
Currently we work on the next release of BIRT 4.17.
There we use the latest eclipse plattform version:
Version: 2024-09 (4.33)
Build id: I20240815-1810

The eclipse-designer with active BIRT-view has one register with the icon of a color palette normally.
But currently the icon is not longer listed and the eclipse warning means that the icon is missing from "org.eclipse.gef/icons/palette_view.gif)" - see screen below.

Instead of the palette icon we get a red rectangle or a small fram-icon (as default icon).

Is the cion complettly removed or is the icon changed?
What can we do to fix it?

I tried a small research here on GEF but I figured out no good hint of a change for that icon.

![grafik](https://github.com/user-attachments/assets/a7540908-3394-4090-833c-19f40712cacf)

ptziegler:
The palette view is contributed here, where the new png icon is used:

https://github.com/eclipse/gef-classic/blob/e5d6bac76ff38625b7c04877c0582452f04055ff/org.eclipse.gef/plugin.xml#L211-L220

A quick search doesn't show any indication that the old gif icon is still referenced, so I don't believe there is any misconfiguration on GEF side.

![image](https://github.com/user-attachments/assets/222a806a-b2ca-412d-b213-d4ae6c10e692)

After trying the latest GEF plugin, I can see that the icon is correctly shown in the Palette view:

![image](https://github.com/user-attachments/assets/ff901097-9940-4291-82a8-f685c5bde44a)

Weirdly enough, I get the same error when opening the old BIRT workspace I have checked out...

![image](https://github.com/user-attachments/assets/1cf3af0e-335b-4798-b351-cb0a5be3e3b1)

A quick search reveals that the old icon is still referenced in the workspace.xmi file located in the metadata folder. So I agree with @merks theory that the Eclipse itself is caching those URLs which are then not updated properly.

![image](https://github.com/user-attachments/assets/08b5b3fd-4b0d-4055-aa38-62f5b538c66b)

ptziegler:
After starting the application with `-DclearPersistedState`, I now get the following error:

```
!MESSAGE Unable to create resource URLImageDescriptor(platform:/plugin/org.eclipse.gef/src/org/eclipse/gef/internal/icons/palette_view.png)
!STACK 0
[...]

PR #513, opened 2024-08-18
Duplicate palette icon to avoid issues when using the Palette View

ptziegler:
The path of the icon URI in the plugin.xml file is using the workspace path beginning with "src/...". But in the actual bundle, this prefix doesn't exist. This commit restore the dedicated "icons" folder which also holds an instance of this file, which was removed with 24daefbe8edfb9eb2e4c3db7b8f5e318b1b33860.

See https://github.com/eclipse/gef-classic/issues/512


org.eclipse.gef/.classpath
```diff
@@ -7,5 +7,6 @@
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="icons"/>
```
merks, line 10: I'm quite sure this isn't needed.  None of the Platform bundles so this that:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui/.classpath

https://github.com/eclipse-platform/eclipse.platform.ui/blob/e3628a3a67c44771465d7d317b3424920c425090/bundles/org.eclipse.ui/plugin.xml#L1442-L1447

The build.properties will ensure the icons are in the final jar and it should work fine for debug launches as well.

merks (changes requested):
I thin it's best to remove the icons folder as a source folder.

org.eclipse.gef/.classpath
ptziegler: That value is indeed unnecessary. I've just removed it.

merks (approved):
Thanks.

azoitl (approved):

PR #517, opened 2024-08-19
Simplify Zest tests by using the SWTBot

ptziegler:
This removes classes such as the WidgetVisitor in favor of using the methods provided by the SWTBot. Note that because the SWTBot doesn't understand the Zest model, the tests still have to be executed in the UI thread.


ptziegler:
I noticed some sporadic test failures in the Zest test over the past few weeks. As part of cleaning up the suite, I've started to migrate part of the tests to SWTBot. Most noticeably the handling of widgets.

Perhaps in the future, one can create a proper "Zest" bot and contribute it to the test project and clean up those tests even more. But for the time being, this has to suffice.

azoitl (approved):

azoitl:
@ptziegler I think you are doing an amazing job for getting our test coverage up!

PR #526, opened 2024-08-21
[Zest 2.0] Cleanup compiler warnings in GraphContainer

ptziegler:
There are several instances where raw types are used instead of the GraphNode and GraphConnection widgets. Several private method can be marked as static.

The warnings caused by usage of the deprecated Zest 1.0 API have been suppressed, as those are needed for backwards compatibility to e.g. apply layouts using the old layout algorithms.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphContainer.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphContainer.java
@@ -15,7 +15,6 @@
 package org.eclipse.zest.core.widgets;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/GraphContainer.java
```diff
@@ -176,13 +177,9 @@ public void close(boolean animate) {
 		updateFigureForModel(getModelFigure());
 	}
 
-	private static void addNodeToOrderedList(List orderedNodeList, GraphNode node) {
-		Iterator orderedNodeIterator = orderedNodeList.iterator();
+	private static void addNodeToOrderedList(List<GraphNode> orderedNodeList, GraphNode node) {
 		int counter = 0;
-		while (orderedNodeIterator.hasNext()) {
-			// Look through the list of nodes below and find the right spot for
-			// this
```
azoitl, line 184: should we preserve this comment?

ptziegler: Good point. Seems like that comment was removed by the auto-formatter. I've restored it.

ptziegler:
@azoitl Thank you for taking the time and looking at those PRs! 😄 There are still a few more left, but I'm slowly approaching the state I'm happy with.

azoitl:
the least I can do when you spend so much time in getting us Zest 2

azoitl (approved):

PR #528, opened 2024-08-22
[Zest 2.0] Fix: Layout inside GraphContainers is not applied correctly.

ptziegler:
This can be observed when using the ZoomSnippet with the Zest 2.0 layouts; The nodes inside the containers arent't aligned according to the RadialLayoutAlgorithm but instead all placed in the top-left corner.

This is because we call applyLayout(false), which is a no-op operation for this algorithm.

One can reproduce this issue by checking the ZoomSnippet, where the nodes inside the containers are not position correctly, after expanding the container.


ptziegler:
See here:

https://github.com/eclipse/gef-classic/blob/e115fadae3208262d7bbec23fe766cc87a2e98bb/org.eclipse.zest.layouts/src/org/eclipse/zest/layouts/algorithms/RadialLayoutAlgorithm.java#L190-L194

Issue #518, opened 2024-08-20
[Zest 2.0] ZestRootLayer constants specify layer order


BauePhil:
So I already had a bit of a problem with the old version of the ZestRootLayer class, as that had a boolean constant to determine if the connections should be behind or in front of nodes. Now the new version is a lot cleaner but has the same issue. It just has a constant for each layer's order now.
My easy fix would be to have overridable getter methods that return the constant, which I could push ~~onto the PR~~.
A perhaps prettier solution could be to use something like a LayeredPane and layers, but I could not get this to work right now.
@ptziegler and @azoitl as you two seem to be the leading pair for the reintegration of zest 2.0 what do you think?

ptziegler:
If I understand you correctly, you want to make the order of layers configurable, so that e.g. the connections can be show in front of the nodes (and vice versa)?

> A perhaps prettier solution could be to use something like a LayeredPane and layers, but I could not get this to work right now.

I agree that this would be the cleanest solution. However, I fear that there are several other places where it is assumed that the figures are added directly to the root layer, which would then have to be adapted as well.

As a starting point, I can already say that one would would have to work with the [FreeformLayeredPane](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/FreeformLayeredPane.java) and [FreeformLayer](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/FreeformLayer.java) and then slowly clean up everything that's causing issues.

BauePhil:
< If I understand you correctly, you want to make the order of layers configurable, so that e.g. the connections can be show in front of the nodes (and vice versa)?

Yes exactly.

> As a starting point, I can already say that one would would have to work with the [FreeformLayeredPane](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/FreeformLayeredPane.java) and [FreeformLayer](https://github.com/eclipse/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/FreeformLayer.java) and then slowly clean up everything that's causing issues.

[...]

PR #532, opened 2024-08-22
[Zest 2.0] added getter for ZestRoorLayer ordering constants

BauePhil:
To be able to overwrite the ordering of the layers. See #518.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/internal/ZestRootLayer.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/internal/ZestRootLayer.java
@@ -64,37 +64,61 @@ public class ZestRootLayer extends FreeformLayer {
 	 */
 	private boolean isLayerKnown = false;
 
+	protected int getConnectionsLayer() {
+		return CONNECTIONS_LAYER;
+	}
+
+	protected int getSubgraphsLayer() {
+		return SUBGRAPHS_LAYER;
+	}
+
+	protected int getNodesLayer() {
+		return NODES_LAYER;
+	}
+
+	protected int getConnectionsHighlightedLayer() {
+		return CONNECTIONS_HIGHLIGHTED_LAYER;
+	}
+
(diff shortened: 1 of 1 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/internal/ZestRootLayer.java
```diff
@@ -64,37 +64,61 @@ public class ZestRootLayer extends FreeformLayer {
...
+	}
+
 	/**
 	 * Adds a node to the ZestRootLayer
 	 *
 	 * @param nodeFigure The figure representing the node
 	 */
 	public void addNode(IFigure nodeFigure) {
-		addFigure(nodeFigure, NODES_LAYER);
+		addFigure(nodeFigure, getNodesLayer());
```
ptziegler, line 97: I'll be honest; I don't like this at all. A `getNodesLayer()` method should always return `NODES_LAYER`. Anything else I would consider a conceptual flaw.

Can't you just overwrite the `addNode(IFigure)` method and then use a different constant? For example:
```java
public void addNode(IFigure nodeFigure) {
	addFigure(nodeFigure, CONNECTIONS_LAYER)
}
```

You might have to make the `changeFigureLayer()` method protected to overwrite the highlight methods, but that's fine. It still looks a little weird adding figures to other layers, but at least it's immediately obvious what's happening, when looking at it.

ptziegler (changes requested):

org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/internal/ZestRootLayer.java
BauePhil: > Can't you just overwrite the addNode(IFigure) method and then use a different constant?

Oh... yes that makes a lot more sense. I might have been a bit blinded by the constants, that I missed I don't even have to use them.
Thank you. I will close this PR.

PR #536, opened 2024-08-23
[Zest 2.0] Fix ambiguity in RotateGestureListener regarding connections

ptziegler:
The RotateGestureListener assumes that it is only possible to nodes (and containers) inside a Zest graph, but not connections. Trying to do so causes a ClassCastException.

To avoid this problem, we now explicitly check the currently selected items are all nodes and only perform the rotation if all of them are nodes.


ptziegler:
I'm not entirely sure how rotation is defined when the selection is "mixed", i.e. containing both nodes and connections. One idea was to allow the rotation of connections by rotating their source and destination nodes but how would it then behave if the connection and e.g. the source node is selected?

So in the end, I simply consider this an illegal state where rotation is not possible.

azoitl (approved):
For these kind of problems a mixed selection is always hard. I think your solution is consistent and therefore I like it.

Issue #531, opened 2024-08-22
ToolTipHelper doesn't work properly on Linux


ptziegler:
The tooltip is locked to the first screen it was showed in. Even if the figures are moved to a different monitor, the tooltip is still shown on the original monitor.

To reproduce:

- Run Zest GraphSnippet13 
- Expand USA container
- Hover over "Chris A." until the tooltip appears.
- Move the shell to a different monitor.
- Hover over "Chris A." again

Outcome:
- The tooltip is shown next to the figure as if it still were on the original monitor.

Expectation:
- The tooltip should appear right next to the figure on the new monitor.

ptziegler:
Idea: Dispose the shell after the popup is hidden again, rather than simply setting it invisible.

https://github.com/eclipse/gef-classic/blob/b1f6f28bafbb32601b821642164fdc6b3a0ea2ea/org.eclipse.draw2d/src/org/eclipse/draw2d/PopUpHelper.java#L152-L157

ptziegler:
Correction: Even if the shell is moved somewhere in the same monitor, the tooltip remains at the same location...

ptziegler:
Urgh... I remember now: At least on Linux, moving a shell is rarely, if at all done while invisible.

PR #537, opened 2024-08-24
Fix: ToolTipHelper doesn't work properly on Linux

ptziegler:
The pop-up helper reuses the shell that is used to display the tooltip. When a new tooltip is shown, the (currently invisible) shell is moved from its old location and then made visible. This doesn't work in GTK and the shell remains at its old location.

After the shell has been made visible, it is checked whether the shell has been correctly moved and if necessary, the shell coordinates are updated again. Note that it is still necessary to set the bounds while the shell is invisible, in order to properly handle the first tooltip that is shown.

Resolves https://github.com/eclipse/gef-classic/issues/531


ptziegler:
Just to illustrate the current behavior:

https://github.com/user-attachments/assets/22097833-a9b4-4108-8228-320c6b56843f

ptziegler:
Urgh... this approach breaks the GEF tooltips.

azoitl (approved):

PR #539, opened 2024-08-31
[Zest 2.0] Resolve remaining i18n warnings in Zest Core

ptziegler:
Classes such as the NodeSearchDialog use translatable strings while other classes which use them as internal values simply suppress the warning.


merks:
FYI, there are infrastructure problems:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4948

ptziegler:
I had a suspicion something was wrong... Thanks for the heads-up!

PR #540, opened 2024-08-31
[Zest 2.0] Create abstract base class for new Zest layout algorithms.

ptziegler:
Clients should not implement the LayoutAlgorithm interface directly as we may want to add new methods in the future (similar to the Zest 1.0 case). Instead, they should implement the AbstractLayoutAlgorithm class.


ptziegler:
The Zest1 "AbstractLayoutAlgorithm" class has been moved, hence the large diff. I think it's easier to see the changes when ignoring whitespace changes: https://github.com/eclipse/gef-classic/pull/540/files?w=1

azoitl (approved):

Issue #555, opened 2024-09-12
Tooltip not in correct position on Linux


Phillipus:
The following change has caused a regression on Linux (I tested on Ubuntu 24.04 with Wayland):

https://github.com/eclipse/gef-classic/commit/843e56c0df42aec34a487b97483a0b6c866fff95

Tooltips are displayed in the centre of the screen and grab the focus (see https://github.com/archimatetool/archi/issues/1076)

This will need a different approach.

ptziegler:
Based on the original bug description, I assume the problem shows up when doing something like this? I.e. displaying the tooltip before two nodes are connected?

[recording.webm](https://github.com/user-attachments/assets/d8a3d712-9772-451f-9877-5891ee22f639)

Can you give provide a small example and/or reproducer?

Phillipus:
Yes, that's right. But this is a side-effect.

The real issue is that on Linux Wayland the tooltip is displayed in the centre of the screen, not over the object, and grabs the focus. This is because the Shell created in `PopupHelper` has the Display as its parent because of the change in https://github.com/eclipse/gef-classic/commit/843e56c0df42aec34a487b97483a0b6c866fff95 That change should be reverted and a new solution provided like [this one](https://github.com/eclipse/gef-classic/issues/469#issuecomment-2213563131).

ptziegler:
I've booted up my Ubuntu VM and now I can see it as well.

Phillipus:
I've tested the solution at https://github.com/eclipse/gef-classic/issues/469#issuecomment-2213563131 and that is working for Archi.

ptziegler:
I agree, that seems to be the most efficient solution. Do you want to create a PR?

```java
if(toolTipHelper == null || (toolTipHelper != null && toolTipHelper.getShell().isDisposed())) {
	toolTipHelper = new ToolTipHelper(control);
}
```

Just a minor remark: The second null check is redundant.

Phillipus:
https://github.com/eclipse/gef-classic/pull/556

PR #556, opened 2024-09-12
Fix regression in Tooltip parent Shell on Linux Wayland

Phillipus:
This reverts 843e56c0df42aec34a487b97483a0b6c866fff95 and instead checks for a disposed Shell in SWTEventDispatcher#getToolTipHelper()

See https://github.com/eclipse/gef-classic/issues/555


ptziegler (approved):
Thanks! The Jenkins seems to be a little finicky right now. I'll merge it once everything is stable again.

merks:
repo.eclipse.org is having 403 problems again

PR #561, opened 2024-09-16
Use CopyOnWriteArrayList in LayoutNotifier

ptziegler:
See https://github.com/eclipse/gef/issues/120

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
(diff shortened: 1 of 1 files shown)


azoitl (approved):

ptziegler:
In my personal opinion, I believe that the initial handling of concurrency was simply a quirk of the implementation, rather than something that was done intentionally. Which is also why those regressions only popped up when the for-loops were converted to for-each loops, as this intent was never made explicitly clear.

Since the original change, we had four issues and seven PRs, all related to a ConcurrentModificationException. Which made it clear that users rely on this behavior and effectively elevated this quirk to a use-case.

> Do you think it is acceptable that listeners are (de)registered while notifying them? Or do you see the remaining need to correct the behavior of listeners to not concurrently (de)register listeners?

I believe that one can only decide on a case-by-case basis whether listeners should deregister themselves on an event. But I do believe that there are situations where this makes sense. For example from https://github.com/eclipse-gef/gef-classic/issues/663:

> Selection listeners that were removing themselves after selection is changed cause now exception

I.e. when an edit part is selected, it registers a SelectionChangedListener is added and removed again, once a different edit part is selected. Latter happens while processing a SelectionChangedEvent and thus requires synchronization.

Such a behavior is also not unprecedented, because e.g. the following will work just fine in SWT:
```java
Listener s = new Listener() {
    @Override
    public void handleEvent(Event event) {
        event.widget.removeListener(event.type, this);
    }
};
```

So to sum it up: I think that switching to a concurrent list is the proper, fix because it replicates the behavior of the previous implementation, which is something that has proven itself over two decades.

PR #563, opened 2024-09-17
Update changelog and readme regarding Zest 2.0

ptziegler:


azoitl:
Should we also put the short summary of the changes and what users need to do as you wrote in your very nice email to cross projects into the changelog?

ptziegler:
I suppose that makes sense. Though it'll also have to wait until tomorrow, because this final sprint really left me drained 😅 .

ptziegler:
I've added a few words regarding the deprecated LayoutAlgorithm interface and listed the relevant, deprecated features.

azoitl (approved):

azoitl:
Thx I think this is a very nice overview.

PR #565, opened 2024-09-19
Fix bogus calculation of graph node location

ptziegler:
The location of the graph node should match whatever has been calculated in the layout algorithm and should not include the node size.


ptziegler:
@merks can you try out the Analyzer Editor again, once this has been merged?

merks:
> @merks can you try out the Analyzer Editor again, once this has been merged?

I will update my TP from a nightly build when it's available and I test again tomorrow.  Thanks so much!!

PR #570, opened 2024-09-27
[Zest 2.0] Ignore node size when calculating position in layout

ptziegler:
The location of a node describes its top-left pixel, rather than its center. When using layout algorithms that are based on the grid layout, this causes the offset to be half the space between the rows and columns plus the half the width and height of the nodes.
Latter is undesirable and may cause nodes to be partially moved outside the client area.

The same issue can be observed in the other layout algorithms (Spring/Tree/Radial) when fitting the nodes into the client area, because the same mistake is repeated in the AlgorithmHelper class.


ptziegler:
I noticed this while migrating the Zest layout example. This is what the layout currently looks like:

image

Even though it's supposed to look like:

image

azoitl (approved):

PR #571, opened 2024-09-27
[Zest 2.0] Refactor Layouts example to Zest 2.0 API

ptziegler:
This adapts the SimpleNode, SimpleConnection and SimpleGraph classes of the example project to implement the NodeLayout, ConnectionLayout and LayoutContext interface, respectively, replacing the old Zest 1.0 interfaces.

Those classes are used by the SimpleSwingExample and SimpleSWTExample snippets to layout randomly generated graphs. Both examples have been updated to support the new layouts. Most noticeably, the nodes and connections are now exclusively stored in the layout context.

Furthermore, those classes have been cleaned up, including:
- Removal of empty stub methods in favor of using the adapter classes for listeners. Example: SelectionListener -> SelectionAdapter
- Layout reacts to changes to the shell/frame size.
- Typification of generic fields.
- Internationalization of UI-specific strings.


ptziegler:
image
image

azoitl (approved):
This is great as it can help our users to better understand how to use the new Zest2 api

PR #572, opened 2024-09-28
Use Typed Generics for Handle Generation and Management #155

azoitl:


ptziegler:
@azoitl Can you have a look at the merge conflict?

azoitl:
> @azoitl Can you have a look at the merge conflict?

Sure, I expected that the parallel work will result in some.

org.eclipse.gef/src/org/eclipse/gef/editpolicies/SelectionHandlesEditPolicy.java
```diff
@@ -51,38 +51,33 @@ protected void addSelectionHandles() {
 		removeSelectionHandles();
 		IFigure layer = getLayer(LayerConstants.HANDLE_LAYER);
 		handles = createSelectionHandles();
-		for (Object handle : handles) {
-			layer.add((IFigure) handle);
-		}
+		handles.stream().filter(IFigure.class::isInstance).map(IFigure.class::cast).forEach(layer::add);
```
ptziegler, line 54: Would it make sense to have a private helper function to get the Handle figures? The return type should then probably be a stream...

ptziegler (approved):

Issue #580, opened 2024-09-30
"IllegalStateException: Workbench has not been created yet." on loading PinFigure


cedric780:
Hello,
I am trying to migrate our product from old GEF (3.11 I think) to last gef-classic , and I have a "IllegalStateException: Workbench has not been created yet." error that prevent any GEF view to display.

Our product is a pure Eclipse 4 RCP application that nevers start the Eclipse 3 workbench.

```
java.lang.ExceptionInInitializerError: null
	at org.eclipse.gef.internal.ui.palette.editparts.PinFigure.<init>(PinFigure.java:45)
	at org.eclipse.gef.internal.ui.palette.editparts.DrawerFigure.<init>(DrawerFigure.java:169)
	at org.eclipse.gef.internal.ui.palette.editparts.DrawerEditPart$1.<init>(DrawerEditPart.java:68)
	at org.eclipse.gef.internal.ui.palette.editparts.DrawerEditPart.createFigure(DrawerEditPart.java:68)
	at org.eclipse.gef.editparts.AbstractGraphicalEditPart.getFigure(AbstractGraphicalEditPart.java:473)
	at org.eclipse.gef.internal.ui.palette.editparts.DrawerEditPart.getFigure(DrawerEditPart.java:117)
	at org.eclipse.gef.internal.ui.palette.editparts.DrawerEditPart.getFigure(DrawerEditPart.java:1)
	at org.eclipse.gef.editparts.AbstractGraphicalEditPart.addChildVisual(AbstractGraphicalEditPart.java:210)
	at org.eclipse.gef.editparts.AbstractEditPart.addChild(AbstractEditPart.java:219)
        ... stripped out ...
	at org.eclipse.e4.ui.internal.workbench.PartServiceImpl.activate(PartServiceImpl.java:681)
	at org.eclipse.e4.ui.internal.workbench.PartServiceImpl.showPart(PartServiceImpl.java:1262)
	at org.modelio.platform.rcp.inputpart.InputPartService.showInputPart(InputPartService.java:155)
	at org.modelio.diagram.editor.plugin.DiagramEditorsManager.lambda$0(DiagramEditorsManager.java:134)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4047)
[...]

PR #581, opened 2024-09-30
Create "open folder" icon and use local copy

ptziegler:
This copies the "open folder" icon from org.eclipse.ui into the GEF plug-in, thus removing a strong dependency to the Eclipse 3 workbench. This prevents GEF from being used in a pure Eclipse 4 application.

Resolves https://github.com/eclipse/gef-classic/issues/580


azoitl (approved):

azoitl:
I hoped that we could reduce the number of pictures that we provide by GEF Classic and use images from the platform to get a more consistent look and feel. But as #580 has shown this is more complicated then expected. And this is the way to go.

ptziegler:
As far as I understand, Eclipse 4 doesn't have a concept of "shared" images, so clients have to implement everything on their own. And even the shared images that _are_ provided by the platform are weird... Why is there a shared image for an open folder but not a closed folder?

azoitl:
And I would expect that people using a plain E4 application to even customize more then a typically E3 user.

PR #585, opened 2024-10-06
Fixed most warnings in text example.

azoitl:


org.eclipse.gef.examples.text/src/org/eclipse/gef/examples/text/model/commands/CompoundTextCommand.java
```diff
@@ -78,8 +76,7 @@ public void execute() {
 
 	@Override
 	public void executePending() {
-		for (int i = 0; i < pending.size(); i++) {
-			Command cmd = (Command) pending.get(i);
+		for (Command cmd : applied) {
```
ptziegler, line 79:
```suggestion
		for (Command cmd : pending) {
```

ptziegler: I think history has shown that it's usually not a good idea to convert for loops into for-each loop without checking whether it's impossible for a ConcurrentModifcationException to occur.
Would it make sense to use CopyOnWriteArrayLists for something as fundamental as a compound command?

azoitl: Ah just learned that I could have applied your suggested change directly here. Did it locally and force pushed.

Regarding your for-each loop. Yes you are right. However the command that we are talking here about is a command which is currently only within the text example. Furthermore it is not even used there. So I think it is fine to have it like that. Furthermore the text example would anyhow need some love as some of its features are currently broken.

azoitl: But that is some work for another day.

ptziegler: That's fine by me. It also looked good to me as well so I think we can go with how it is right now.

PR #589, opened 2024-10-14
Make drawer and tool entry figures configurable via PaletteCustomizer

ptziegler:
This adds the public interfaces IDrawerFigure and IToolEntryFigure, which are implemented by the figures created in the DrawerEditPart and ToolEntryEditPart, respectively.

Those interfaces can be accessed with with configure methods in the PaletteCustomizers. Clients can use them to adjust the figures without having to re-implement the palette editpart-factory and without having to reference the internal classes.


ptziegler:
For context: I'm currently in the progress of migrating our palette implementation to use the PaletteViewer. As part of this, I'd like to set our own color theme, so that the palette looks like this

image

and not like this

image

azoitl (changes requested):
I like this approach a lot. The only request I would have to add since tags for all interfaces and if not to much work also documentation for the new interface methods.

azoitl:
Oh just saw it is work in progress. Was to curios.

PR #590, opened 2024-10-14
Move painting of DrawerFigure background to custom border class.

ptziegler:
By moving the paint operation from the paintFigure() method out of the DrawerFigure class and instead doing so via the paintBackground() method of a custom implementation of the AbstractBackground class, clients can override the default background without having to extend the internal DrawerFigure class.


ptziegler:
Contributes to https://github.com/eclipse/gef-classic/pull/589

azoitl (approved):

Issue #569, opened 2024-09-25
[Linux] Overlay scrollbar obscures "pinned" icon


ptziegler:
I wonder if it makes sense to disable the overlay scrollbar by default...

![image](https://github.com/user-attachments/assets/12a0be04-00b9-42d1-bca9-70e4f644322d)

azoitl:
Would make sense for me. Even if would break a bit the look and feel of the target platform. But overlay scrollbars are anyhow an usability nightmare.

PR #592, opened 2024-10-15
Make overlay scrolling configurable via the palette settings

ptziegler:
This adds a new checkbox to the settings dialog that allows the user to replace the overlay scrollbar with the native scrollbar. Linux only, as the overlay is not available on any other platforms.

Resolves https://github.com/eclipse/gef-classic/issues/569


ptziegler:
Not really sure about adding it to the dialog, as the addition of static fields causes the API tools to act up...

image

ptziegler:
In order to avoid API breakage, I've created an internal subclass of the settings dialog. If people only use the PaletteSettingsDialog, they won't have any UI to configure it, but are free to implement it, given that the propery name is public.

image

azoitl (approved):
Dumb question: Should we provide not only a preference dialog but also a preference page that users could hook into the Eclipse Platform preference dialog? Or do you expect that applications needing that would do that by themselves in app specific ways?

ptziegler:
I think a preference page is too much, especially because we don't have any yet.

At the end of the day, this whole overlay issue is something that probably only very rarely. First you have to use Linux and then you also have to explicitly enable the vertical scrollbar. It can _also_ happen with the default configuration, but only when the editor is really small.

Because the global scrollbar is never set withing GEF, this is something that has happen in user code, but then the user can simply set `setScrollbarsMode()` on the `FigureCanvas`. This property is just a way to make it a little easier.

Hm... with a property page, it would also become a global property, so it would be very easy to enable it for palettes where you don't really want it. And for my use-case, I would want to use my own preference store, which would make things a lot more tedious... 😐

azoitl:
@ptziegler I didn't mean as part of this PR but for a new PR. However you bring quite some points that shows we shouldn't provide one even es reference in gef itself. It is maybe a better idea to have it in the Logic Editor example to serve as best practice for our uers.

PR #594, opened 2024-10-23
Internationalize CloudIO Example

ptziegler:


azoitl (commented):
I normally find it a better practice to have a single Message.java and message.properties per bundle. This simplifies the task of translating. Therefore I would propose to merge all of them into a single Message in the root package of the bundle.

ptziegler:
That's fine by me. Having one Message file per package is the default setting in the Externalize wizard, but I don't mind keeping everything in a single file.

PR #598, opened 2024-10-25
Make palette selection and hover colors customizable

ptziegler:
This defines a new ColorPalette interface that can be set in the PaletteViewer. The colors provided by this interface are then used in the palette figure for e.g. the tool entries.

Clients can extend the DefaultColorPalette to define their own colors and therefore change the look of the palette.

### DIFF
--- a/org.eclipse.gef.examples.shapes/fragment.e4xmi
+++ b/org.eclipse.gef.examples.shapes/fragment.e4xmi
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="ASCII"?>
+<fragment:ModelFragments xmi:version="2.0" xmlns:xmi="http://www.omg.org/XMI" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:commands="http://www.eclipse.org/ui/2010/UIModel/application/commands" xmlns:fragment="http://www.eclipse.org/ui/2010/UIModel/fragment" xmlns:menu="http://www.eclipse.org/ui/2010/UIModel/application/ui/menu" xmlns:ui="http://www.eclipse.org/ui/2010/UIModel/application/ui" xmi:id="_NgZIIJVIEe-9Ku3Ok7Aflw">
+  <imports xsi:type="menu:Menu" xmi:id="_XpVcIJVKEe-9Ku3Ok7Aflw" elementId="org.eclipse.ui.main.menu"/>
+  <fragments xsi:type="fragment:StringModelFragment" xmi:id="_OIiGMJVIEe-9Ku3Ok7Aflw" featurename="handlers" parentElementId="xpath:/">
+    <elements xsi:type="commands:Handler" xmi:id="_SvOkgJVIEe-9Ku3Ok7Aflw" elementId="org.eclipse.gef.examples.shapes.handler.paletteHandler" contributionURI="bundleclass://org.eclipse.gef.examples.shapes/org.eclipse.gef.examples.shapes.handler.ColorPaletteHandler" command="_WIwFEJVIEe-9Ku3Ok7Aflw"/>
+  </fragments>
+  <fragments xsi:type="fragment:StringModelFragment" xmi:id="_Ug_dEJVIEe-9Ku3Ok7Aflw" featurename="commands" parentElementId="xpath:/">
+    <elements xsi:type="commands:Command" xmi:id="_WIwFEJVIEe-9Ku3Ok7Aflw" elementId="org.eclipse.gef.examples.shapes.command.paletteCommand" commandName="Use Custom Color Palette"/>
+  </fragments>
+  <fragments xsi:type="fragment:StringModelFragment" xmi:id="_DGMksJVJEe-9Ku3Ok7Aflw" featurename="menuContributions" parentElementId="xpath:/">
+    <elements xsi:type="menu:MenuContribution" xmi:id="_FRJZUJVJEe-9Ku3Ok7Aflw" elementId="org.eclipse.gef.examples.shapes.menucontribution.paletteMenu" positionInParent="after=edit" parentId="org.eclipse.ui.main.menu">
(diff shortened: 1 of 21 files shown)


ptziegler:
Contributes to https://github.com/eclipse/gef-classic/pull/589. But with how complicated this is going to be, I think it's preferable to split it up into smaller chunks...

The current state allows you to customize the colors used for selection and hover, but not e.g. the background.
image

org.eclipse.gef/src/org/eclipse/gef/internal/ui/palette/editparts/DrawerFigure.java
```diff
@@ -143,8 +146,26 @@ public void paintBackground(IFigure figure, Graphics g, Insets insets) {
 	 *                used to display the drawer header with an EditPartTipHelper,
 	 *                if the header is not completely visible). It can be
 	 *                <code>null</code> (the tip won't be displayed).
+	 * @deprecated Use {@link #DrawerFigure(Control, ColorPalette)} instead to
+	 *             support custom color palettes.
 	 */
+	@Deprecated(forRemoval = true, since = "3.20")
 	public DrawerFigure(final Control control) {
```
ptziegler, line 153: I wonder whether it's a good or a bad idea to keep those old constructors. They're internal, so we're free to remove them without breaking API. But I know of at least one project that is using those internal classes which might expect this constructor...

ptziegler: Keeping technical dept for the sake of what-ifs is probably not a good idea... Especially for something that's very easy to adapt, if user were to use those internal classes.

azoitl: I totally agree here with you. We have enough technical dept already.

org.eclipse.gef.examples.shapes/src/org/eclipse/gef/examples/shapes/action/ColorPaletteAction.java
```diff
@@ -0,0 +1,51 @@
...
+import org.eclipse.gef.GraphicalViewer;
+import org.eclipse.gef.ui.actions.EditorPartAction;
+import org.eclipse.gef.ui.palette.PaletteViewer;
+
+import org.eclipse.gef.examples.shapes.palette.ShapesColorPalette;
+
+/**
+ * Flips between the custom and the default color palette whenever executed.
+ */
+public class ColorPaletteAction extends EditorPartAction {
```
azoitl, line 27: shouldn't we use for new code the command, handler, menu extension points?

ptziegler: You're probably right... But then I would go all the way and use the E4 mechanism. As much as I hate dealing with the e4xmi files...

This is how it looks like in the IDE:
image

ptziegler: Though it's probably not wrong to also have an E4 example...

org.eclipse.gef.examples.shapes/fragment.e4xmi
```diff
@@ -0,0 +1,18 @@
...
+  </fragments>
+  <fragments xsi:type="fragment:StringModelFragment" xmi:id="_DGMksJVJEe-9Ku3Ok7Aflw" featurename="menuContributions" parentElementId="xpath:/">
+    <elements xsi:type="menu:MenuContribution" xmi:id="_FRJZUJVJEe-9Ku3Ok7Aflw" elementId="org.eclipse.gef.examples.shapes.menucontribution.paletteMenu" positionInParent="after=edit" parentId="org.eclipse.ui.main.menu">
+      <children xsi:type="menu:Menu" xmi:id="_VaBhYJVJEe-9Ku3Ok7Aflw" elementId="org.eclipse.gef.examples.shapes.menu.palette" label="%menu.label.1">
+        <visibleWhen xsi:type="ui:ImperativeExpression" xmi:id="_5lzhoJVKEe-9Ku3Ok7Aflw" contributionURI="bundleclass://org.eclipse.gef.examples.shapes/org.eclipse.gef.examples.shapes.handler.ColorPaletteExpression"/>
+        <children xsi:type="menu:HandledMenuItem" xmi:id="_knZOAJVKEe-9Ku3Ok7Aflw" elementId="org.eclipse.gef.examples.shapes.handledmenuitem.usecustompalette" label="%handledmenuitem.label.1" enabled="false" type="Check" command="_WIwFEJVIEe-9Ku3Ok7Aflw"/>
+      </children>
+    </elements>
+  </fragments>
+</fragment:ModelFragments>
```
ptziegler, line 18: Which brings me to why I dislike the E4 model... With the plugin.xml, I can immediately see what's going on. Here I just see a big blob of XML. 🥲

azoitl: That is the reason I was suggesting command, handler, menu extension points as these are IMHO better readable.

ptziegler: @azoitl So what should we do now? The E4 approach still has commands and handlers, they are just stored in a different format. And I generally think it's good to have an example how GEF can be used in a pure E4 environment

azoitl: @ptziegler sorry was a bit busy since I came back from OCX. I find it really nice to have at list partly an E4 example. So I would go with it.

azoitl (approved):

org.eclipse.gef.examples.shapes/fragment.e4xmi
ptziegler: No worries, I'm more than grateful that you still find the time, despite your busy schedule 🙏

ptziegler:
I just noticed that this approach is very similar to what can already achieved with the [ColorProvider](https://github.com/eclipse-gef/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/ColorProvider.java).

I think it makes sense to replace the ColorPalette with the ColorProvider, so that I don't reinvent the wheel here.

PR #599, opened 2024-10-28
Migrate Draw2D JUnit tests from JUnit4 to JUnit5

ptziegler:


org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/swtbot/AbstractSWTBotTests.java
```diff
@@ -33,31 +34,39 @@
...
-			doTest(annotation, base);
+		public void interceptTestMethod(Invocation<Void> invocation,
+				ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext)
+				throws Throwable {
+			Object target = invocationContext.getTarget().orElse(null);
+			if (target instanceof AbstractSWTBotTests testObject) {
+				Snippet annotation = invocationContext.getExecutable().getAnnotation(Snippet.class);
+				Objects.requireNonNull(annotation, "Test is missing @Snippet annotation."); //$NON-NLS-1$
+				testObject.doTest(annotation, invocation);
+			}
```
ptziegler, line 67: There's probably a better way to do this but at the end of the day, I'm not overly familiar with the JUnit Extension API. In order to execute the SWTBot tests, we need to execute some additional code that creates and disposes the widget covered by the test...

azoitl: I'm a big fan of if it works and improves the existing situation. So I would go withit.

azoitl (approved):

PR #604, opened 2024-10-30
Use new repo location for the image link for the update site

merks:


merks:
This isn't tested by a verification build so merging now.

PR #607, opened 2024-11-02
Improve EventListenerList Performance Bugzilla #255534

azoitl:
By separating the event listener types into individual lists and using a map the startup performance could be greatly reduced. It is for small number of EditParts (i.e., 2000) already twice as fast as the previous implementation.
Furthermore iterating through the event listeners is simpler and most probably also faster.

The work is based on the patch submitted by Iwan Birrer to the following Bugzilla entry:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=255534

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
@@ -12,7 +12,12 @@
  *******************************************************************************/
 package org.eclipse.draw2d;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 /**
  * This class is intended for internal use only. TODO: If this is for internal
@@ -20,7 +25,8 @@ import java.util.Iterator;
  */
 public final class EventListenerList {
 
-	private volatile Object[] array;
+	private final Map<Class<?>, Deque<Object>> listeners = new HashMap<>();
+	private static final Deque<Object> EMPTY_DEQUE = new ArrayDeque<>();
 
 	/**
 	 * Adds a listener of type <i>c</i> to the list.
@@ -33,15 +39,7 @@ public final class EventListenerList {
 			throw new IllegalArgumentException();
 		}
 
-		int oldSize = (array == null) ? 0 : array.length;
-		Object[] newArray = new Object[oldSize + 2];
-		if (oldSize != 0) {
-			System.arraycopy(array, 0, newArray, 0, oldSize);
-		}
-		newArray[oldSize] = c;
-		oldSize++;
-		newArray[oldSize] = listener;
-		array = newArray;
+		listeners.computeIfAbsent(c, newC -> new ConcurrentLinkedDeque<>()).add(listener);
 	}
 
 	/**
@@ -52,50 +50,8 @@ public final class EventListenerList {
 	 * @return whether this list contains a listener of type <i>c</i>
 	 */
 	public synchronized <T> boolean containsListener(Class<T> c) {
-		if (array == null) {
-			return false;
-		}
-		for (int i = 0; i < array.length; i += 2) {
-			if (array[i] == c) {
-				return true;
-			}
-		}
-		return false;
-	}
-
-	static class TypeIterator<T> implements Iterator<T> {
-		private final Object[] items;
-		private final Class<T> type;
-		private int index;
-
-		TypeIterator(Object[] items, Class<T> type) {
-			this.items = items;
-			this.type = type;
-		}
-
-		@Override
-		public T next() {
-			@SuppressWarnings("unchecked") // check is performed in hasNext
-			T result = (T) items[index + 1];
-			index += 2;
-			return result;
-		}
-
-		@Override
-		public boolean hasNext() {
-			if (items == null) {
-				return false;
-			}
-			while (index < items.length && items[index] != type) {
-				index += 2;
-			}
-			return index < items.length;
-		}
-
-		@Override
-		public void remove() {
-			throw new UnsupportedOperationException("Iterator removal not supported"); //$NON-NLS-1$
-		}
+		Deque<Object> specListeners = listeners.get(c);
+		return specListeners != null && !specListeners.isEmpty();
 	}
 
 	/**
@@ -104,8 +60,9 @@ public final class EventListenerList {
 	 * @param listenerType the type
 	 * @return an Iterator of all the listeners of type <i>c</i>
 	 */
+	@SuppressWarnings("unchecked")
 	public synchronized <T> Iterator<T> getListeners(final Class<T> listenerType) {
-		return new TypeIterator<>(array, listenerType);
+		return (Iterator<T>) listeners.getOrDefault(listenerType, EMPTY_DEQUE).iterator();
 	}
 
 	/**
@@ -115,8 +72,9 @@ public final class EventListenerList {
 	 * @return an Iterable of all the listeners of type <i>c</i>
 	 * @since 3.13
 	 */
+	@SuppressWarnings("unchecked")
 	public synchronized <T> Iterable<T> getListenersIterable(final Class<T> listenerType) {
-		return () -> new TypeIterator<>(array, listenerType);
+		return (Iterable<T>) listeners.getOrDefault(listenerType, EMPTY_DEQUE);
 	}
 
 	/**
@@ -126,28 +84,17 @@ public final class EventListenerList {
 	 * @param listener the listener
 	 */
 	public synchronized <T> void removeListener(Class<T> c, Object listener) {
-		if (array == null || array.length == 0) {
-			return;
-		}
 		if (listener == null || c == null) {
 			throw new IllegalArgumentException();
 		}
 
-		int index = 0;
-		while (index < array.length) {
-			if (array[index] == c && array[index + 1] == listener) {
-				break;
+		Deque<Object> specListeners = listeners.get(c);
+		if (specListeners != null) {
+			specListeners.remove(listener);
+			if (specListeners.isEmpty()) {
+				listeners.remove(c);
 			}
-			index += 2;
 		}
-		if (index == array.length) {
-			return; // listener was not found
-		}
-
-		Object[] newArray = new Object[array.length - 2];
-		System.arraycopy(array, 0, newArray, 0, index);
-		System.arraycopy(array, index + 2, newArray, index, array.length - index - 2);
-		array = newArray;
 	}
 
 }
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
@@ -50,8 +50,7 @@ public final class EventListenerList {
 	 * @return whether this list contains a listener of type <i>c</i>
 	 */
 	public synchronized <T> boolean containsListener(Class<T> c) {
-		Deque<Object> specListeners = listeners.get(c);
-		return specListeners != null && !specListeners.isEmpty();
+		return listeners.containsKey(c);
 	}
 
 	/**


org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
```diff
@@ -33,15 +39,7 @@ public synchronized <T> void addListener(Class<T> c, Object listener) {
...
-		int oldSize = (array == null) ? 0 : array.length;
-		Object[] newArray = new Object[oldSize + 2];
-		if (oldSize != 0) {
-			System.arraycopy(array, 0, newArray, 0, oldSize);
-		}
-		newArray[oldSize] = c;
-		oldSize++;
-		newArray[oldSize] = listener;
-		array = newArray;
+		listeners.computeIfAbsent(c, newC -> new ArrayList<>(2)).add(listener);
```
ptziegler, line 42: That would probably have to be a thread-safe array, as indicated by the test failures. Though I wonder if this then removes the performance boost, as it now has to copy the internal array again...

```suggestion
		listeners.computeIfAbsent(c, newC -> new CopyOnWriteArrayList<>(2)).add(listener);
```

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
```diff
@@ -104,8 +59,9 @@ public void remove() {
 	 * @param listenerType the type
 	 * @return an Iterator of all the listeners of type <i>c</i>
 	 */
+	@SuppressWarnings("unchecked")
 	public synchronized <T> Iterator<T> getListeners(final Class<T> listenerType) {
-		return new TypeIterator<>(array, listenerType);
+		return (Iterator<T>) listeners.getOrDefault(listenerType, Collections.emptyList()).iterator();
```
ptziegler: Unlike before, this iterator now supports the "remove" operation. Is this desirable or something we would have to actively suppress?

ptziegler:
Do you also want to add the [snippet](https://bugs.eclipse.org/bugs/attachment.cgi?id=118061) that was attached to the Bugzilla item? I think it would be nice to have an easy way to compare the performance of the different implementations.

azoitl:
> Do you also want to add the [snippet](https://bugs.eclipse.org/bugs/attachment.cgi?id=118061) that was attached to the Bugzilla item? I think it would be nice to have an easy way to compare the performance of the different implementations.

I was not sure about the legal implications of adding this one.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
azoitl: Hmh. In the meantime I tried a version where I copy for the iterators. Which is a bit slower but modification save. The copy on write is interestingly faster until 10k of EditParts. Then my version of copying on access is faster. However copy on write is still faster then then the original version. So I guess we should use it or?

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
```diff
@@ -12,15 +12,21 @@
...
 
 /**
  * This class is intended for internal use only. TODO: If this is for internal
  * use only, we should move it to the internal package.
  */
 public final class EventListenerList {
 
-	private volatile Object[] array;
+	@SuppressWarnings("rawtypes")
+	private final Map<Class, List<Object>> listeners = new HashMap<>();
```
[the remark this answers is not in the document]
azoitl: no I just didn't think of it.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
azoitl: With the new design it wouldn't matter to much. I think it was just not allowed as in the old implementation it would have been extremly hard to implement.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
ptziegler: Based on the snippet in the bug, I got the following results:

**Current Implementation**

2001 edit parts: 71 ms
10001 edit parts: 387 ms
20001 edit parts: 903 ms
40001 edit parts: 2955 ms
80001 edit parts: 9300 ms
160001 edit parts: 35599 ms

**ArrayList with Copy Constructor**

2001 edit parts: 66 ms
10001 edit parts: 198 ms
20001 edit parts: 365 ms
40001 edit parts: 1079 ms
80001 edit parts: 3001 ms
160001 edit parts: 9328 ms

**CopyOnWriteArrayList**

2001 edit parts: 72 ms
10001 edit parts: 242 ms
20001 edit parts: 559 ms
40001 edit parts: 1582 ms
80001 edit parts: 4893 ms
160001 edit parts: 16947 ms

So the initialization is the fastest when using a plan ArrayList, which is what one would expect. However, what the test doesn't cover is the access to those listeners, which should now be noticeably slower as a new array is created every time. So effectively, the original problem has been solved by moving it somewhere else.

To be honest, I' not a huge fan of either approaches and I wonder if there is an alternative we simply haven't considered yet...

ptziegler:
> > Do you also want to add the [snippet](https://bugs.eclipse.org/bugs/attachment.cgi?id=118061) that was attached to the Bugzilla item? I think it would be nice to have an easy way to compare the performance of the different implementations.
> 
> I was not sure about the legal implications of adding this one.

Fair point. Though I still believe that a controlled test to measure the performance would help out a lot.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
```diff
@@ -52,50 +50,7 @@ public synchronized <T> void addListener(Class<T> c, Object listener) {
...
-				index += 2;
-			}
-			return index < items.length;
-		}
-
-		@Override
-		public void remove() {
-			throw new UnsupportedOperationException("Iterator removal not supported"); //$NON-NLS-1$
-		}
+		return listeners.containsKey(c);
```
ptziegler, line 53: After looking at this again, I don't think this check covers all cases.

Example:
If you add a listener for the first time, a new entry is added to the map. But if the listener is removed again, the entry is not removed. So this method returns `true`, even though the list of listeners is empty.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
ptziegler: Hm... what about using a Deque/ConcurrentLinkedDeque instead of a List/(CopyOn)ArrayList?

We have free control over what data structure those listeners are stored in and performance-wise, it's very close to a plain ArrayList while also giving us thread-safety.

2001 edit parts: 73 ms
10001 edit parts: 240 ms
20001 edit parts: 435 ms
40001 edit parts: 1170 ms
80001 edit parts: 3629 ms
160001 edit parts: 10177 ms

azoitl: Hm... I'm not so fond of using linked lists. They have quite a memory overhead and the `EventListenerList` is used for every `EditPart` and `Figure`.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
azoitl: You are right. Fixed that.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
ptziegler: What do you mean by "quite a memory overhead"? Based on the example with 160000 edit parts, I see the entire program taking ~600MB with the current implementation, 713MB when using an array list and 754MB when using a linked list, which to me is all in the same ballpark.
Regardless, if you think an array list is a better fit here, then that's fine by me.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
```diff
@@ -126,28 +83,14 @@ public synchronized <T> Iterable<T> getListenersIterable(final Class<T> listener
...
 
-		int index = 0;
-		while (index < array.length) {
-			if (array[index] == c && array[index + 1] == listener) {
-				break;
-			}
-			index += 2;
+		List<Object> specListeners = listeners.get(c);
+		if (specListeners != null) {
+			specListeners.remove(listener);
```
ptziegler, line 92: Something else I just remembered when using classes as keys... Would it make sense to remove the key-value pair from the map when the number of listeners is empty, in order to remove any strong references to said classes?

ptziegler:
Maybe to just summarize the open topics:

- I'm not a big fan of creating a copy whenever the getters are called, as it feels like it just shifts the startup delay. I'm fine with either an ArrayList or a LinkedList.

- If a concurrent list is used, we also need to decide how to handle "null" listeners, which are currently supported, even though I don't see a real use case for them...

azoitl:
> Maybe to just summarize the open topics:
> 
>     * I'm not a big fan of creating a copy whenever the getters are called, as it feels like it just shifts the startup delay. 

me neither.

> I'm fine with either an ArrayList or a LinkedList.

I only had concerns that a LinkedList normally uses at least one additional pointer (next) mostly two (next and prev) so we would have a 200% overhead compared to an ArrayList. But as your measurement proved differently I think we could go with the ConcurrentLinkedDeque or did I miss anything?

> 
>     * If a concurrent list is used, we also need to decide how to handle "null" listeners, which are currently supported, even though I don't see a real use case for them...

I'm not sure if I get you here. You mean removing the  map entry if the list is empty? If yes I would do so.

ptziegler:
> I only had concerns that a LinkedList normally uses at least one additional pointer (next) mostly two (next and prev) so we would have a 200% overhead compared to an ArrayList. But as your measurement proved differently I think we could go with the ConcurrentLinkedDeque or did I miss anything?

The linked list is still larger, I only wondered whether that's by a significant margin. At the end of the day, it's your PR so you're free to choose the approach you prefer. ;D

> I'm not sure if I get you here. You mean removing the map entry if the list is empty? If yes I would do so.

That's an oversight on my part. I though that you could call `addListener(..., null)` without error, but I missed the check at the start of the method.

Tough I still believe that removing the map entry when the list of listeners is empty to be a good idea. Even if it just prevents any hard reference to the classes and thus potentially blocking class loaders from being garbage collected. I experienced a similar issues a few months ago with the platform, which made me cautions whenever I see something related to classes.

azoitl:
I did some further reading and and some more experiments. The Deque/ConcurrentLinkedDeque is faster then the CopyOnWriteArrayList. Also when I used this approach in 4diac IDE with larger diagrams and many editors (app. 100 editors with large diagrams) i didn't see a major increase in memory usage. So I would go with the Deque/ConcurrentLinkedDeque  approach for now.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
```diff
@@ -52,50 +50,8 @@ public synchronized <T> void addListener(Class<T> c, Object listener) {
...
-			}
-			return index < items.length;
-		}
-
-		@Override
-		public void remove() {
-			throw new UnsupportedOperationException("Iterator removal not supported"); //$NON-NLS-1$
-		}
+		Deque<Object> specListeners = listeners.get(c);
+		return specListeners != null && !specListeners.isEmpty();
```
ptziegler, line 54: Now that the Deque object is removed together with the last listener, I suppose you can go back to this simpler check, right?
```suggestion
		return listeners.containsKey(c);
```

ptziegler (approved):
Also looking good, outside of a very small optimization.

org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
azoitl: good point.

ptziegler:
> I did some further reading and and some more experiments. The Deque/ConcurrentLinkedDeque is faster then the CopyOnWriteArrayList. Also when I used this approach in 4diac IDE with larger diagrams and many editors (app. 100 editors with large diagrams) i didn't see a major increase in memory usage. So I would go with the Deque/ConcurrentLinkedDeque approach for now.

That's good to hear! I also apologize for being really picky about this class. 😅 I just always fear that a lot of stuff can easily go wrong with this kind of optimizations.

azoitl:
> > I did some further reading and and some more experiments. The Deque/ConcurrentLinkedDeque is faster then the CopyOnWriteArrayList. Also when I used this approach in 4diac IDE with larger diagrams and many editors (app. 100 editors with large diagrams) i didn't see a major increase in memory usage. So I would go with the Deque/ConcurrentLinkedDeque approach for now.
> 
> That's good to hear! I also apologize for being really picky about this class. 😅 I just always fear that a lot of stuff can easily go wrong with this kind of optimizations.

I'm very happy that you are here so strict on me. This is a very central class breaking it may have rather large consequences.

PR #615, opened 2024-11-12
Switch to CopyOnWriteArrayList in EventListenerList

ptziegler:
The ConcurrentLinkedDeque might lead to a deadlock in a multi-thread environment. This doesn't happen with a CopyOnWriteArrayList, as each write operation is done on a separate array.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/EventListenerList.java
@@ -12,12 +12,12 @@
  *******************************************************************************/
 package org.eclipse.draw2d;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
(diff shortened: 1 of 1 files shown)


ptziegler:
@azoitl Using a CopyOnWriteArrayList was still faster than the plain array, right? I'm more than happy to look deeper into this at a later stage, but this issue effectively blocks the WindowBuilder M3.

azoitl:
> @azoitl Using a CopyOnWriteArrayList was still faster than the plain array, right? I'm more than happy to look deeper into this at a later stage, but this issue quite literally blocks the WindowBuilder M3.

Yes, and it is even faster then exponentially growing a plain array, or using a record with two entries in an ArrayList. Tried a few things. I think now the most performance is coming from the map on the listener type.

azoitl (approved):

PR #617, opened 2024-11-15
Restore ability to modify selection in GroupRequest within DeleteAction

ptziegler:
The list returned by getSelectedEditParts() should be a view of the internal selection, similar to getSelectedObjects(). Meaning callers are able to modify by the selection by calling set(int,Object).

In order to reduce the chance of an undesirable ClassCastException, we only check whether the first element in the list is an EditPart, assuming that all remaining elements are as well. I.e. we don't expect a "mixed" selection of different types. This is similar to what has already been done in e.g. the DeleteAction or the MatchSizeAction.

This is a follow-up to 78c3fed3ced2c035c5a495a752523bb4b758a916.

Resolves https://github.com/eclipse-gef/gef-classic/issues/616

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
@@ -65,11 +65,14 @@ public abstract class SelectionAction extends WorkbenchPartAction {
 	}
 
 	/**
-	 * Returns a <code>List</code> containing the currently selected objects.
+	 * Returns a {@code List<EditPart} containing the currently selected objects.
+	 * The {@code set} method is supported but the {@code add}, {@code addAll},
+	 * {@code remove} {@code removeAll} methods are not. Changes to this list modify
+	 * the internal {@link #selection}.
 	 *
 	 * @return A List containing the currently selected objects.
 	 */
-	protected List<Object> getSelectedObjects() {
+	protected List<?> getSelectedObjects() {
 		if (!(getSelection() instanceof IStructuredSelection)) {
 			return Collections.emptyList();
 		}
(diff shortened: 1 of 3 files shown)


ptziegler:
The lazy check with objects.get(0) I mentioned:

https://github.com/eclipse-gef/gef-classic/blob/4660b69397dbfb6823d0f1a58092cf58734ff75b/org.eclipse.gef/src/org/eclipse/gef/ui/actions/DeleteAction.java#L93-L115

https://github.com/eclipse-gef/gef-classic/blob/4660b69397dbfb6823d0f1a58092cf58734ff75b/org.eclipse.gef/src/org/eclipse/gef/ui/actions/MatchSizeAction.java#L72-L119

azoitl (approved):

org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
```diff
@@ -65,26 +65,41 @@ public void dispose() {
...
-	 * Returns a <code>List</code> containing the currently selected objects.
+	 * Returns a {@code List<EditPart} containing the currently selected objects.
+	 * The {@code set} method is supported but the {@code add}, {@code addAll},
+	 * {@code remove} {@code removeAll} methods are not. Changes to this list modify
+	 * the internal {@link #selection}.
 	 *
 	 * @return A List containing the currently selected objects.
 	 */
-	protected List<Object> getSelectedObjects() {
+	protected List<?> getSelectedObjects() {
```
merks, line 75: This is an API breaking change and break BIRT's build:

https://github.com/eclipse-birt/birt/pull/1972#issuecomment-2482704548

Please revert it.

PR #621, opened 2024-11-17
Update to latest SonarQube scanner and exclude generated poms

ptziegler:
The SonarQube version that is used by Maven has been relocated, leading to a warning at the start of the build. Furthermore, the "consumer poms" that are automatically created by Tycho are flagged as "missing blame information", as they haven't been comitted and should thus not be scanned.


ptziegler:
:D
image

azoitl (approved):

azoitl:
Nice

PR #622, opened 2024-11-18
Restore the return type of SelectionAction.getSelectedObjects()

merks:
- The return type cannot be changed from List<Object>  to List<?> without breaking client calling or override this protected method.

https://github.com/eclipse-gef/gef-classic/pull/617

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
@@ -72,7 +72,8 @@ public abstract class SelectionAction extends WorkbenchPartAction {
 	 *
 	 * @return A List containing the currently selected objects.
 	 */
-	protected List<?> getSelectedObjects() {
+	@SuppressWarnings("unchecked")
+	protected List<Object> getSelectedObjects() {
 		if (!(getSelection() instanceof IStructuredSelection)) {
 			return Collections.emptyList();
 		}


merks:
Guys, I'd like to undo this breakage as soon as possible to restore BIRT's builds to a working state:

https://github.com/eclipse-birt/birt/pull/1972#issuecomment-2482753642

org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
```diff
@@ -72,7 +72,8 @@ public void dispose() {
 	 *
 	 * @return A List containing the currently selected objects.
 	 */
-	protected List<?> getSelectedObjects() {
+	@SuppressWarnings("unchecked")
+	protected List<Object> getSelectedObjects() {
```
ptziegler, line 76: The main motivation was to resolve a compilation error in ``getSelectedEditParts()` because of a cast from List<Object> to List<EditPart>. If it still compiles, it's fine to go back to this.

ptziegler (approved):

org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java
merks: Thanks for the quick turnaround!  🥇

wimjongman: 💞

merks:
Yes, and I suppose that's the same question for `org.eclipse.jface.viewers.IStructuredSelection.toList()` where the answer appears to be "do no harm" or "if it isn't broken don't fix it". 

I suppose part of the question is also, is the change binary compatible though not source compatible?  I didn't actually look at the history to see what states this as been in historically.  It appears to have first been changed here:

image

Changing this now to something else for RC1 is not so ideal for anyone who might have tried to accommodate the previous change...

I suppose there is no perfect answer.  😢

ptziegler:
> I suppose part of the question is also, is the change binary compatible though not source compatible?

As far as I remember, generic type information is not preserved in byte code, so from the perspective of the JVM, the return type was changed from `List` to `List`. So that would indeed be the case.

> Changing this now to something else for RC1 is not so ideal for anyone who might have tried to accommodate the previous change...

I agree, that might not've been the smartest decision...

merks:
No good deed goes unpunished.  😨

PR #623, opened 2024-11-18
Publish build results from Jenkins to GitHub

ptziegler:
Main goal is to analyze what steps need to be taken to have the Jenkins build results like compiler warnings and Maven problems be published to the GitHub checks.


ptziegler:
Hm... The upload is done via the github-checks plugin, while the diff with the reference build is done by the git-forensics plugin. I assume both need to be installed in our Jenkins instance.

ptziegler:
@azoitl wdyt? For the next release, I'd like to make the quality checks a little harsher so that e.g. builds fail if they introduce new warnings. To make things a little bit more user-friendly, those warnings should therefore be visible on GitHub as well.

image

PR #630, opened 2024-12-17
Fixed Examples build.properties

azoitl:
The build.properties src.inlucde entry had more entries then necessary. This was cleaned

Furthermore several warnings and issues reported by sonar have been fixed in the example creation wizard.

### DIFF
--- a/org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
+++ b/org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
@@ -204,7 +204,6 @@ public abstract class ProjectUnzipperNewWizard extends Wizard implements INewWiz
 									new String[] { ORG_ECLIPSE_PDE_PLUGIN_NATURE, ORG_ECLIPSE_JDT_CORE_JAVANATURE });
 							project.setDescription(desc, subMonitor.newChild(10));
 
-							subMonitor.worked(10);
 							if (subMonitor.isCanceled()) {
 								throw new InterruptedException();
 							}
--- a/org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
+++ b/org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
@@ -14,7 +14,6 @@
 package org.eclipse.gef.examples.ui.pde.internal.wizards;
 
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -22,6 +21,7 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.util.Enumeration;
 import java.util.zip.ZipEntry;
@@ -41,7 +41,7 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Status;
-import org.eclipse.core.runtime.SubProgressMonitor;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.jface.operation.IRunnableWithProgress;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.wizard.Wizard;
@@ -63,7 +63,7 @@ import org.eclipse.gef.examples.ui.pde.internal.l10n.Messages;
  *
  * @see Wizard
  */
-abstract public class ProjectUnzipperNewWizard extends Wizard implements INewWizard, IExecutableExtension {
+public abstract class ProjectUnzipperNewWizard extends Wizard implements INewWizard, IExecutableExtension {
 
 	/**
 	 * Java Nature
@@ -84,32 +84,32 @@ abstract public class ProjectUnzipperNewWizard extends Wizard implements INewWiz
 	/**
 	 * The name of the project creation page
 	 */
-	private String pageName;
+	private final String pageName;
 
 	/**
 	 * The title of the project creation page
 	 */
-	private String pageTitle;
+	private final String pageTitle;
 
 	/**
 	 * The description of the project creation page
 	 */
-	private String pageDescription;
+	private final String pageDescription;
 
 	/**
 	 * The name of the project in the project creation page
 	 */
-	private String pageProjectName;
+	private final String pageProjectName;
 
 	/**
 	 * The list of paths pointing to the location of the project archives
 	 */
-	private URL[] projectZipURL;
+	private final URL[] projectZipURL;
 
 	/**
 	 * The list of formats to be applied to the user supplied name
 	 */
-	private String[] nameFormats;
+	private final String[] nameFormats;
 
 	/**
 	 * The configuration element associated with this new project wizard
@@ -126,7 +126,7 @@ abstract public class ProjectUnzipperNewWizard extends Wizard implements INewWiz
 	 * @param projectZipURLIn   The URL pointing to the location of the project
 	 *                          archive
 	 */
-	public ProjectUnzipperNewWizard(String pageNameIn, String pageTitleIn, String pageDescriptionIn,
+	protected ProjectUnzipperNewWizard(String pageNameIn, String pageTitleIn, String pageDescriptionIn,
 			String pageProjectNameIn, URL projectZipURLIn) {
 		this(pageNameIn, pageTitleIn, pageDescriptionIn, pageProjectNameIn, new URL[] { projectZipURLIn },
 				new String[] { "{0}" }); //$NON-NLS-1$
@@ -141,7 +141,7 @@ abstract public class ProjectUnzipperNewWizard extends Wizard implements INewWiz
 	 * @param pageProjectNameIn   The project name in the project creation page
 	 * @param projectZipURLListIn The list of URL pointing to the location of the
 	 *                            project archives
-	 * @param formatsIn           The list of formats to be applied to the user
+	 * @param nameFormatsIn       The list of formats to be applied to the user
 	 *                            supplied name. The {@link java.text.MessageFormat}
 	 *                            class should be consulted to understand
 	 *                            substitutions. The &quot;{0}&quot; substitution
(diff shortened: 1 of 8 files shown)


org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
```diff
@@ -208,7 +202,7 @@ public void execute(IProgressMonitor monitor) throws InterruptedException {
 								 * Copy plug-in project code
 								 */
 								extractProject(projectFolderFile, getProjectZipURL()[i],
-										new SubProgressMonitor(monitor, 100));
+										SubMonitor.convert(monitor, 100));
```
ptziegler, line 205: This is a really common pattern I've seen countless times by now, which sadly breaks the IProgressMonitor contract. From the JavaDoc:

```
Since it is illegal to call beginTask on the same IProgressMonitor more than once, the same instance of IProgressMonitor must not be passed to convert more than once.
```

Instead I would go with something like this:
```java

public boolean performFinish() {

    try {
        IRunnableWithProgress operation = new WorkspaceModifyOperation() {

            @Override
            public void execute(IProgressMonitor monitor) throws InterruptedException {
                try {
                    SubMonitor subMonitor = SubMonitor.convert(monitor, Messages.monitor_creatingProject, 150);

                    /*
                        * Create the project folder
                        */
                    IPath projectPath = wizardNewProjectCreationPage.getLocationPath();

                    for (int i = 0; i < nameFormats.length; i++) {
                        String projectName = MessageFormat.format(nameFormats[i],
                                wizardNewProjectCreationPage.getProjectName());
                        String projectFolder = projectPath.toOSString() + File.separator + projectName;
                        File projectFolderFile = new File(projectFolder);

                        IWorkspace workspace = ResourcesPlugin.getWorkspace();
                        IProject project = workspace.getRoot().getProject(projectName);

                        // If the project does not exist, we will create it
                        // and populate it.
                        if (!project.exists()) {
                            projectFolderFile.mkdirs();
                            subMonitor.worked(10);

                            /*
                                * Copy plug-in project code
                                */
                            extractProject(projectFolderFile, getProjectZipURL()[i], subMonitor.newChild(100));

                            if (subMonitor.isCanceled()) {
                                throw new InterruptedException();
                            }

                            if (projectPath.equals(workspace.getRoot().getLocation())) {
                                project.create(subMonitor.newChild(10));
                            } else {
                                IProjectDescription desc = workspace.newProjectDescription(project.getName());
                                desc.setLocation(new Path(projectFolder));

                                project.create(desc, subMonitor.newChild(10));
                            }
                        }

                        // Now, we ensure that the project is open.
                        project.open(subMonitor.newChild(10));

                        renameProject(project, projectName);

                        // Add Java and PDE natures
                        IProjectDescription desc = workspace.newProjectDescription(project.getName());
                        desc.setNatureIds(
                                new String[] { ORG_ECLIPSE_PDE_PLUGIN_NATURE, ORG_ECLIPSE_JDT_CORE_JAVANATURE });
                        project.setDescription(desc, subMonitor.newChild(10));

                        subMonitor.worked(10);
                        if (subMonitor.isCanceled()) {
                            throw new InterruptedException();
                        }
                    } // end for

                } catch (IOException e) {
                    throw new RuntimeException(e);
                } catch (CoreException e) {
                    throw new RuntimeException(e);
                }
            }
        };

        getContainer().run(false, true, operation);

        /* Set perspective */
        BasicNewProjectResourceWizard.updatePerspective(getConfigurationElement());

    } catch (InterruptedException e) {
        return false;

    } catch (Exception e) {

        GefExamplesPlugin.getDefault().getLog().log(new Status(IStatus.ERROR, "org.eclipse.gef.examples.ui.pde", //$NON-NLS-1$
                IStatus.OK, e.getMessage(), e));
        return false;
    }

    return true;
}
```

It's also a bad idea to pass the same monitor to e.g. `project.open()`, because it may internally create a SubMonitor (or even just call `done()`). Meaning you also have to create a sub-monitor for each of those calls.

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
```diff
@@ -317,36 +303,33 @@ private void extractProject(File projectFolderFile, URL url, IProgressMonitor mo
...
-					InputStreamReader is = null;
-					OutputStreamWriter os = null;
-
-					try {
-						is = new InputStreamReader(zipFile.getInputStream(zipEntry), "ISO-8859-1"); //$NON-NLS-1$
-						os = new OutputStreamWriter(new FileOutputStream(file), ResourcesPlugin.getEncoding());
+					try (InputStreamReader is = new InputStreamReader(zipFile.getInputStream(zipEntry),
+							StandardCharsets.ISO_8859_1);
+							OutputStreamWriter os = new OutputStreamWriter(new FileOutputStream(file),
+									ResourcesPlugin.getEncoding());) {
```
ptziegler, line 332: I think Windows-1252 is not really an appropriate encoding anymore. Given that this is an example, I would propose to use UTF-8 instead.

```suggestion
					try (InputStreamReader is = new InputStreamReader(zipFile.getInputStream(zipEntry),
							StandardCharsets.UTF_8);
							OutputStreamWriter os = new OutputStreamWriter(new FileOutputStream(file),
									ResourcesPlugin.getEncoding());) {
```

ptziegler (changes requested):

ptziegler:
As a side note: The code snippet:
```java
byte[] buffer = new byte[102400];
while (true) {
    int len = is.read(buffer);
    if (len < 0) {
        break;
    }
    os.write(buffer, 0, len);
}
```

can be simplified to `is.transferTo(os)`.

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
azoitl: Thx for pointing that us. This is important that I learned that. Is the last comment also the reason you removed the monitor.done in the finally?

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
azoitl: you are totally right. Changed that accordingly.

azoitl:
> As a side note: The code snippet:
> 
> ```java
> byte[] buffer = new byte[102400];
> while (true) {
>     int len = is.read(buffer);
>     if (len < 0) {
>         break;
>     }
>     os.write(buffer, 0, len);
> }
> ```
> 
> can be simplified to `is.transferTo(os)`.

changed that as well.

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
ptziegler: > This is important that I learned that. Is the last comment also the reason you removed the monitor.done in the finally?

Partially, yes. Calling `done()` is generally the responsibility of whoever created the IProgressMonitor instance. So this is something that should be handled by the caller, not the callee.

The SubMonitor API specifically says that you don't need to call `done()` on the created SubMonitor. And there is also no need, given that all of the allocated ticks are consumed by the time the method finishes.

azoitl: Thx for the clarification. This is very helpful.

azoitl: forgot to mention. Changed everything as you suggested and did some more clean-ups. Also tested it in a dedicated Eclipse instance.

ptziegler: To be honest, most of my knowledge comes from the [Using Progress Monitor](https://www.eclipse.org/articles/Article-Progress-Monitors/article.html) article, as well as trial and error 😅 

I think this is the interesting part:
# 4.0 Responsibilities of callers and callees
Imagine you need to invoke another method that accepts an IProgressMonitor. What are the responsibilities of the caller and what are the responsibilities of the callee?

The caller:

- Will pass an IProgressMonitor instance which has not had beginTask invoked on it yet (or an implementation such as SubMonitor which permits beginTask to be invoked multiple times).
- Will not expect that the callee invokes done() on the monitor. The caller must either use an SubMonitor (or a similar implementation which does not require done() to be invoked), or it must take responsibility for calling done() on the monitor after the callee has finished.
- Will not pass in a null monitor unless the JavaDoc for the callee says that it accepts null.
- Will pass in a monitor which ignores the String argument to beginTask unless the JavaDoc for the callee says that it requires otherwise.

The callee:

- Will call beginTask 0 or 1 times on the monitor, at its option.
- Will not promise to call done() on the monitor, although it is allowed to do so.
- Will not call setCanceled on the monitor.
- Will not accept a null monitor unless its JavaDoc says otherwise.
- Will not expect the caller to do anything with the string passed to beginTask unless its JavaDoc says otherwise.

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
```diff
@@ -182,61 +176,36 @@ public boolean performFinish() {
 				@Override
 				public void execute(IProgressMonitor monitor) throws InterruptedException {
 					try {
-						monitor.beginTask(Messages.monitor_creatingProject, 120);
+						SubMonitor subMonitor = SubMonitor.convert(monitor, Messages.monitor_creatingProject, 150);
```
ptziegler, line 179: I just noticed that this was also wrong in the original implementation... The ticks are consumed within a for-loop, so 150 (140 without line 207) are only enough if there is a single format.
```suggestion
						SubMonitor subMonitor = SubMonitor.convert(monitor, Messages.monitor_creatingProject, 140 * nameFormats.length);
```

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
```diff
@@ -182,61 +176,36 @@ public boolean performFinish() {
...
 							// Add Java and PDE natures
 							IProjectDescription desc = workspace.newProjectDescription(project.getName());
 							desc.setNatureIds(
 									new String[] { ORG_ECLIPSE_PDE_PLUGIN_NATURE, ORG_ECLIPSE_JDT_CORE_JAVANATURE });
-							project.setDescription(desc, monitor);
+							project.setDescription(desc, subMonitor.newChild(10));
 
-							monitor.worked(10);
-							if (monitor.isCanceled()) {
+							subMonitor.worked(10);
```
ptziegler, line 207: Now I wonder if it is really necessary to consume 10 ticks for "nothing", especially when 10 ticks were just consumed while setting the description.
```suggestion
```

org.eclipse.gef.examples.ui.pde/src/org/eclipse/gef/examples/ui/pde/internal/wizards/ProjectUnzipperNewWizard.java
azoitl: Guess I should read that in more detail.

ptziegler (approved):

PR #631, opened 2024-12-17
Ignored warning of tests for deprecated methods

azoitl:
As long as the deprecated methods are not removed the tests should stay. However the warning is only distracting from other problems therefore it is removed.


ptziegler (approved):

ptziegler:
The build failure after the merge is unrelated. See https://github.com/eclipse-gef/gef-classic/pull/632

PR #632, opened 2024-12-20
Use Java 21 for Jenkins job/GitHub workflow actions

ptziegler:
The Eclipse Help requires Java 21 due to a dependency to Lucene 10, meaning that our builds also need to use at least Java 21.

The BREE for the test fragments has been raised, as they launch an Eclipse IDE and therefore won't work with older Java versions.

In order to avoid regressions due to accidentally using newer API form the JRE, a toolchain has been configured to compile the remaining bundles that still support Java 17 with both Java 17 and 21.


ptziegler:
My main concern with using Java 21 is that we have no safety mechanism in case we accidentally use e.g. the SequencedCollection in one of our plugins. If we want to supporting Java 17, a proper toolchain is the bare minimum.

azoitl (approved):
I would go forward with your proposal. With that we have at least running tests again and contributors can provide PRs and get them checked. 

Not yet sure what it will mean for GEF.

ptziegler:
The minimum BREE has only been changed for the tests so right now nothing has changed. But as you said, the main motivation right now is to get the build pipeline functional again.

ptziegler:
The Jenkins pipeline is still broken. I'll have a look as well
> [ERROR]     Non-resolvable parent POM for org.eclipse.gef_root:org.eclipse.gef.promotion:3.20.0-SNAPSHOT: The following artifacts could not be resolved: org.eclipse.gef_root:org.eclipse.gef.releng:pom:3.20.0-SNAPSHOT (absent): Could not find artifact org.eclipse.gef_root:org.eclipse.gef.releng:pom:3.20.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 8, column 11 -> [Help 2]

PR #633, opened 2024-12-21
Correct version of parent pom on "promotion" module

ptziegler:
This module is only build on the "master" Jenkins build, which is why no error was thrown when raising the version of the parent pom.


ptziegler:
Merging it, as it only changes the pom.

PR #637, opened 2024-12-23
Make palette background configurable and delete PaletteColorUtil

ptziegler:
This removes the internal PaletteColorUtil class and replaces all references to the WIDGET_LIST_BACKGROUND (and its variants) with calls to the PaletteColorProvider.


ptziegler:
With should be the last PR that is needed to finish https://github.com/eclipse-gef/gef-classic/pull/589

azoitl (approved):

PR #639, opened 2024-12-27
Implement IGNORE_INVISIBLE_LAYOUT Zest Style for Zest 2.0

ptziegler:
With this change, invisible graph nodes and connections are ignored within the LayoutContext and therefore within the layout algorithms.

Note that as part of this change, the Zest styles are no longer set inside the Graph constructor, but via a separate setGraphStyle() method. This is because the constructor styles are also used for creating the underlying FigureCanvas, which would otherwise throw an exception due to the unsupported styles.

While it technically breaks compatibility, it is practically impossible to have used those styles up until now.

Fixes https://github.com/eclipse-gef/gef-classic/issues/608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## GEF
 
 ## Zest
+- The Zest Graph style have to be set via the `setGraphStyle()` method rather than the constructor. Reason being that otherwise, the same style is used to initialize the underlying `FigureCanvas`.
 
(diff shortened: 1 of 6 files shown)


azoitl (commented):
Schould we also add this information to the changelog?

ptziegler:
> Schould we also add this information to the changelog?

Added.

azoitl (approved):

PR #640, opened 2024-12-27
Change Ant's echo log level from "Warning" to "Info"

ptziegler:
Due to the new version of the maven-antrun-plugin, every echo instruction is treated as a warning.

See https://ant.apache.org/manual/Tasks/echo.html

### DIFF
--- a/org.eclipse.draw2d.doc.isv/buildDoc.xml
+++ b/org.eclipse.draw2d.doc.isv/buildDoc.xml
@@ -7,18 +7,18 @@
 	</target>
 
 	<target name="generateJavadoc">
-		<echo message="Cleaning reference/api..." />
+		<echo level="info" message="Cleaning reference/api..." />
 		<delete dir="reference/api" />
 		<mkdir dir="reference/api" />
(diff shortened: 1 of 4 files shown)


ptziegler:
For comparision, this is what the log output looks like with version 1.7:

```
[INFO] Executing tasks

main:
     [echo] Creating backup of about.mappings.
     [copy] Copying 1 file to /home/patrick/Eclipse/gef-classic-master/git/gef-classic/org.eclipse.draw2d
     [echo] Replacing @build@ token within about.mappings with build id.
```

And this is what it looks like now:
```
[INFO] Executing tasks
[WARNING]      [echo] Creating backup of about.mappings.
[INFO]      [copy] Copying 1 file to /home/patrick/Eclipse/gef-classic-master/git/gef-classic/org.eclipse.draw2d
[WARNING]      [echo] Replacing @build@ token within about.mappings with build id.
```

ptziegler:
This leaves 4 build warnings/errors which I have to investigate separately.

azoitl (approved):

PR #642, opened 2024-12-27
Update the Shapes example to show how to customize the Palette

ptziegler:
See https://www.eclipse.org/forums/index.php/t/72678/


ptziegler:
This effectively implements the theme used in WindowBuilder. The first image is the normal theme, the second image the 3D theme.
image
image

org.eclipse.gef/src/org/eclipse/gef/internal/ui/palette/editparts/DrawerEditPart.java
```diff
@@ -99,6 +100,11 @@ public <T> T getAdapter(final Class<T> key) {
 		if (key == MouseWheelHelper.class) {
 			return key.cast(new ViewportMouseWheelHelper(this));
 		}
+		if (key == Clickable.class) {
+			DrawerFigure drawerFigure = getFigure();
+			Clickable drawerToggle = drawerFigure.getCollapseToggle();
+			return key.cast(drawerToggle);
+		}
```
ptziegler, line 107: I'm not entirely sure whether it's a good idea to adapt the `DrawerEditPart` to a `Clickable`, or only the `DrawerFigure`. The latter means more work, as the `DrawerFigure` then needs to implement the `IAdaptable` interface.

And given that all of these are internal classes, you can't access them directly but instead have to go via the `EditPart` and `IFigure` interfaces.

azoitl: I think stating with the editpart is a good idea.

azoitl (approved):

PR #645, opened 2025-01-02
Fix deprecation warnings in Zest examples

ptziegler:
This primarily solves the warnings caused by the deprecated GraphNode constructor, but also does a general cleanup and modernization of the snippets.


ptziegler:
There are still roughly 100 i18n warnings, but those I'd like to handle separately.

azoitl (approved):

PR #646, opened 2025-01-02
Publish GEF/Zest examples and Zest CloudIO to update site

ptziegler:
Primarily done to enable the baseline-check for those bundles, but also to make them available to users. This includes:

- org.eclipse.gef.examples.digraph1
- org.eclipse.gef.examples.digraph2
- org.eclipse.gef.examples.ui.capabilities
- org.eclipse.gef.examples.ui.pde
- org.eclipse.zest.examples.cloudio
- org.eclipse.zest.examples.layouts
- org.eclipse.zest.cloudio

### DIFF
--- a/org.eclipse.gef.examples-feature/feature.xml
+++ b/org.eclipse.gef.examples-feature/feature.xml
@@ -69,4 +69,20 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="org.eclipse.gef.examples.digraph1"
+         version="0.0.0"/>
+
+   <plugin
+         id="org.eclipse.gef.examples.digraph2"
+         version="0.0.0"/>
+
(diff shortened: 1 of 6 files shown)


azoitl:
I like the idea of making the examples more accessible. But is a API baseline check for the examples really a good idea? Examples is nothing users should base there work on so fixing the API may be rather limiting for improving/reworking the examples.

ptziegler:
Almost all of the examples don't export their packages, so effectively, there is no API to check against.

But the API check is also not the intention. It's simply to have a validation that e.g. the minor version needs to be updated. That should still work without API, simply because the bundles are not binarily identical.

azoitl:
Ah that is a good point. Thx for the clarification.

azoitl (approved):

PR #648, opened 2025-01-03
Add new chapter regarding how Zest graphs work

ptziegler:
This adds a new topic to the Zest documentation regarding how the Zest Graph, GraphNode and GraphConnection works and how they can be customized by the user.


ptziegler:
I tried to figure out what exactly the `modelFigure` and `nodeFigure` inside the GraphNode represent, which then led me to writing everything down. @azoitl I think you are more familiar with the "hide node" feature, so perhaps you can double-check whether I got that part right?

sebHollersbacher (approved):

azoitl (approved):

PR #653, opened 2025-01-08
Cleanup Zest test cases

ptziegler:
- Remove old JUnit 3 API
- Remove dead code
- Suppress i18n warnings for error messages
- Suppress warnings related to internal API
- Cleanup remaining compile warnings

### DIFF
--- a/org.eclipse.zest.tests/.settings/.api_filters
+++ b/org.eclipse.zest.tests/.settings/.api_filters
@@ -9,4 +9,45 @@
             </message_arguments>
         </filter>
     </resource>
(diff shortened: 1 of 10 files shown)


azoitl (commented):
Do we really need api filters for tests?

ptziegler:
> Do we really need api filters for tests?

Probably not. The IDE is complaining that there is no baseline for the test bundles anyway, given that they are not published. I'll create a separate PR where I'll remove it for the other test bundles as well.

azoitl (approved):

PR #654, opened 2025-01-09
Remove API Analysis nature from Zest test bundle

ptziegler:


ptziegler:
Turns out the API analysis is only enabled for the Zest tests.

azoitl (approved):

PR #656, opened 2025-01-11
Refresh Zest GraphNode figures if node style has been changed

ptziegler:
There are currently two ways the Zest styles for a graph node can be set: Either via the constructor or by calling the setNodeStyle() method.

Certain styles affect how the nodes are visualized and therefore, the underlying figure needs to be updated together with the node style. This currently only happens when the node is created. As a result, certain styles such as the NODES_HIDE_TEXT and NODES_FISHEYE are effectively ignored if retroactively added to a node.

With this change, the model figure is explicitly updated (which is also done when e.g. updating the font). Note that mixing SWT and Zest styles is usually not a good idea, as it introduces the risk in case they share one or more bits. Which is why the test snippets have been updated as well, to emphasize the use of setNodeStyle() as the indented workflow.

Fixes https://github.com/eclipse-gef/gef-classic/issues/647


ptziegler:
https://download.eclipse.org/staging/2025-03 seems to be having some network issues right now. I'll rerun the failing jobs sometime later.

azoitl (approved):

PR #657, opened 2025-01-11
Double size for AND,OR,XOR gates for better visibility

Amadeus1791:
Changes:
- Increased size of AND, OR, and XOR gates to double their original size
- Maintained original aspect ratios and proportions
- Note: Connection and anchor point alignments will be addressed in a follow-up PR

BEFORE:
before

AFTER: 
after

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/OrGateFigure.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/OrGateFigure.java
@@ -21,19 +21,19 @@ import org.eclipse.draw2d.geometry.Rectangle;
  * @author danlee
  */
 public class OrGateFigure extends GateFigure {
-	public static final Dimension SIZE = new Dimension(15, 17);
+	public static final Dimension SIZE = new Dimension(30, 34);
 	protected static final PointList GATE_OUTLINE = new PointList();
 
 	static {
-		GATE_OUTLINE.addPoint(2, 10);
-		GATE_OUTLINE.addPoint(2, 2);
+		GATE_OUTLINE.addPoint(4, 20);
 		GATE_OUTLINE.addPoint(4, 4);
-		GATE_OUTLINE.addPoint(6, 5);
-		GATE_OUTLINE.addPoint(7, 5);
-		GATE_OUTLINE.addPoint(8, 5);
-		GATE_OUTLINE.addPoint(10, 4);
-		GATE_OUTLINE.addPoint(12, 2);
+		GATE_OUTLINE.addPoint(8, 8);
 		GATE_OUTLINE.addPoint(12, 10);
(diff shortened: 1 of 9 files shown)


ptziegler:
Just out of curiosity: What is the motivation behind increasing the size of those figures, give that the logic editor only serves as an example?

Amadeus1791:
@ptziegler 
I would like to make some attempts to enhance the visual appeal and usability of the logic designer. The primary reason for this change is that the original component sizes, particularly in comparison to elements like the LED, appeared disproportionately small. This not only affected the visual balance of the editor but also made interacting with the components more difficult. For instance, selecting them with the mouse was often challenging due to their small size.

azoitl:
I would say the logic editor is somehow the reference example, having it more modern, can also help GEF Classic users to better assess the possibilities of GEF Classic. With increasing monitor resolutions the existing sizes where way to small.

ptziegler:
@Amadeus1791 Thanks for the explanation! 😄 

Regarding the test failure, I think that this is simply a rounding error when aligning the figures by their center. 
 
https://github.com/eclipse-gef/gef-classic/blob/b5fa170981254231f262d3f960a41a5cbe9e12dd/org.eclipse.gef.tests/src/org/eclipse/gef/test/swtbot/LogicDiagramTests.java#L262-L270

I don't think that this is a general issue, so I think it should be sufficient to simply add a little bit of tolerance to the equality check.

```java
		assertEquals(first.getBounds().getCenter().y, second.getBounds().getCenter().y, 1.0);
```

azoitl (approved):

org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/OrGateFigure.java
```diff
@@ -21,19 +21,19 @@
...
-		GATE_OUTLINE.addPoint(8, 5);
-		GATE_OUTLINE.addPoint(10, 4);
-		GATE_OUTLINE.addPoint(12, 2);
+		GATE_OUTLINE.addPoint(8, 8);
 		GATE_OUTLINE.addPoint(12, 10);
+		GATE_OUTLINE.addPoint(14, 10);
+		GATE_OUTLINE.addPoint(16, 10);
+		GATE_OUTLINE.addPoint(20, 8);
+		GATE_OUTLINE.addPoint(24, 4);
+		GATE_OUTLINE.addPoint(24, 20);
```
ptziegler, line 36: I just noticed that sadly, this calculation is not entirely correct.

For the original figure the points went from x = 2 to x = 12 (i.e 10 pixel). Now the points go from x = 4 to x = 24 (20 pixel). With a line width of 1 pixel, this then results in a total width of 11 and 21 pixels. However, the new figure is supposed to be 22 pixels wide.

In short, the Or, XOR and AND figures are no longer centered.

image

PR #659, opened 2025-01-13
Double size of remaining components of logic editor

Amadeus1791:
- Those changes would complete the double sizing of all components for better visibility. 
- The other changes were done before in #658 and #657)

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/CircuitBorder.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/CircuitBorder.java
@@ -22,24 +22,24 @@ import org.eclipse.draw2d.geometry.Rectangle;
 
 public class CircuitBorder extends AbstractBorder {
 
-	protected static Insets insets = new Insets(8, 6, 8, 6);
+	protected static Insets insets = new Insets(16, 12, 16, 12);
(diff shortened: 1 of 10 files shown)


azoitl:
It looks like the changes in the LED led to some issues with grid alignment.

ptziegler:
Just a thought: If the size of the figures are doubled, then having a threshhold of only 5 pixels for the snap-to-grid is really low. Would it make sense to double that as well?

Amadeus1791:
Thanks for you feedback, I appreciate a lot. I added a follow-up commit to fix grid and snap-to-geometry tests. Those two test were affected by the component size changes. It happened since the snap points shifted relative to the grid due to the changed component boundaries.

azoitl (approved):

ptziegler:
@azoitl I think for the future, it would be better to just squash multiple commits. Or is there any benefit in keeping them separate?

image

azoitl:
In this specific case I was not sure. Normally I do not squash if each of the commits is by itself an own independent contribution. I kept it separate because I didn't want to loos the information that changing sizes may have other impact. Not sure if that was good.

ptziegler:
This itself is fine, but I usually like to avoid doing a "rebase and merge" when there are multiple commits. Otherwise it's simply not obvious that both commits are related, if one would ever need to revert them.

azoitl:
@ptziegler that is a good point. Will keep that in mind.

PR #665, opened 2025-01-16
Don't use instanceof for null checks

iloveeclipse:
It was a bug in older ecj compiler that allowed to use instanceof patterns in Java 17 code.

Fixes https://github.com/eclipse-gef/gef-classic/issues/664

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/dnd/TemplateTransferDropTargetListener.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/dnd/TemplateTransferDropTargetListener.java
@@ -127,10 +127,11 @@ public class TemplateTransferDropTargetListener extends AbstractTransferDropTarg
 		}
 		EditPartViewer viewer = getViewer();
 		viewer.getControl().forceFocus();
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@ptziegler : could you please review & merge? Thanks.

merks:
@iloveeclipse 

It looks fine and I would merge it, but the state is broken:

image

iloveeclipse:
> It looks fine and I would merge it, but the state is broken:

There shouldn't be rebase needed, it is on top of master?

merks:
There is no green button that I can press. Very strange because it looks based on master

image

Apparently I can do this, but most projects don't like people doing that:

image

I can try pushing the change to my own fork...

merks:
So I created this and we can see if that has a better state after the build is done:

https://github.com/eclipse-gef/gef-classic/pull/667

merks:
After doing https://github.com/eclipse-gef/gef-classic/pull/667 then this one went green.  But since that's merged we can close this.

iloveeclipse:
Thanks Ed!

PR #668, opened 2025-01-16
Update the setup to configure a JRE for Java 21

merks:


merks:
This has to build impact so merging now.

PR #670, opened 2025-01-16
Deprecate transfer listeners for removal

ptziegler:
The TransferDragSourceListener and TransferDropTargetListener are effecitively equivalent to their JFace counterpart and should be used instead.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -141,8 +141,11 @@ public abstract class AbstractEditPartViewer implements EditPartViewer {
 
 	/**
 	 * @see EditPartViewer#addDragSourceListener(org.eclipse.gef.dnd.TransferDragSourceListener)
+	 * @deprecated Use {@link #addDragSourceListener(TransferDragSourceListener)}
+	 *             instead. This method will be removed after the 2027-03 release.
 	 */
 	@Override
+	@Deprecated(since = "3.21", forRemoval = true)
 	public void addDragSourceListener(org.eclipse.gef.dnd.TransferDragSourceListener listener) {
 		addDragSourceListener((TransferDragSourceListener) listener);
 	}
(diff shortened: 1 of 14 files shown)


org.eclipse.gef/src/org/eclipse/gef/ui/parts/AbstractEditPartViewer.java
```diff
@@ -294,9 +306,13 @@ public Control getControl() {
 	 * Returns <code>null</code> or the DelegatingDragAdapater. The adapter is
 	 * created automatically when
 	 * {@link #addDragSourceListener(TransferDragSourceListener)} is called.
+	 * <i>Important</i> The {@link org.eclipse.gef.dnd.DelegatingDragAdapter} will
+	 * be removed in a future release, so clients should store the return type in a
+	 * {@link DelegatingDragAdapter}, if required.
 	 *
 	 * @return <code>null</code> or the adapter
 	 */
+	@SuppressWarnings("removal")
```
ptziegler, line 315: I'm a little bit stuck with the `getDelegatingDragAdapter()` and `getDelegatingDropAdapter()` methods. Users should switch to the JFace adapters but can't, because those methods still return the deprecated types. But I can't change the return type without breaking API.

azoitl: Isn't that a bit similar to what happened with the zoomlisteners?

ptziegler: Yes and no. In the end, there needs to be a way for users to migrate away from the deprecated API _before_ the adapters are deleted.

In case of the ZoomListener, the `addZoomListener()` and `removeZoomListener()` methods were overloaded, which is possible because they have different parameters.

For the `getDelegatingDragAdapter()` and `getDelegatingDropAdapter()`, the only difference is the return type and therefore overloading is not possible. 

If I were to do the same here, I would have to give those "new" methods a different name. Which I'm not really fond of because I think the current names are already good the way they are.

ptziegler: Alternatively, I could also change it so that you get those adapters via `getAdapter(DelegatingDropAdapter.class)` and ``getAdapter(DelegatingDragAdapter)`, which would save the hassle of juggling with the method names.

ptziegler: Let's focus on what _can_ be deprecated: The TransferDragSourceListener and TransferDropTargetListener. How to handle the adapters is something we can still look at at a later date.

azoitl: I think this is a good idea!

azoitl (approved):

PR #672, opened 2025-01-17
Deprecate CommandStackListener for removal

ptziegler:
This interface has been deprecated a long time ago in favor of the CommandStackEventListener.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -38,6 +38,8 @@ import org.eclipse.gef.EditPart;
 import org.eclipse.gef.GraphicalEditPart;
 import org.eclipse.gef.GraphicalViewer;
 import org.eclipse.gef.commands.CommandStack;
+import org.eclipse.gef.commands.CommandStackEvent;
+import org.eclipse.gef.commands.CommandStackEventListener;
 import org.eclipse.gef.commands.CommandStackListener;
 import org.eclipse.gef.ui.actions.ActionBarContributor;
 import org.eclipse.gef.ui.actions.ActionRegistry;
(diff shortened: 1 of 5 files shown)


org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
```diff
@@ -334,7 +352,7 @@ protected void hookGraphicalViewer() {
 	public void init(IEditorSite site, IEditorInput input) throws PartInitException {
 		setSite(site);
 		setInput(input);
-		getCommandStack().addCommandStackListener(this);
```
azoitl, line 337: This change may break clients overriding the `commandStackChanged()` method. We somehow also need to call this. Maybe in the new `stackChanged(()`method?

ptziegler: > This change may break clients overriding the `commandStackChanged()` method. We somehow also need to call this. Maybe in the new `stackChanged(()`method?

Good point. Luckily, CommandStackEvent extends EventObject, so the method call can simply be delegated.

ptziegler: Hm... subclasses could also implement CommandStackEventListener without calling the super-method which could also cause issues.

I think it's better to keep using the CommandStackListener for the time being, given that all of this is internal.

azoitl: But now a different amount of events is generated at different points in the command stack state. This makes me a bit nervous. But I may be overreacting.

azoitl: Oh that could be a problem yes.

azoitl (approved):

org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
ptziegler: > Oh that could be a problem yes.

But that's the risk of every change and something I would consider rather unlikely (because why wouldn't use the existing commandStackChanged() method instead)?

As an example, the addition of the getSelectedEditParts() method in the SelectionAction caused errors in our application, because our subclass had a method with identical signature:

https://github.com/eclipse-gef/gef-classic/blob/2a9227d58fd95cb388b7b524606742e767271123/org.eclipse.gef/src/org/eclipse/gef/ui/actions/SelectionAction.java#L98-L104

In the end, that's just normal software development 🤷

azoitl: You are right and in the end we want to make GEF correcter and better maintainable. That does not come for free.

PR #673, opened 2025-01-18
Remove deprecation from getDistanceOrthogonal #21

azoitl:
getDistanceOrthogonal is used in several places in GEF. There is no straight forward replacement. To reason for deprecation is precision of subclasses. As there is only one subclass with higher precision an improved version of that method is added there. Furthermore the impact on precision will anyhow only happen if both sides are PrecisionPoints.

https://github.com/eclipse-gef/gef-classic/issues/21

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
@@ -113,6 +113,23 @@ public class PrecisionPoint extends Point {
 	}
 
 	/**
+	 * provide a precision version of
+	 * org.eclipse.draw2d.geometry.Point#getDistanceSquared(org.eclipse.draw2d.geometry.Point)
+	 *
+	 * @since 3.19
+	 */
+	@Override
+	public int getDistanceSquared(Point p) {
+		double deltaX = p.preciseX() - this.preciseX();
+		double deltaY = p.preciseX() - this.preciseX();
+		long result = (long) (deltaX * deltaX + deltaY * deltaY);
+		if (result > Integer.MAX_VALUE) {
+			return Integer.MAX_VALUE;
+		}
+		return (int) result;
+	}
+
+	/**
 	 * Returns a precise copy of this.
 	 *
 	 * @return a precise copy
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
@@ -43,7 +43,6 @@ public class PrecisionPoint extends Point {
 	 * Constructor for PrecisionPoint.
 	 */
 	public PrecisionPoint() {
-		super();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


ptziegler:
I agree that it doesn't make sense to deprecate `getDistanceOrthogonal()`, but not e.g. `getDistanceSquared()`, which suffers from the same problem.

That said: I'm not sure if it makes sense to override the method in `PrecisePoint`, given that you still have the imprecision due to the cast to `int`. Perhaps it's better to create a `getPreciseDistanceOrthogonal(Point)` / `getDistanceOrthorgonal(PrecisionPoint)` method?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
```diff
@@ -43,7 +43,6 @@ public class PrecisionPoint extends Point {
 	 * Constructor for PrecisionPoint.
 	 */
 	public PrecisionPoint() {
-		super();
```
ptziegler, line 46: I assume this was removed by the auto-formatter. But then I would get rid of the empty constructor alltogether.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
```diff
@@ -102,6 +101,17 @@ public Dimension getDifference(Point p) {
...
+	/**
+	 * provide a precision version of
+	 * org.eclipse.draw2d.geometry.Point#getDistanceOrthogonal(org.eclipse.draw2d.geometry.Point)
+	 *
+	 * @since 3.19
+	 */
+	@Override
+	public int getDistanceOrthogonal(Point p) {
+		return (int) (Math.abs(this.preciseX() - p.preciseX()) + Math.abs(this.preciseY() - p.preciseY()));
+	}
```
ptziegler, line 113: Would it make sense to do the same for `getDistanceSquared(Point)` to make things consistent?

azoitl:
> That said: I'm not sure if it makes sense to override the method in `PrecisePoint`, given that you still have the imprecision due to the cast to `int`. Perhaps it's better to create a `getPreciseDistanceOrthogonal(Point)` / `getDistanceOrthorgonal(PrecisionPoint)` method?

I don't think this makes much sense. The reason I overwrote it is to reduce rounding errors for the deltas and sums. returning a double would mostly only be needed of you sum-up several of the distances. But what could make sense is to round the result instead of casting it to int. What do you think?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
azoitl: Yes this was the auto-formater.  I can not remove this constructor otherwise you can not create a PrecisionPoint without coordinates.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
azoitl: I like that idea. Will do.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionPoint.java
ptziegler: Ah, I didn't notice there were other constructors.

ptziegler (approved):

PR #676, opened 2025-01-19
Modernize Gate Figures (AND, OR, XOR)

Amadeus1791:
This PR proposes a more modern design for the gates of the logic designer:
- Introduced thicker outlines for improved clarity.
- Added a new shared outline color (dark gray) to replace the monotone black color.
- Rounded edges in the AND gate for a softer and modernized appearance.
- Updated colors: The AND gate uses a soft red, the OR gate azure blue and the XOR gate a more vibrant yellow.
- Adjusted offsets, and clipping logic due to the new line width.

Follow-up PRs will be: 
- modern design of the reamining components 
- adjustment of the corresponding feedback figures 

Before:
gates - before

After:
gates -after

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/LogicColorConstants.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/LogicColorConstants.java
@@ -16,9 +16,10 @@ import org.eclipse.swt.graphics.Color;
 
 public interface LogicColorConstants {
 
-	public final static Color andGate = new Color(null, 220, 70, 80);
-	public final static Color orGate = new Color(null, 0, 134, 255);
-	public final static Color xorGate = new Color(null, 240, 240, 40);
+	public final static Color andGate = new Color(null, 204, 61, 61);
+	public final static Color orGate = new Color(null, 41, 121, 255);
+	public final static Color xorGate = new Color(null, 255, 204, 0);
+	public static final Color outlineColor = new Color(null, 64, 64, 64);
 	public final static Color logicGreen = new Color(null, 123, 174, 148);
 	public final static Color logicHighlight = new Color(null, 66, 166, 115);
 	public final static Color connectorGreen = new Color(null, 0, 69, 40);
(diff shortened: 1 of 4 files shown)


org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/LogicColorConstants.java
```diff
@@ -16,9 +16,10 @@
 
 public interface LogicColorConstants {
 
-	public final static Color andGate = new Color(null, 220, 70, 80);
-	public final static Color orGate = new Color(null, 0, 134, 255);
-	public final static Color xorGate = new Color(null, 240, 240, 40);
+	public final static Color andGate = new Color(null, 204, 61, 61);
```
azoitl, line 19: The first parameter is not needed anymore so as you are anyhow touching the code here you could remove all nulls in this interface.

azoitl (changes requested):

org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/LogicColorConstants.java
azoitl: thx

azoitl (approved):

ptziegler:
Just a minor thing I noticed while trying it out as well. I get the impression that the edges are sometimes not drawn properly. At 400% zoom and with a thinner outline, for better visualization of the problem:

image

I assume some form of rounding error is happening here?

azoitl:
with the latest version with thicker borders this is gone. But I noticed that we have some rounding issues in Draw2d that we definitely will need to look into. Especially which what is going on with swt31.

ptziegler:
> with the latest version with thicker borders this is gone

I'm using 55394a8f9226e1b4e7e648c0b869cd9d897a17b6, which hopefully contains everything up until now.

For comparison, the OR gate is using the "new" implementation and the XOR gate the "old" one.  I don't believe that this is a rounding error but simply the filled oval being too big.

image

You won't notice this with thicker border, but I think this is simply because they camouflage the affected area.

image

Looking closer, I think this is caused by the oval being offset, due it being shifted along the x coordinate, which is then compensated by reducing the width twice.

image

Because the width is reduced by two, but the height only reduced by one, both ovals have different geometric shapes, which then causes those visual artifacts.

The thing I don't understand is why those new offsets are better. The oval was perfectly centered before:
image

And I don't see any noise, even with the thicker border.
image

azoitl:
@ptziegler thx for looking deeper into that. I experienced such issues in another Zest project (not public) so I didn't think it about to carefully. As you showed @ptziegler I was wrong. @Amadeus1791 can you please look into that.

PR #680, opened 2025-01-26
Create example on how to use separate fonts for palette entries

ptziegler:
This snippet allows the user to select different fonts for the palette drawer and tool entries via the "Settings" dialog and can be seen by opening the "Palette with separate fonts" view in the Eclipse IDE.

Closes https://github.com/eclipse-gef/gef-classic/issues/371


ptziegler:
This is how it looks like:
image

I don't think this use-case really fits into any of the existing examples, so I've created a new plugin where in the future, I would put any other snippets as well.

azoitl (approved):

PR #708, opened 2025-03-15
Generalize PrintOperation to handle generic Graphics implementations

ptziegler:
This removes the dependency from PrintOperation to PrintGraphics, allowing subclasses to inject their own drawing backend by overriding the createGraphics(SWTGraphics,Printer) method.

The following methods have been deprecated, with drop-in replacements available that work on plain Graphics objects:
- PrinterGraphics getFreshPrinterGraphics()
- setupGraphicsForPage(PrinterGraphics)

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/PrintOperation.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/PrintOperation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2010 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -31,7 +31,7 @@ public abstract class PrintOperation {
 							// print job
 	private Insets printMargin = new Insets(0, 0, 0, 0);
 	private Printer printer;
-	private PrinterGraphics printerGraphics;
+	private Graphics printerGraphics;
 	private SWTGraphics g;
 
 	/**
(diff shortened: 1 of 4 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/PrintOperation.java
```diff
@@ -65,23 +65,63 @@ protected void cleanup() {
...
+	 * @return Graphics The new Graphics
+	 * @since 3.19
+	 */
+	protected Graphics getFreshGraphics() {
+		try {
+			return getFreshPrinterGraphics();
+		} catch (ClassCastException ignore) {
+			// createGraphics(SWTGraphics,Printer) has been subclasses
+		}
+		return doGetFreshGraphics();
```
ptziegler, line 94: This is a little bit of a necessary evil, as `getFreshPrinterGraphics()` might've been overriden by subclasses. This check can't fail for existing code, as they must use a `PrinterGraphics` object.

 But for new code where a generic Graphics object is used, this check will always fail and then use the fallback implementation. This also has the unpleasant side-effect that currently, the Graphics objects are created twice. However, this problem should sort itself out once we are able to delete the deprecated API.

org.eclipse.draw2d/src/org/eclipse/draw2d/PrintOperation.java
```diff
@@ -185,11 +225,35 @@ public void setPrintMargin(Insets margin) {
...
+	 * @since 3.19
+	 */
+	protected void setupGraphicsForPage(Graphics g) {
+		try {
+			setupGraphicsForPage((PrinterGraphics) g);
+			return;
+		} catch (ClassCastException ignore) {
+			// createGraphics(SWTGraphics,Printer) has been subclasses
+		}
+		doSetupGraphicsForPage(g);
```
ptziegler, line 250: Same reasoning as with the `getFreshGraphics()`.

ptziegler:
For e.g. https://github.com/eclipse-gef/gef-classic/pull/237, you would use something like this:

```java
PrintOperation op = new PrintFigureOperation(p, fig) {
    protected Graphics createGraphics(SWTGraphics g, Printer p) {
        return g;
    }
};
```

azoitl (changes requested):
Interesting solution. I never thought I could completely ignore `PrinterGraphics`. I like the flexibility it gives.

As this is the first change to draw2d can you please bump the minor version.

ptziegler:
> Interesting solution. I never thought I could completely ignore `PrinterGraphics`. I like the flexibility it gives.

The `PrinterGraphics` doesn't provide any additional API that is required for the `PrintOperation`. So there is no real benefit to using a more specific type.

I'm a little disappointed that `getFreshPrinterGraphics()` can't be overloaded to return a plain Graphics object without breaking compatibility, but I believe introducing `getFreshGraphics()` is as good as I can make it.

> As this is the first change to draw2d can you please bump the minor version.

Done.

azoitl (approved):

azoitl:
> > Interesting solution. I never thought I could completely ignore `PrinterGraphics`. I like the flexibility it gives.
> 
> The `PrinterGraphics` doesn't provide any additional API that is required for the `PrintOperation`. So there is no real benefit to using a more specific type.

And this is exactly what I failed to see.

 
> I'm a little disappointed that `getFreshPrinterGraphics()` can't be overloaded to return a plain Graphics object without breaking compatibility, but I believe introducing `getFreshGraphics()` is as good as I can make it.
 
I totally agree.

Issue #705, opened 2025-03-12
GEF Classic 3.24 Release for Eclipse 2025-06


azoitl:
This issue is used to track and coordinate the work for the GEF Classic 3.24.0 release to be contributed to the Eclipse 2025-06 release train. It shall serve as checklist as well as template for the upcoming release.

- [x] Create 3.24 Milestone
- [x] Update version number and target platforms:
  - [x] Update the version number of all features to 3.24.0
  - [x] Bump maintenance version number of all plugins
  - [x] Update the target platform for the API baseline to GEF Classic 3.24.0
  - [x] Update the pom file target platform checking to GEF Classic 3.24.0
  - [x] Update the development target platform to Eclipse 2025-06
- [x] 07.04.2025: Release 3.24.0 M1 for 2025-06 M1
  - [x] create release record
- [x] 28.04.2025: Release 3.24.0 M2 for 2025-06 M2
- [x] 19.05.2025: Release 3.24.0 M3 for 2025-06 M3
- [x] 26.05.2025: Release 3.24.0 RC1 for 2025-06 RC1
- [x] 02.06.2025: Release 3.24.0 for 2025-06 RC2
  - [x] create R3_24_maintenance branch
- [x] 11.06.2025: 2025-06 Release
  - [x] tag release commit and mark it as release in Github, create release notes

azoitl:
@ptziegler I tried to do the milestone release but somehow our maven build fails. I could trace the problem back till #705 but I'm not really sure what the problem is as we get strange exception from junit. I tried to upgrade to Tycho 4.0.12 but then I get complaints about Java 17. Do you have any idea what is wrong here?

merks:
I think you must update tycho because older versions don’t work with latest junit. 

I had to tweak a bunch of things to get datatools working 

https://github.com/eclipse-datatools/datatools/commit/daaa60f9523d48cc6a65d3b14c97cbba2a06ad91

This in the target configuration is likely key
```
<executionEnvironment>JavaSE-21</executionEnvironment>
```

ptziegler:
Hm... The main problem is that the Draw2D tests are no longer executed, which then leads to an error as no results can be recorded.

It works with Tycho 4.0.12 and Java 21 as @merks said, so my guess is that one of the M1's is mixing up the JUnit 5 plugin that is used for the tests.

[...]

PR #709, opened 2025-03-15
Prepared Eclipse GEF Classic for 324 Devlopements #705

ptziegler:
- Update the version number of all features to 3.24.0
 - Bump maintenance version number of all plugins
 - Update the target platform for the API baseline to GEF Classic 3.24.0
 - Update the pom file target platform checking to GEF Classic 3.24.0
 - Update the development target platform to Eclipse 2025-06
 - Added changelog entry

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 		<cbi-plugins.version>1.5.2</cbi-plugins.version>
 		<tycho.scmUrl>scm:git:https://github.com/eclipse-gef/gef-classic.git</tycho.scmUrl>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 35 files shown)


pom.xml
```diff
@@ -58,6 +58,9 @@
 					<sourceReferences>
 						<generate>true</generate>
 					</sourceReferences>
+					<archive>
+						<addMavenDescriptor>false</addMavenDescriptor>
+					</archive>
```
ptziegler, line 63: This is likely what caused the baseline check to complain for every plugin. This adds the Maven metadata to the META-INF folder, including e.g. dependencies. Given that those change every release, you always get a diff.

azoitl (approved):
@ptziegler I wanted to start this today. You where faster :-)

pom.xml
azoitl: Cool that you found that.

Issue #710, opened 2025-03-18
Vertical ruler in GEF looks bad on HiDPI displays


wiresketch:
I'd like to resurrect the discussion about fixing vertical rulers on HiDPI displays that I've started back in 2017 on bugzilla here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=522342

Although I've tried to contribute a patch at that point, no progress was made since GEF was in limbo back then. The issue however has gotten worse on Windows since then. Here's what it looks like with a 200% scaling:

![Image](https://github.com/user-attachments/assets/6f1e827d-fc3a-4c6c-b85c-85a7c5c6ee75)

Things look a little bit better on Linux, but it still visible that numbers are washed out on vertical ruler compared to horizontal ruler:

![Image](https://github.com/user-attachments/assets/5e2465e4-88b0-46b7-8d03-1d615856ae8f)

I believe that the best solution is the same as I've proposed back then: using advanced graphics to draw rotated text for vertical major marks.

I'm willing to rework my original patch and contribute it here for review.

azoitl:
@wiresketch thx for bringing this up again. We had some discussions with text and native scaling (e.g., #706). Just out of curiosity are you are you using ScaledGraphics or SWTGraphics?

wiresketch:
@azoitl I do not see how #706 is related to this. The issue I am describing is caused by the implementation of RulerFigure.drawVerticalMajorMark method which creates an intermediate image for rotated text. The way this method is using SWT images is outdated and doesn't support HiDPI monitors nor does it support environments having multiple monitors with different DPIs. The GEF component in question is RulerComposite and internally it uses ScrollingGraphicalViewer for top and left rulers and the figures are drawn using SWTGraphics as it would be expected. This is not related to printing in any way.

The fix I am proposing is to use SWTGraphics to draw rotated text for the left ruler without using intermediate SWT images.

azoitl:
@wiresketch thx for clarification. Every time I see some cropped text I put it for me in the same group of problems. ScaledGraphics has quite some issue with it and therefore I was asking that. 

Forgot to write in my last message that I would be happy if you could clean-up your patch and submit it as PR.

PR #711, opened 2025-03-18
Fixes ruler issues on HiDPI displays #710

wiresketch:
The fix for  #710.

The main fix removes the use of temporary rotated Image for vertical marks. The result before and after on 200% display on a Windows machine is the following:

ruler-hidpi-before

ruler-hidpi-after

I've also removed the background fill under text marks performed using graphics.fillRectangle(forbiddenZone). I believe that this is unnecessary and also draws over the bottom/right border as can be also seen in before and after images. To observe this better click on each image and check the border under the digits for horizontal ruler, and to the right for vertical ruler.

The final tweak is that I've enabled text aliasing for RulerFigure. Text aliasing works by default on Linux and macOS, but is needed to be enabled explicitly on Windows. These are before and after results:

ruler-hidpi-no-antialias

ruler-hidpi-text-antialias png

The difference can only be clearly seen when text marks are multi-digit, which happens when ruler unit is set to pixels. Without text antialias vertical digits are not well aligned.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerFigure.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerFigure.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.gef.internal.ui.rulers;
 
+import org.eclipse.swt.SWT;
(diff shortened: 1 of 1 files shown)


azoitl (approved):
LGTM. Thx for the fix.

PR #713, opened 2025-03-22
Add support for mouse-wheel events to figures

ptziegler:
This introduces a new MouseWheelListener interfaces that can be used by figures to respond to mouse-wheel events.

In addition, this also solves the issue where the figure under the mouse cursor was not updated, if it was moved via a scroll event, leading to "mouseExited" and "mouseEntered" not being fired and other visual artifacts such as tool-tips not disappearing.


ptziegler:
This is required for https://github.com/eclipse-windowbuilder/windowbuilder/issues/986.
In short: Our property table is realized as a GEF viewer and when hovering over a figure, a tooltip is shown. The problem is that this tooltip remains, even if you scroll the table and therefore no longer matches the figure under the cursor.

[Recording.webm](https://github.com/user-attachments/assets/ef4f66b0-c9ac-4416-9cd6-66011f6aba00)

org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/swtbot/AbstractSWTBotTests.java
```diff
@@ -136,6 +147,20 @@ protected void doTest(Snippet annotation, Invocation<Void> statement) throws Thr
 		}
 	}
 
+	protected void scrollVertical(int x, int y, int count) {
```
azoitl, line 150: should have have a mousewheel in the name to make it clearer and distinguish from other scrolls (e.g., scrollbar)?

azoitl (approved):

PR #717, opened 2025-04-07
Build GEF against both the SimRel version and the 3.20.0 release

ptziegler:
To make sure the latest GEF release remains compatible with older Eclipse release, we should not only build and test against the upcoming release, but also against previous releases.

With this change, the BREE for all test bundles is reverted back to Java 17, in order to allow compilation and execution with a Java 17 JDK and without having to configure a Maven toolchain.

Furthermore, the lower bound of all bundle dependencies is adjusted to be compatible with the older releases.


ptziegler:
I think that's the cleanest solution to make sure that we're compatible with Java 17; By building and testing against an actual Java 17.

By building and testing against an older Eclipse release, we also make sure that recent versions of GEF can still be used with older RCP versions.

merks:
Yes, EMF makes these things configurable too.

ptziegler:
Though I think EMF goes above and beyond and can even compile against ancient Eclipse releases. 😅 I don't think that's quite necessary for GEF.

I'm still not quite happy with some of the labels (e.g. "baseline" is way too generic), so I'll probably revisit this later today.

azoitl (approved):

ptziegler:
I've played around with some other releases, but the 2024-03 is the lowest we can go due to https://github.com/eclipse-platform/eclipse.platform.swt/commit/eb0f99db46d5ee8f6c345d8a3096efb401b41a03. I think that's fine for the majority of consumers.

PR #726, opened 2025-04-24
Use monitor client area for tool tip positions

akoch-yatta:
This PR adapts the calculation to correct tooltip positions. With monitor specific scaling added for the windows implementation of SWT Display#getClientArea is currently no reliable source to calculate whether the tooltip is inside the display bounds. It is replaced with Monitor#getClientArea instead.

### How to test
Setup similar to:
- Windows
- Multiple monitors with different zoom settings, most left monitor is primary monitor
- monitor-specific UI scaling activated

Move the IDE to the right monitor and position it so e.g. the tooltip of the Palette would be outside of the monitor -> without this PR the tooltip will be either positioned completely outside of the view of way more left as expected

In my opinion it makes more sense to position the tooltips always completely on the current monitor anyway independent of the mentioned issue.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ToolTipHelper.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ToolTipHelper.java
@@ -57,18 +57,21 @@ public class ToolTipHelper extends PopUpHelper {
 	 * otherwise it will be painted directly above cursor.
 	 */
 	private Point computeWindowLocation(IFigure tip, int eventX, int eventY) {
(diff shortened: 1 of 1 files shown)


ptziegler:
I wonder if this fix is somehow related to https://github.com/eclipse-gef/gef-classic/issues/204.

> In my opinion it makes more sense to position the tooltips always completely on the current monitor anyway independent of the mentioned issue.

That I agree with and from what I read, is also how the SWT tooltip is doing it now.

I'll try to test it during the weekend but so at least at first glance, everything looks reasonable.

ptziegler (approved):

ptziegler:
@akoch-yatta Thanks!

PR #729, opened 2025-05-02
Convert GEF images from PNG to SVG

ptziegler:
This moves the SVG images from the "icons" project to the GEF bundle and makes them the default source data. In case SVG is not yet supported by SWT, the original PNG images are used instead.


org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java
```diff
@@ -111,4 +118,32 @@ public static ImageData scaledImageData(ImageDescriptor descriptor, int zoom) {
...
+		List<BundleRequirement> requirements = wiring.getRequirements("eclipse.swt"); //$NON-NLS-1$
+		for (BundleRequirement requirement : requirements) {
+			String filter = requirement.getDirectives().get(Constants.FILTER_DIRECTIVE);
+			if ("(image.format=svg)".equals(filter)) { //$NON-NLS-1$
+				isSvgSupported = true;
+				break;
+			}
+		}
+		return isSvgSupported;
+	}
```
ptziegler, line 148: I'm still not sure whether this is the right approach or whether I'm overthinking stuff...

merks: I’m not suggesting to use emf but we had to solve the problem of working in an OSGi runtime and with a flat classpath decades ago. 

https://github.com/eclipse-emf/org.eclipse.emf/blob/157748e0ecdc2368d1cf451ca9c7fc78431c6c91/plugins/org.eclipse.emf.common/src/org/eclipse/emf/common/CommonPlugin.java#L106

ptziegler: Hm... Yeah, I think a simple try-catch block is good enough. There's no point going above and beyond to check whether SVGs are supported.

azoitl (approved):

Issue #376, opened 2024-02-13
Animations are not drawn


daell82:
Hello, 

after looking into the flow example again, I've noticed that the animations for the graph layout are not drawn anymore. I was able to track it down to the changes introduced in #127. This also affects the animations for palette drawers and animated zoom. 

Currently, for me this is reproducible on Windows. On Linux I am not sure whether this was ever supported.

ptziegler:
Can you perhaps post a screenshot of what you mean?

I've reverted the change locally, but I couldn't see any differences. Though that might partially be due to me using Linux, which you've already hinted at.

daell82:
Yes, for the recording I've duplicated the drawer in the palette, otherwise it is not collapsible.

When the `canvas.update()` is omitted, the animations are still processed for their delay but not drawn, which also feels a little bit "laggy". 


https://github.com/eclipse/gef-classic/assets/139544371/3eb7f1d5-13d1-4df6-b6c9-32a4f9505603

ptziegler:
Ah, so that what you mean. Thanks for the example!

Hm, it doesn't look like those animations are played at all on Linux, regardless of whether the update method is called...
I have to dig up my old Windows laptop to see what's going on here.

daell82:
No problem.

On the other hand, I was not able to reproduce the flickering issues mentioned in #127 on Windows. Maybe it could be a quick solution, if the `canvas.update()` is only performed on that platform.

azoitl:
I also noticed that with the improvements of SWT we sometimes see improvements in Draw2d too. 

@Phillipus can you please comment here as well?

Phillipus:
I don't think the changes introduced in https://github.com/eclipse/gef-classic/pull/127 have anything to do with animations. What I found was that SWT needs time to call `Display.getDefault().readAndDispatch()` to update the animation. See https://github.com/archimatetool/archi/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/Animation.java

Phillipus:
Actually, it seems that reverting #127 allows animation to work on Windows, but animation is not working on Mac and Linux which is why I modified the `Animation` class. So, even if #127 it won't fix the problem on all platforms.

Phillipus:
You could try https://github.com/eclipse/gef-classic/pull/377 to see if that works.

daell82:
That change fixes the palette animations and the animated zoom (which I use from GMF in my application). 👍 

[...]

PR #730, opened 2025-05-04
Force repaint in NativeGraphicsSource to fix broken animation

ptziegler:
When the FigureCanvas is created with SWT.DOUBLE_BUFFERED, animations are not painted correctly. This is because an instance of NativeGraphicsSource is used internally, which doesn't paint synchronously.

Because the animation is done inside the UI thread, this paint operation is only processed after the animation is done.

Resolves https://github.com/eclipse/gef-classic/issues/376


ptziegler:
This was already merged via https://github.com/eclipse-gef/gef-classic/pull/445 but later reverted due to test failures, which has been addressed with https://github.com/eclipse-gef/gef-classic/pull/728.

azoitl (approved):

PR #735, opened 2025-05-18
Avoid potential infinite loop if button is released by ActionListener

ptziegler:
If the button is release by the "action performed" event, it might happen that the released() method is called before the pressed() button is finished (i.e. via a call to Display.readAndDispatch().

Therefore, the timer that is created is never released and is executed indefinitely.

Closes https://github.com/eclipse-gef/gef-classic/issues/733


ptziegler:
This is a bigger issue for which I don't have an immediate solution. So I think the best approach is to adapt https://github.com/eclipse-gef/gef-classic/pull/730 so that the animation fix isn't applied for scrolling within the palette drawer.

ptziegler:
The Zest tests are too tightly coupled to the UI thread, which is now causing problems. I'll try to create a separate PR for that, though likely not before the RC1

azoitl:
@ptziegler as you currently working on it. Should we wait for this for M3 or not?

ptziegler:
> @ptziegler as you currently working on it. Should we wait for this for M3 or not?

Yes, but the tests are being a real pain. A lot of the implicit synchronisation needs to be done explicitly with the changes to the Animation. Please don't wait for me. I'll figure it out eventually and if it doesn't work now, I'll give up for today.

merks:
It’s fine to contribute m3 on Wednesday.

ptziegler:
Sometimes the easiest solution is the best one... Using `readAndDispatch()` is really unpredictable because you have no control over what events are processed by it and using it in the Animation class directly causes a lot of undesirable side-effects.

azoitl (approved):

PR #736, opened 2025-05-19
Convert Zest tests into propert SWTBot tests

ptziegler:
The current approach of executing the Zest-SWTBot tests is prone to timing issues when interacting with the Draw2D Animation class.

To avoid this, the test suite itself should be executed in a separate thread and all UI access should be done using the SWTBot API.

Note that SWTBot doesn't provide a Zest API, so a prototype defining the required methods has been createdas well


ptziegler:
Merging it now, given that those are only JUnit tests.

PR #742, opened 2025-05-27
Convert Flow icons to SVG

ptziegler:


azoitl:
@ptziegler I'm a bit puzzled why `*.gif` worked and `*.svg' is not. Do you have any idea?

ptziegler:
Changing the file extension to .png seems to have done the trick. I've looked at the jar and all required files seem to be included.

PR #743, opened 2025-05-27
Adjust palette cursors so that anchors are placed on even coordinates

ptziegler:
If the coordinates for the anchors contains fractions, the generated PNG will be blurry as this can't be expressed in a plain grid. So instead of e.g. (1.207, 1,207), the anchros should be (0.5, 0.5), which takes up exactly one pixel, rather than two.

Closes https://github.com/eclipse-gef/gef-classic/issues/737


ptziegler:
The main problem is that the store width is also scaled relative to the zoom. Even though it should always remain exactly one pixel wide. There is a `vector-effect:non-scaling-stroke` you can apparently set but not with Inkscape, apparently.

PR #744, opened 2025-05-27
Revert "Cleanup GEF cursors and remove some of the blockiness"

ptziegler:
This reverts commit 537b65f771330d8b7420ccb91398a25f6f419e8d.


ptziegler:
Something clearly doesn't work with the cursors yet. If they are created with 150% on a 150% monitor, the zoom is applied twice. And no matter how I configure it, the cursors always look much more blurry than they should be. So I think it's better to sit this one out for now.

ptziegler:
> It is quite interesting how often we have problems with the cursors. There is a reason that in 4diac IDE we are not using any GEF cursors at all. But I guess this is a different story.

I think a big part is because the Cursor constructor is simply unclear to me. Why does it expect an ImageData, as opposed to e.g. an ImageDataProvider? At what zoom does the image data have to be? 100%? The current display/monitor zoom? If it is 100%, how is the scaling done?

> Should we use this for RC1? Anything else missing?

I'm now effectively done with my tests. You can start the build for the RC1, if you want, otherwise I would've done that this evening.

azoitl:
Just triggered the milestone build.

PR #748, opened 2025-06-18
Remove the exclusion check for the TreeViewer

ptziegler:
The collection of IFigures that is used in the findObjectAtExcluding() of the TreeViewer is compared against an EditPart. This check will always return "false" and is therefore obsolete.

Given that a TreeViewer is based on a Tree widget and not IFigure's, specifying an exclusion set is generally not supported.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/TreeViewer.java
@@ -35,6 +35,7 @@ import org.eclipse.jface.util.TransferDragSourceListener;
 import org.eclipse.jface.util.TransferDropTargetListener;
 import org.eclipse.jface.viewers.StructuredSelection;
 
+import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.geometry.Point;
 
 import org.eclipse.gef.EditPart;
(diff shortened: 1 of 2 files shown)


ptziegler:
@azoitl Given that you cleaned up the drag/drop listeners some time ago, perhaps you have more insight into this. As far as I understand, the exclusion list will always be empty for the TreeViewer, as the EditParts implement TreeEditPart and not GraphicalEditPart.

https://github.com/eclipse-gef/gef-classic/blob/3b9b44f49ff407d9b0b307643d546654f74107d5/org.eclipse.gef/src/org/eclipse/gef/dnd/AbstractTransferDropTargetListener.java#L89-L98


Meaning that the only way to exclude elements is by moving the `excludes.contains(EditPart)` to the Conditional and not the `findObjectAtExcluding(...)`?

Or to rephrase it: We would need different implementations for the TreeViewer and the GraphicalViewer.

azoitl:
@ptziegler when I cleaned that part I stopped at the treeviewer because I never could really grasp how this ever worked. The original documentation of the exclusion set said that it contains edit parts. However all implementations of graphical viewers and the classes using them always provided and expected IFigures. Except for the Treeviewer. Therefore I decided to change it the way it is now for all none Treeviewers. I agree that for treeviewers a different implementation would be needed.

ptziegler:
> The original documentation of the exclusion set said that it contains edit parts.

Thanks for the info. That's actually the reason that brought me to this topic, because that's how it (used to) work in WindowBuilder:
image

But I also dug up an old GEF 2.1.3 SDK and even there, the exclusion set was already using IFigures. So I somewhat doubt it has even been something else. 

I'm currently playing around with the Outline viewer of the Logic editor and was wondering why the issue doesn't show up there. Because in principle, it should be possible to drag a tree item and drop it on one of its children. Turns out that the viewer contains some addition code to avoid this exact problem.

https://github.com/eclipse-gef/gef-classic/blob/3b9b44f49ff407d9b0b307643d546654f74107d5/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/edit/LogicTreeContainerEditPolicy.java#L61-L63

So yeah... I think this issue was never addressed and people just worked around it.

PR #750, opened 2025-06-25
Replacing usages of Device#getDPI

ShahzaibIbrahim:
Having the scale factor being based on the screen DPI leads to unexpected result. Having a screen dpi independent factor leads to consistent results.

### How to test

- Run this example 
- See if it doesn't break after the current changes. 
```
import org.eclipse.swt.layout.FillLayout;

import org.eclipse.draw2d.FigureCanvas;
import org.eclipse.draw2d.Graphics;
import org.eclipse.draw2d.LightweightSystem;
import org.eclipse.draw2d.SWTGraphics;
import org.eclipse.draw2d.geometry.Dimension;
import org.eclipse.draw2d.geometry.Rectangle;

import org.eclipse.gef.rulers.RulerProvider;

public class RulerFigureExample {

		Display display = new Display();
		Shell shell = new Shell(display);
		shell.setText("RulerFigure Standalone Example");
		shell.setSize(400, 100);
		shell.setLayout(new FillLayout());

		// Create FigureCanvas (Draw2D widget)
		FigureCanvas canvas = new FigureCanvas(shell);
		LightweightSystem lws = new LightweightSystem(canvas);

		// Create horizontal RulerFigure
		RulerFigure ruler = new RulerFigure(true, new RulerProvider() {
		});
		// Optionally set interval: major marks every 10 units, 5 divisions
		ruler.setInterval(10, 5);

		// Set preferred size
		Dimension prefSize = ruler.getPreferredSize(-1, -1);
		ruler.setBounds(new Rectangle(0, 0, 380, prefSize.height));

		// Set the root figure of the canvas to the ruler
		lws.setContents(ruler);

		// Add a PaintListener to the canvas's SWT control to explicitly call
		// paintFigure
		canvas.addPaintListener(e -> {
			// Create Draw2D Graphics from SWT GC
			Graphics graphics = new SWTGraphics(e.gc);
			// Manually call paintFigure on the ruler
			ruler.paintFigure(graphics);
		});

		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}
}

```

### Screenshot
image

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerFigure.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerFigure.java
@@ -13,7 +13,6 @@
 package org.eclipse.gef.internal.ui.rulers;
 
 import org.eclipse.swt.SWT;
-import org.eclipse.swt.widgets.Display;
 
 import org.eclipse.draw2d.ColorConstants;
 import org.eclipse.draw2d.Figure;
@@ -52,6 +51,7 @@ public class RulerFigure extends Figure {
 	 * paintFigure(Graphics) method.
 	 */
 	private static final int BORDER_WIDTH = 3;
+	private static final float DOTS_PER_INCH = 96.0f;
 
 	private boolean horizontal;
 	private int unit;
@@ -78,7 +78,7 @@ public class RulerFigure extends Figure {
 			case RulerProvider.UNIT_PIXELS -> dpu = 1.0;
 			case RulerProvider.UNIT_CUSTOM -> dpu = rulerProvider.getCustomRulerDPU();
 			default -> {
-				dpu = transposer.t(new Dimension(Display.getCurrent().getDPI())).height;
+				dpu = DOTS_PER_INCH;
 				if (getUnit() == RulerProvider.UNIT_CENTIMETERS) {
 					dpu = dpu / 2.54;
 				}
--- a/org.eclipse.draw2d.examples/sandbox/swt/bugs/Bug81467.java
+++ b/org.eclipse.draw2d.examples/sandbox/swt/bugs/Bug81467.java
@@ -15,7 +15,6 @@ package swt.bugs;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.printing.Printer;
-import org.eclipse.swt.widgets.Display;
 
 import org.eclipse.draw2d.PrinterGraphics;
 import org.eclipse.draw2d.SWTGraphics;
@@ -23,6 +22,7 @@ import org.eclipse.draw2d.SWTGraphics;
 public class Bug81467 {
 
 	public static void main(String[] args) {
+		final float DOTS_PER_INCH = 96.0f;
 		Font font = new Font(null, "Helvetica", 12, 0); //$NON-NLS-1$
 		Printer printer = new Printer();
 		printer.startJob("bugzilla 81467"); //$NON-NLS-1$
@@ -31,7 +31,7 @@ public class Bug81467 {
 		SWTGraphics graphics = new SWTGraphics(gc);
 		PrinterGraphics printGraphics = new PrinterGraphics(graphics, printer);
 
-		printGraphics.scale((double) printer.getDPI().x / Display.getDefault().getDPI().x);
+		printGraphics.scale(printer.getDPI().x / DOTS_PER_INCH);
 		printGraphics.setFont(font);
 		printGraphics.drawString("Hello world", 50, 50); //$NON-NLS-1$
 		printGraphics.dispose();


azoitl:
@Phillipus this may be interesting for you.

ShahzaibIbrahim:
> Can you please extract the 100 into some constant with a meaningful name and reuse this constant in all places.

Hi, I would like to do the same but honestly I don't have a very meaningful name for this constant. As I mentioned, it is a dpi independent factor. Perhaps you have something in mind?

azoitl:
I like your name proposals @ptziegler. However after spending more time today on this code I have a different concern. Is this change not breaking the idea of the ruler for inches and cms? I don't know any tool that is using inches or cm as unit for any drawing but for these the ruler would now look quite different and the units would not really match, or? How does this then effect printing?

ShahzaibIbrahim:
> > Hi, I would like to do the same but honestly I don't have a very meaningful name for this constant. As I mentioned, it is a dpi independent factor. Perhaps you have something in mind?
> 
> @ShahzaibIbrahim If I understand this correctly, you're now basically just converting between fractions and percents, rather than a DPI-dependent factors. So perhaps two constants `FRACTION_TO_PERCENT = 100.0` and `PERCENT_TO_FRACTION = 1 / FRACTION_TO_PERCENT` would do the trick? @azoitl wdyt?

Sounds good to me. Would it be okay If i place the constant in Printer class?

ptziegler:
> Would it be okay If i place the constant in Printer class?

As long as those are private field it's fine. If they are needed anywhere else later on, they can always be moved to a more central place.

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerFigure.java
```diff
@@ -78,7 +78,7 @@ protected double getDPU() {
 			case RulerProvider.UNIT_PIXELS -> dpu = 1.0;
 			case RulerProvider.UNIT_CUSTOM -> dpu = rulerProvider.getCustomRulerDPU();
 			default -> {
-				dpu = transposer.t(new Dimension(Display.getCurrent().getDPI())).height;
+				dpu = FRACTION_TO_PERCENT;
```
ptziegler, line 81: I played around with this setting and I think in terms of backwards compatibility, the value must be 96 and not 100. But then the name of the constant is misleading again... Perhaps it should then be changed to DOTS_PER_INCH?

@azoitl This does indeed break the ruler, but in a good way.

This is the ruler at 100% zoom:

image

This is the ruler at 150% zoom, using `Display.getCurrent().getDPI())`:

150%_zoom

And this is the ruler at 150% zoom, but with a fixed dpu of 96.

image

The guide should remain at 2 and 4, but currently scales with the display zoom.

ptziegler:
I'll have a look at the printer this evening.

org.eclipse.gef/src/org/eclipse/gef/internal/ui/rulers/RulerFigure.java
azoitl: @ptziegler thx for testing and explaining. with 96 this makes sense at least for Windows and Linux. Not yet sure about MacOS, which I currently can not test. 

Is this also related to #754 and the work to always draw on 100% and introduce the native zoom into the Graphics?

azoitl: Thx for the clarification. This are then good news.

ptziegler:
@ShahzaibIbrahim I've looked at your changes and added some feedback.If you update the constants then I think this will be a solid change.

org.eclipse.draw2d/src/org/eclipse/draw2d/PrinterGraphics.java
```diff
@@ -113,7 +114,7 @@ float zoomLineWidth(float w) {
 	public void setLineAttributes(LineAttributes attributes) {
 		if (attributes.style == SWT.LINE_CUSTOM && attributes.dash != null && attributes.dash.length > 0) {
 			float[] newDashes = new float[attributes.dash.length];
-			float printerDot = (float) printer.getDPI().y / Display.getCurrent().getDPI().y + 0.0000001f;
+			float printerDot = printer.getDPI().y / FRACTION_TO_PERCENT + 0.0000001f;
```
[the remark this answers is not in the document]
ptziegler: > Note though that this (wrong) behavior also appears with the 2024-12, so I think something in Draw2D is messing up the calculations.

After further testing, I noticed that printer.getDPI() always returns (600,600), regardless of the display zoom. Dividing this by 96 therefore always results in a constant. The scaling therefore seems to be the native scaling done by SWT, even though it shouldn't.

ShahzaibIbrahim:
> @ShahzaibIbrahim I've looked at your changes and added some feedback.If you update the constants then I think this will be a solid change.

From what I understand, the only requested change is to change the value of DOTS_PER_INCH from 100 to 96, is that right?

ptziegler:
> The problem that you are showing me is more of a font problem. With autoScaleOnRuntime the native zoom is updated to the current zoom where the shell is and the font are scaled accordingly but with autoScaleOnRuntime off, the nativeZoom is not adapted and hence the mismatch.

Just for clarification: I'm only testing this with a single monitor at 150% zoom. The problem is not the zoom level between different monitors, but between the monitor and the printer.

ShahzaibIbrahim:
@ptziegler so I will keep PrinterGraphics#zoomFontHeight method as is and replace the occurrences of getDPI on other places. Will try to replace the PrinterGraphics some other time.

ptziegler:
@ShahzaibIbrahim I played around with the printer a little bit more and I think the underlying issue is that information gets lost about the underlying zoom level.

Consider the following snippet and 125% zoom:
```java
package test;

import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Font;
import org.eclipse.swt.graphics.Point;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Label;

public class Snippet {
		System.setProperty("swt.autoScale.updateOnRuntime", "true");
		String text = "Hello World";
		
		Shell shell = new Shell(SWT.NO_TRIM);
		shell.setLayout(new FillLayout());
		
		Font font = new Font(display, "Helvetica", 32, SWT.NONE);
		
		GC gc = new GC(shell);
		gc.setFont(font);
		Point size = gc.textExtent(text);
		gc.dispose();
		
		Label label  = new Label(shell, SWT.NONE);
		label.setFont(font);
		label.setText(text);
		
		shell.setSize(size.x, size.y);
		shell.open();
		
		while (!display.isDisposed()) {
			display.readAndDispatch();
		}
	}
}
```

Regardless of whether `scaleOnRuntime` is enabled, both shells are exactly the same. This is a problem for Draw2D, specifically for Labels, because the size of the label is based on the size of the text (at 125% zoom, of course).

image

You can even see this in a pure SWT example. This snippet will produce the scaling issues when printing the widget. However, when you use the Printer font in the paint() method as opposed to the Display font, the scaling issue disappears.

```java
package test;

import org.eclipse.swt.SWT;
import org.eclipse.swt.events.SelectionAdapter;
import org.eclipse.swt.events.SelectionEvent;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.graphics.Device;
import org.eclipse.swt.graphics.Font;
import org.eclipse.swt.graphics.Point;
import org.eclipse.swt.layout.BorderData;
import org.eclipse.swt.layout.BorderLayout;
import org.eclipse.swt.printing.PrintDialog;
import org.eclipse.swt.printing.Printer;
import org.eclipse.swt.printing.PrinterData;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.Canvas;

public class Snippet2 {
	private static final String displayText = "Hello World";
	private static Point displaySize;
	private static Font displayFont;
	
		Shell shell = new Shell();
		shell.setLayout(new BorderLayout());
		
		displayFont = new Font(display, "Helvetica", 32, SWT.NONE);
		
		GC gc = new GC(shell);
		gc.setFont(displayFont);
		displaySize = gc.textExtent(displayText);
		gc.dispose();
		
		Canvas canvas = new Canvas(shell, SWT.NONE);
		canvas.setLayoutData(new BorderData(SWT.CENTER));
		canvas.addPaintListener(event -> paint(display, event.gc));

		Button printMe = new Button(shell, SWT.PUSH);
		printMe.setText("Print It");
		printMe.setLayoutData(new BorderData(SWT.TOP));
		printMe.addSelectionListener(new SelectionAdapter() {
			@Override
			public void widgetSelected(SelectionEvent e) {
				PrintDialog dialog = new PrintDialog(shell);
				PrinterData pd = dialog.open();
				if (pd == null) {
					return;
				}
				Printer p = new Printer(pd);

				p.startJob("Print");
				GC gc = new GC(p);
				paint(p, gc);
				gc.dispose();
				p.endJob();
			}
		});
		
		shell.setSize(300, 300);
		shell.open();
		
		while (!display.isDisposed()) {
			display.readAndDispatch();
		}
	}
	
	private static void paint(Device device, GC gc) {
		Font f = new Font(device, displayFont.getFontData()[0]);
		gc.setFont(displayFont);
		
		gc.setBackground(new Color(255, 0, 0));
		gc.fillRectangle(0, 0, displaySize.x, displaySize.y);
		gc.setForeground(new Color(0, 255, 0));
		gc.drawString(displayText, 0, 0);
	}
}
```

ptziegler (approved):
The changes to the DPI for the ruler is sound and works fine when using different zoom levels. :+1:

PR #752, opened 2025-07-01
Separate cursor icon from its decoration for the shared cursors.

ptziegler:
The way cursors are rendered on Windows is that regardless of the zoom level, the stroke width of the cursor is always 1px wide. When using an SVG, this would normally be achieved by setting the "non-scaling-stroke" vector effect.

Sadly, this effect is not supported by the jSVG rasterizer that is used by SWT and as a result, the stroke width is scaled proportionally to the zoom level. I.e. 1px at 100%, 2px at 200% and so on.

To work around this limitation, the cursor is separated from the remaining components and constructed on-the-fly, The complete cursor is then a composite of the SVG and this generated cursor.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/SharedCursors.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/SharedCursors.java
@@ -15,9 +15,12 @@ package org.eclipse.gef;
 import org.eclipse.swt.graphics.Cursor;
 
 import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.jface.viewers.DecorationOverlayIcon;
+import org.eclipse.jface.viewers.IDecoration;
 
 import org.eclipse.draw2d.Cursors;
 
+import org.eclipse.gef.internal.InternalCursor;
 import org.eclipse.gef.internal.InternalGEFPlugin;
 import org.eclipse.gef.internal.InternalImages;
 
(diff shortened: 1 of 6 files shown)


ptziegler:
A side-by-side comparision of the cursors at 400% zoom

image

image

My main motivation for doing it like this is because that's how Windows does it. And switching from the Windows cursor to our cursor is something that immediately catches the eye, because of how much thicker the lines are.

@azoitl Your thoughts on this?

ptziegler:
The bottom right of the cursor isn't very smooth. I'll want to double-check how Windows does it and update the shape. But the general idea is done.

ptziegler:
The curve at the bottom right isn't really used anymore in "modern" cursors. Instead, it's just a straight line.

image

ptziegler:
> How are you capturing your screenshots are you using a tool like Greenshot? I recently ran into an issue with Greenshot where the cursor would appear incorrectly scaled if the app was launched on a monitor with a different zoom level.

I think the tool is called [Lightscreen](https://lightscreen.com.ar/). I think the issue might be related; I changed the zoom level of my monitor after the application had been launched.

azoitl (approved):
Looks amazing. How shall we handle this in relation to PR #751.

ptziegler:
@weisJ Awesome! I played around with an older version of the cursors and I can confirm that the nested \<svg\> tag does the trick.

For this specific case, generating the cursor on the fly yields a slightly better result, but that is something I also see in Inkscape, meaning it's a problem with my SVG.

PR #754, opened 2025-07-05
Repaint client area after scrolling to account for rounding errors

ptziegler:
If the display zoom differs from 100%, scrolling the canvas might introduce visual artifacts caused by rounding errors. To account for such errors, the entire visible area needs to be repainted.

To reproduce this issue, simply take e.g. the palette and scroll up, then up again. If High-DPI scaling is enabled, the palette entries should be rendered a few pixel larger than they were before, even though in the model, they have the same size.

Closes https://github.com/eclipse-gef/gef-classic/issues/753

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -86,6 +86,10 @@ public class FigureCanvas extends Canvas {
 		getHorizontalBar().setValues(model.getValue() + hBarOffset, model.getMinimum() + hBarOffset,
 				model.getMaximum() + hBarOffset, model.getExtent(), Math.max(1, model.getExtent() / 20),
 				Math.max(1, model.getExtent() * 3 / 4));
+		// Repaint whole canvas to fix visual artifacts caused by HighDPI rounding errors
+		if (RangeModel.PROPERTY_VALUE.equals(event.getPropertyName())) {
+			viewport.repaint(viewport.getClientArea(Rectangle.SINGLETON).setLocation(0, 0));
+		}
 	};
 
 	private final PropertyChangeListener verticalChangeListener = event -> {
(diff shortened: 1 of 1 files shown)


ptziegler:
Just to give an idea how nasty the rounding errors currently are:

[Recording_2025-07-06_005014.webm](https://github.com/user-attachments/assets/673720f8-f1c3-419c-9016-b07b138c1be1)

azoitl:
I have quite some problems with the proposed fixed. This is a performance nightmare when scrolling in large diagrams. I did some work that repaints on scrolling where minimized. If that is really a problem then we should fix the expose calculation where scrolling is handled and repaint everything there.  

What configuraiton value are you using for 'swt.autoScale' we saw problems when using 'exact'. It got better with quarter.

ptziegler:
> This is a performance nightmare when scrolling in large diagrams. I did some work that repaints on scrolling where minimized. If that is really a problem then we should fix the expose calculation where scrolling is handled and repaint everything there.

That's unsurprising. But I don't really see any other way around this, until SWT is able to handle fractional values. I don't think there is much benefit in exposing any additional calculation, when this is not something the user should ever have to worry.

This PR is also primarily about playing with different ideas to handle the current state. To be honest, I'm a little bit surprised that people haven't brought up those issues yet. But it might be that they either aren't yet using the 2025-06 or it simply isn't as noticeable outside of very specific viewers.

> What configuraiton value are you using for 'swt.autoScale' we saw problems when using 'exact'. It got better with quarter.

The configuration hasn't been changed, so it should use the default value (exact). I tried out `quarter` and it feels like this makes things even worse. Though in general, both settings behave effectively the same.

azoitl:
What you could look into is the [`FigureCanvas.performScroll()`](https://github.com/eclipse-gef/gef-classic/blob/5584d36f1501fa151c3e7262f52e2260ac4f109b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureCanvas.java#L417). There we currently calculate what has to be redrawn. Here you could do a full redraw. For GTK this is any how happening. For windows this would result in a degraded scroll performance but could fix your problem.

The other option someone brought up is to to set the GC to a 100% native scaling, draw everything with 100% bit insert a scaling rootfigure that would handle the native scaling. We could even put this very low into the viewer or figure canvas so that it is automatically in for all of our users. Or we could start with a prototype just for the pallette to start with.

ptziegler:
> The other option someone brought up is to to set the GC to a 100% native scaling, draw everything with 100% bit insert a scaling rootfigure that would handle the native scaling. We could even put this very low into the viewer or figure canvas so that it is automatically in for all of our users. Or we could start with a prototype just for the pallette to start with.

That's similar to the idea I'm currently pursuing. Or rather the approach Swing has chosen. When you create a Graphics instance, it will always be created with 100% scaling, regardless of the underlying monitor. Or in other words: You have to explicitly call `scale(nativeZoom, nativeZoom)` if you want the scaling to match your monitor. Ideally, something like this could be done for the SWTGraphics as well.

azoitl:
> That's similar to the idea I'm currently pursuing. Or rather the approach Swing has chosen. When you create a Graphics instance, it will be created with 100% scaling, regardless of the underlying monitor. Or in other words: You have to explicitly call `scale(nativeZoom, nativeZoom)` if you want the scaling to match your monitor. Ideally, something like this could be done for the SWTGraphics as well.

Isn't this problematic for mouse position handling. If you do it with an scaling figure you can build on existing translateToAbsolute/translateToRelative.

ptziegler:
> Isn't this problematic for mouse position handling. If you do it with an scaling figure you can build on existing translateToAbsolute/translateToRelative.

My general idea is to disable auto-scaling in the FigureCanvas and instead do so in the EventDispatcher and UpdateManager. So for the Figures, no additional scaling needs to be done.
I think in the end, doing it either way will be roughly the same amount of work.

azoitl:
> My general idea is to disable auto-scaling in the FigureCanvas and instead do so in the EventDispatcher and UpdateManager. So for the Figures, no additional scaling needs to be done. I think in the end, doing it either way will be roughly the same amount of work.

Which would have the advantage that it would work automatically for all Draw2d applications. Curios where this will lead you.

PR #755, opened 2025-07-17
Migrate Label test case from WindowBuilder project to Draw2D

ptziegler:
This test case is used by the WindowBuilder project to verify the correctness of its forked Label class. Because this class is about to be deleted in favor of the Draw2D class, the corresponding test case should be moved to the GEF project.


ptziegler:
See https://github.com/eclipse-windowbuilder/windowbuilder/pull/1174

ptziegler:
The Maven warnings are unrelated: https://github.com/eclipse-gef/gef-classic/pull/756

azoitl (approved):

PR #756, opened 2025-07-18
Cleanup configuration of sonar-maven-plugin

ptziegler:
This change updates the plugin from 5.0.0.4389 to 5.1.0.4751 and also moves part of its configuration from the Jenkinsfile to the parent pom.

In order to avoid the "'sonar.java.binaries' is empty" warning, this property is now initialized with the project-relative "target/classes" path.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -143,14 +143,10 @@ def void mvn() {
            --no-transfer-progress \
            -DapiBaselineTargetDirectory=${WORKSPACE} \
            -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-           -Dproject.build.sourceEncoding=UTF-8 \
            -Dbuild.id=${BUILD_TIMESTAMP} \
            -Dcommit.id=$GIT_COMMIT \
            -Dbuild.type=$BUILD_TYPE \
            -Dorg.eclipse.justj.p2.manager.build.url=$JOB_URL \
-           -Dsonar.exclusions=**/.tycho-consumer-pom.xml \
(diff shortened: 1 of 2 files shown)


ptziegler:
> <pre><code>Property &#x27;env.SONARQUBE_SCANNER_PARAMS&#x27; is encrypted. The encryption of scanner properties is deprecated and will soon be removed.</code></pre>

This seems to point be something configured on the Jenkins... Either the SonarQube plugin got updated or got too old. But this environment variable is not something we set directly.

ptziegler:
I'll have to read up on this and then create a HelpDesk issue. I think the Jenkins configuration simply needs to be updated.

Jenkinsfile
```diff
@@ -143,14 +143,10 @@ def void mvn() {
            --no-transfer-progress \
            -DapiBaselineTargetDirectory=${WORKSPACE} \
            -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-           -Dproject.build.sourceEncoding=UTF-8 \
```
ptziegler, line 146: This is already set in the parent pom and therefore doesn't need to be explicitly listed.

ptziegler:
> [DEBUG] env.SONARQUBE_SCANNER_PARAMS: { "sonar.host.url" : "https:\/\/sonarcloud.io", "sonar.token" : "******"}

There's the property. But it's still not clear to me whether it's defined in the Jenkins setting or by the SonarQube Jenkins plugin...

ptziegler:
I have no clue what to do with the last warning. This should not block any other PRs, so I'm merging it now.

Issue #757, opened 2025-07-22
`IllegalArgumentException` when committing label direct edit that deactivates source `EditPart`


ddscharfe:
## Description
Directly editing a label and committing the change can lead to an `IllegalArgumentException` if the executed direct edit command causes the source `EditPart` to be deactivated.

## Stack Trace
```Java
java.lang.IllegalArgumentException
at org.eclipse.draw2d.EventListenerList.removeListener(EventListenerList.java:86)
at org.eclipse.gef.editparts.AbstractEditPart.removeEditPartListener(AbstractEditPart.java:913)
at org.eclipse.gef.tools.DirectEditManager.unhookListeners(DirectEditManager.java:437)
at org.eclipse.gef.tools.DirectEditManager.bringDown(DirectEditManager.java:117)
```

## Cause
`DirectEditManager::bringDown` is invoked twice, which results in a call to `EditPart::removeEditPartListener(null)`. If the source `EditPart` is an instance of `AbstractEditPart`, this leads to a call to `EventListenerList::removeListener(Class<T> c, null)`, which throws an `IllegalArgumentException` because the listener is `null`.

Prior to commit 8d8e63e1d5373bf3accf0a726eda91d6df4d957c, this exception did not occur due to a guard clause that returned early when the listener was null:
<img alt="Image" src="https://github.com/user-attachments/assets/709fa734-9f83-41d1-91ef-261bfeed26c5" />

## Possible Fixes
Some options to resolve this:
1. Don't throw an `IllegalArgumentException` in `EventListenerList::removeListener`
2. Add a null check before calling `removeEditPartListener` in `DirectEditManager::unhookListeners`.
3. Add a null check inside `AbstractEditPart::removeEditPartListener`.

Option *3* seems most appropriate, as it aligns with the contract defined in the `EditPart` interface, which states:
```
	/**
	 * Removes the first occurance of the specified listener from the list of
	 * listeners. Does nothing if the listener was not present.
	 *
	 * @param listener the listener being removed
	 */
	void removeEditPartListener(EditPartListener listener);
```

To reproduce the bug, see the minimal showcase:
[DirectEditManagerWithMockTest.java](https://github.com/ddscharfe/gef-classic-skija/blob/direct-edit-exception/org.eclipse.gef.tests/src/org/eclipse/gef/test/DirectEditManagerWithMockTest.java) in [direct-edit-exception](https://github.com/ddscharfe/gef-classic-skija/tree/direct-edit-exception).

[...]

PR #758, opened 2025-07-22
Make DirectEditManager more robust when brought down multiple times #757

ptziegler:
When bringDown() is called while processing a "direct edit" command, an IllegalArgumentException is thrown because the edit-part listener is removed twice.

This is because the listener is null in the second round, which is not a valid parameter to pass to the underlying EventListenerList of the edit part.

An explicit null-check has been added to the DirectEditManager to avoid this issue.


azoitl (approved):
@ddscharfe any comments from your side?

PR #760, opened 2025-07-24
Fix hyperlinks and relative image paths of documentation

ptziegler:
- When viewing the documentation from GittHub, the images are not shown as they are contained in the "guide" folder, rather than the "guide-src" folder.

- References to other documents use the name of the generated HTML file, not the AsciiDoc file. By using .adoc, this is automatically updated during the conversion.

### DIFF
--- a/org.eclipse.draw2d.doc.isv/guide-src/migration-guide.adoc
+++ b/org.eclipse.draw2d.doc.isv/guide-src/migration-guide.adoc
@@ -1,14 +1,18 @@
+ifdef::env-github[]
+:imagesdir: ../guide/
+endif::[]
+
 = Plug-in Migration Guide
 
 == Contents
 
-* link:#3.15[Migrating from 3.14 to 3.15]
+* xref:#migrating-from-314-to-315[Migrating from 3.14 to 3.15]
 
-=== [#3.15]#Migrating from 3.14 to 3.15#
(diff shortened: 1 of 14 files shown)


ptziegler:
To see the issue, just go to https://github.com/eclipse-gef/gef-classic/blob/master/org.eclipse.gef.doc.isv/guide-src/guide.adoc and see that none of the images are shown.

org.eclipse.draw2d.doc.isv/guide-src/migration-guide.adoc
```diff
@@ -1,14 +1,18 @@
+ifdef::env-github[]
+:imagesdir: ../guide/
+endif::[]
+
 = Plug-in Migration Guide
 
 == Contents
 
-* link:#3.15[Migrating from 3.14 to 3.15]
+* xref:#migrating-from-314-to-315[Migrating from 3.14 to 3.15]
```
ptziegler, line 9: That's the name generated by GitHub and the #3.15 is effectively ignored. This means clicking on the reference never jumps to the expected location.

ptziegler:
There are still some broken references because we reference e.g. _../reference/api/org/eclipse/draw2d/Figure.html_, which is something that is generated on-the-fly. I don't think there is much we can do about that.

azoitl (approved):

PR #762, opened 2025-07-24
Minor cleanup of online help

ptziegler:
This cleans up some artifacts that remained when the HTML document was converted to AsciiDoc, the broken about.html file and the broken HREF attributes in the topics xml file.


org.eclipse.draw2d.doc.isv/guide-src/guide.adoc
```diff
@@ -4,8 +4,7 @@ endif::[]
 
 = Draw2d Guide
 
-== Draw2d Programmer's Guide +
-org.eclipse.draw2d +
+== Draw2d Programmer's Guide
```
ptziegler, line 7: image

PR #763, opened 2025-07-24
Move Draw2D examples to doc project and convert to AsciiDoc

ptziegler:


ptziegler:
There are still some orphaned images in the doc-files folder that I'd like to find a new home for. Hence marked as draft.

azoitl (commented):
shouldn't we move these demo docs into the draw2d doc plugin?

ptziegler:
> shouldn't we move these demo docs into the draw2d doc plugin?

Good point. I wasn't sure where to put them exactly but adding them to the help also works. Thankfully the files are simple enough to quickly convert them to AsciiDoc.

azoitl:
> > shouldn't we move these demo docs into the draw2d doc plugin?
> 
> Good point. I wasn't sure where to put them exactly but adding them to the help also works. Thankfully the files are simple enough to quickly convert them to AsciiDoc.

Maybe in an examples subdirectory in guide?

azoitl (approved):
Cool! Thx for that work.

PR #766, opened 2025-08-04
Implement a monitor-aware lightweight system to handle HiDPI scaling

ptziegler:
This lightweight-system disables the native HiDPI scaling that is normally done by SWT and instead scales the figures via a ScalableLayeredPane.

The main purpose of this class is to reduce the noise that is introduced by rounding, errors when dealing with fractional scaling.

To test this functionality, simply pass the MonitorAwareLightweighSystem as an additional argument when constructing the FigureCanvas. See the AutoScaleExample for reference


ptziegler:
Some additional notes:

- This generally works until the user replaces the viewport. Which is what happens in the ScalableRootEditPart, so things aren't as simply for GEF.

- It would be cleaner to swap viewport and scalable pane. But then the figures are painted at 100% zoom, but only on Windows, only when scrolling and even if the SWTGraphics instance is painting the correctly scaled shapes.

akoch-yatta:
@ptziegler thanks for the update. 

@HeikoKlare and me tried it today, but immediately ran into a brickwall when trying it with an editor with a FreeformViewport e.g. the LogicEditor. We spent a bit of time with it to try to make it work and it was usable with the small changes you'll find in https://github.com/vi-eclipse/gef-classic/tree/monitor-aware-lws, but did not work in other zooms except 100%, because it runs into an endless validation loop when the freeform bounds were set and reset in scaled and unscaled values. I assume there is a proper solution possible for that current limitation?

PR #768, opened 2025-08-08
Handle HiDPI scaling in GEF diagrams

akoch-yatta:
@ptziegler For completeness I wanted to have our own state for HiDPI scaling reflected as draft PR here as well. Our approach looks more invasive and less concise as your idea, but as it currently works for all diagrams of us, we wanted to have it as a WIP state accessible as well.


ptziegler:
This looks very similar to my initial approach, where I also put all the scaling in the root figure. The only problem: This doesn't work. At least on Windows. On Linux, everything is fine.

When you scroll the canvas, the damage is repaired as if at 100% zoom, even though the coordinates that are passed to the GC instance are correctly scaled. I was not able to figure out what exactly is causing this.

Here is an example of what I mean:

https://github.com/user-attachments/assets/a6f20927-0ed5-4aed-a55c-24667bb04f9d

akoch-yatta:
As this is not the way we want to go to, I will close this PR

PR #774, opened 2025-09-04
Test JDT release property

ptziegler:


ptziegler:
This is the expected error as even if a Java 21 is used, the code should be executable with Java 17.
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.13:compile (default-compile) on project org.eclipse.draw2d: Compilation failure: Compilation failure: 
Error:  D:\a\gef-classic\gef-classic\org.eclipse.draw2d\src\org\eclipse\draw2d\graph\HorizontalPlacement.java:[312] 
Error:  	Node root = prime.nodes.getFirst();
Error:  	                        ^^^^^^^^
Error:  The method getFirst() is undefined for the type NodeList
```

merks:
@ptziegler 

It's very reassuring to see this actually tested to verify that it works as expected!

PR #775, opened 2025-09-06
GraphicalEditor registers for CommandStackEvents #21

azoitl:
GraphicalEditor still used the deprecated commandstack listener API. With previous work the new api was added now also the registration is moved.

Addresses: https://github.com/eclipse-gef/gef-classic/issues/21

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/GraphicalEditor.java
@@ -178,7 +178,7 @@ public abstract class GraphicalEditor extends EditorPart
 	 */
 	@Override
 	public void dispose() {
-		getCommandStack().removeCommandStackListener(this);
+		getCommandStack().removeCommandStackEventListener(this);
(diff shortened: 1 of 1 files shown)


ptziegler:
I remember there being a reason why I haven't done this back when I marked the `CommandStackListener` for removal, but I don't remember what exactly it was...

I think it was a weird edge-case where subclasses implement the `CommandStackEventListener`, in which case the `stackChanged()` of the subclass would be called, but not the parent class.

However, I don't see a use-case where this would make sense, or even how we would avoid such a thing if the `CommandStackListener` interface were to be removed. :shrug:

I'm tempted to go for it and see what happens. It's not a breaking change, given that subclasses simply have to call `super.stackChanged(...)`.

azoitl:
I would see this also as a very rare case. I checked GMF, Graphiti, and a bit Sirius but I haven't seen this case there. So most probably we are save.

ptziegler:
You're probably right. Let's deal with this if it actually shows up...

ptziegler (approved):

PR #778, opened 2025-09-06
Enable "deprecation" compiler warnings

ptziegler:


ptziegler:
Given that those are plenty of new warnings, I'll split this up into smaller PRs, each focusing on a single component.

PR #783, opened 2025-09-06
Implement HighDPI scaling for Zest figures

ptziegler:
This continues https://github.com/eclipse-gef/gef-classic/pull/770 and implements HighDPI scaling for Zest graphics. If the corresponding "draw2d.disableAutoscale" system property is set, a ZoomChange listener is added to the Graph that automatically scales the root figure to match the display zoom.


ptziegler:
100%
image
150%
image

PR #792, opened 2025-09-16
Fix ScrollableThumbnail offset calculation

### LINKED ISSUE #791: ScrollableThumbnail uses incorrect offset when the sourceFigure is not the viewport
The `ClickScrollerAndDragTransferrer` inside the `ScrollableThumbnail` uses incorrect offsets, causing the selection to not move to the intended location (the mouse click position). It calculates the offset based on the viewport size. However, the thumbnail can specify an additional source figure, which is not necessarily the viewport itself. Moreover, one can overwrite the `getSourceRectangle()` method of thumbnails to specify a custom region for the thumbnail, which should be taken into account.

Felix-Schmid:
Fixes #791.


azoitl (approved):

ptziegler (approved):
The change was done with 59b33cb3f17e670453f5f6d1852356e857490439 as a response to [Bug 48425](https://bugs.eclipse.org/bugs/show_bug.cgi?id=48425). I've tested it with this patch and the Logic editor still behaves correctly with negative coordinates.

It would be great if you could integrate the minor nitpick, otherwise I'd like to merge this tomorrow.

ptziegler:
@Felix-Schmid Awesome, Thanks!

PR #793, opened 2025-09-19
Bug 154145 - [Palette] PaletteEditPart should use JFace resource manager

ptziegler:
Currently the PaletteEditPart creates images and stores them in a static map, These images are never disposed. By using the JFace resource manager, they are now bound to the lifecycle of the underlying viewer.

Closes https://bugs.eclipse.org/bugs/show_bug.cgi?id=154145

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/palette/editparts/PaletteEditPart.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/palette/editparts/PaletteEditPart.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -25,6 +25,7 @@ import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.widgets.Display;
 
 import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.jface.resource.ResourceManager;
 import org.eclipse.ui.IMemento;
 
 import org.eclipse.draw2d.Border;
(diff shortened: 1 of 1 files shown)


azoitl (approved):

ptziegler:
We just discovered a bug in our product as a result of this fix. When creating an image descriptor as follows:

```
Image image = ...;
ImageDescriptor imageDescriptor = ImageDescriptor.createFromImage(image);
image.dispose();
```

One will now get an `SWTException` when using this descriptor for the palette entries. This is because the resource manager internally calls `createResource(...)` instead of `createImage()`. Latter catches the `SWTException` and returns the "missing" image instead, rather than propagating the exception.

```
org.eclipse.swt.SWTException: Graphic is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4950)
	at org.eclipse.swt.SWT.error(SWT.java:4865)
	at org.eclipse.swt.SWT.error(SWT.java:4836)
	at org.eclipse.swt.graphics.Resource.getDevice(Resource.java:167)
	at org.eclipse.jface.resource.ImageDataImageDescriptor.createResource(ImageDataImageDescriptor.java:69)
	at org.eclipse.jface.resource.DeviceResourceManager.allocate(DeviceResourceManager.java:56)
	at org.eclipse.jface.resource.AbstractResourceManager.create(AbstractResourceManager.java:92)
	at org.eclipse.jface.resource.LazyResourceManager.create(LazyResourceManager.java:88)
	at org.eclipse.jface.resource.LocalResourceManager.allocate(LocalResourceManager.java:71)
	at org.eclipse.jface.resource.AbstractResourceManager.create(AbstractResourceManager.java:92)
	at org.eclipse.gef.ui.palette.editparts.PaletteEditPart.setImageDescriptor(PaletteEditPart.java:436)
```

Though I don't think that this is something we have to handle, as the documentation for the image descriptor clearly states:

> Note that disposing the original Image will cause the descriptor to become invalid.

PR #799, opened 2025-09-21
Avoid direct access to deprecated focusPart field in RulerComposite

ptziegler:
If a "FocusIn" event is fired for the RulerComposite, the contents of the RulerViewer should be focused, unless any other part has already requested focus.

Rather than checking whether the deprecated field is checked, we can check whether the FOCUS flag has been set for the focused edit part. This flag is only set when setFocus(...) is called on the EditPartViewer and should therefore be set exactly when the local variable is set.

Conversely, the focus should be cleared by an "FocusOut" event, unless any other part has taken the focus.

Contributes to https://github.com/eclipse-gef/gef-classic/pull/778

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/rulers/RulerComposite.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/rulers/RulerComposite.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2010 IBM Corporation and others.
+ * Copyright (c) 2003, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


azoitl:
I did a quick test and didn't notice any differences. However the `getFocusEditPart()` my provide a different editpart then the check before. Does this matter?

ptziegler:
> However the getFocusEditPart() my provide a different editpart then the check before.

Yes, the method may return `focusEditPart`, `getContents()`, `getRootEditPart()` or the last element in `getSelectedEditParts()`. However, `hasFocus()` only returns `true` for the last editpart that was used in `setFocus(...)` (which also sets the `focusEditPart` field). 

So unless people are doing some very esoteric stuff, the `hasFocus()` of an editpart returns true if and only if it is the `focusEditPart`. We even go out of our way to make sure the field and the flag are in sync:

https://github.com/eclipse-gef/gef-classic/blob/7d07057efdcfff5b70e787568453c95d2df5d23c/org.eclipse.gef/src/org/eclipse/gef/SelectionManager.java#L237-L253

That said, if you were to modify the `focusEditPart` directly without calling `setFocus()` (either in the EditPartViewer or the SelectionManager) then this check won't behave the same. However, I would argue that then you have bigger problems in your viewer...

azoitl (approved):

PR #800, opened 2025-09-22
Replace RunnableChain in DeferredUpdateManager with Deque - Bug 281327

ptziegler:
In the class org.eclipse.draw2d.DeferredUpdateManager, Runnables are queued using a special inner class RunnableChain which basically produces a linked list of Runnables. When run is called on the head of the linked list, first run of its next element is called an then run on the runnable of the element itself.

This produces very large stacks, for each element in the list, a new stack frame is created. Instead of the RunnableChain construct, a simple ArrayDeque can be used as queue which will be faster and use less memory.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


ptziegler:
The change is similar to the patch provided in https://bugs.eclipse.org/bugs/show_bug.cgi?id=281327. However, I use a `Deque` instead of a `List` and keep the local variable.

In addition, the `Runnable`s in the `RunnableChain` are executed in reverse order, which isn't considered in the patch. Whether that makes a difference, I can't tell.

I also thought about using a `LinkedList` instead of an `ArrayDeque`. But I think a plain array avoids memory segmentation and is generally more efficient, even if it means that occasionally it needs to be resized.

ptziegler:
> In addition, the Runnables in the RunnableChain are executed in reverse order, which isn't considered in the patch. Whether that makes a difference, I can't tell.

Screw that... The way the RunnableChain is created means that the newest Runnable is always the head of the queue. So the reverse order is reverse, which then produces the original order.

ptziegler:
> But is it important to do this in reverse order? Because for me this would mean the newest is executed first.

As mentioned in my second comment, I misunderstood how the RunnableChain is constructed. The Runnables are added in reverse order and are also executed in reverse order. So in the actual order is the chronological order.

> Not sure if this can be done and if this leads to measurable performance changes.

Probably not. New Runnables might be added while processing the RunnableChain. If one were to re-use the same Deque, this would lead to a ConcurrentModificationException. 

You probably could work around this by using two Deques you switch between. This mechanism was added with a77068ff75eb1ee3de211f86eec4e96d14207053 in response to [Bug 77407](https://bugs.eclipse.org/bugs/show_bug.cgi?id=77407), meaning it's primarily used for UI stuff. I think creating a new Java object is a negligible overhead.

azoitl:
It took my a bit to long to answer. I only saw your second comment after I submitted my comment. With the case of having the runable in normal order wouldn't an `ArrayList` be better?

I don't think switching between two lists is worth the effort.

PR #801, opened 2025-09-24
Use Rectangle#scale in IScalablePane#getScaledRect

akoch-yatta:
This PR replaces the scaling in IScalablePane#getScaledRect for width and height of a Rectangle with using Rectangle#scale instead.

In https://github.com/eclipse-gef/gef-classic/pull/770 I stumbled open the issue. If you execute the following small Snippet

```
public class LabelIssueExample {
		Shell shell = new Shell();
		shell.setSize(400, 400);
		shell.setLayout(new FillLayout());

		FigureCanvas canvas = new FigureCanvas(shell, SWT.NONE, new LightweightSystem());
		Figure root = new Figure();
		root.setLayoutManager(new ListLayout());

		ScalableLayeredPane layeredPane = new ScalableLayeredPane(false);
		layeredPane.setScale(1.5);
		Label label = new Label("Representing a half adder");
		root.add(layeredPane);
		layeredPane.add(label);
		canvas.getViewport().setContentsTracksHeight(true);
		canvas.getViewport().setContentsTracksWidth(true);
		canvas.getViewport().setContents(root);

		shell.open();

		print(canvas.getLightweightSystem().getRootFigure(), 0);

		while (!display.isDisposed()) {
			display.readAndDispatch();
		}
	}

	private static class ListLayout extends FlowLayout {
		public ListLayout() {
			setMajorSpacing(0);
		}
	}

	public static void print(IFigure figure, int depth) {
		for (int i = 0; i < depth * 2; ++i) {
			System.out.print(" ");
		}
		System.out.println(figure + ": " + figure.getBounds());
		for (IFigure c : figure.getChildren()) {
			print(c, depth + 1);
		}
	}
}
```
 you will see
image

Left before, right this PR. Reason is that the calculated **preferred size** is different from the **size** of the label.

**Important** is to execute this on a 100% monitor, otherwise you won't run into the rounding issue with this Snippet

Using the scale method of Rectangle instead of manipulating the attributes directly does sound reasonable to me anyway, but is there a reason for the current implementation?

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/IScalablePane.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/IScalablePane.java
@@ -30,10 +30,7 @@ public interface IScalablePane extends ScalableFigure {
 
 	public default Rectangle getScaledRect(Rectangle rect) {
 		double scale = getScale();
-		rect.width /= scale;
-		rect.height /= scale;
-		rect.x /= scale;
-		rect.y /= scale;
+		rect.scale(1 / scale);
 		return rect;
 	}
 


ptziegler:
> Using the scale method of Rectangle instead of manipulating the attributes directly does sound reasonable to me anyway, but is there a reason for the current implementation?

The default method of the IScalablePane was merely moved up from the subclasses. It's been like that since the very beginning (08cb5ec8e7c248cfb585793d26373e4bc8d562c7)... Whatever reason there might've been in the past has probably been lost to time, but I would assume that this is simply an oversight that never came up until now.

azoitl:
I checked where this code originally came from. It is there since the first initial commit of GEF Classic. I only extracted it into a common interface to have only one place for changes. While I see that the proposed code is more correct from a scaling perspective it results in completely different width and height values here. I'm not sure what will be the impact on clients.

azoitl (approved):
I gave it a try with Eclipse 4diac. I couldn't see any differences. In the long run as you said it will lead to better results. So I'm in. 

I'm only thinking if the getScaledRect method now still makes sense or if we should move the code back to the two places calling it.

@ptziegler what do you think. Happy to do that.

ptziegler:
The difference should never be more than 1 pixel in each direction, which I think is within a reasonable margin. Going this way definitely beats adding a +1 everywhere and I think it's good to try it out for the M1.

> I'm only thinking if the getScaledRect method now still makes sense or if we should move the code back to the two places calling it.

The method is already part of the API, so we can't reasonably remove it. And conceptionally, I still think it's good to have a common method for both panes.

I also dug through the GEF-classic history and the scaling method originally came with 0d36f69198f4279394b93276e5a0244cf9b2ab7e, except that only the `width` and `height` were scaled. Later, scaling of the `x` and `y` coordinates was added with cc27fe172eba6ba66e47e2fac93f0667d76620af. This method was then copy-pasted to the `ScalableLayeredPane` and from there to the `IScalablePane` interface. I'm therefore fairly certain that not much thought was put into the current implementation.

azoitl:
> The difference should never be more than 1 pixel in each direction, which I think is within a reasonable margin. Going this way definitely beats adding a +1 everywhere and I think it's good to try it out for the M1.

Had to think it through a bit more and you are totally right. I think this is definitely better.

> 
> > I'm only thinking if the getScaledRect method now still makes sense or if we should move the code back to the two places calling it.
> 
> The method is already part of the API, so we can't reasonably remove it. And conceptionally, I still think it's good to have a common method for both panes.

Ah yes I did that. The danger of introducing default methods in interfaces.

PR #803, opened 2025-10-02
Update changelog and documentation regarding improved HighDPI support

ptziegler:


ptziegler:
@akoch-yatta I'll try to give your PR a try today, but I already wrote down some notes regarding the HighDPI support. If you got a minute, can you also take a look whether this matches the thought process behind your implementation?

org.eclipse.gef.doc.isv/guide-src/guide.adoc
```diff
@@ -1068,4 +1069,20 @@ connection to avoid figures. The handles for creating and moving
...
+=== [#highdpi-support]#HighDPI Support#
+
+The `*Figures*` used for visualization of the `*EditParts*` are painted using absolute coordinates. When used together
+with fractional scaling, this inevitably leads to rounding errors and visual artifacts in the editor. To compensate for
+this limitation, GEF may be used in an experimental mode, where the scaling is not done by SWT, but rather by Draw2D. 
+
+By setting the `*draw2d.disableAutoscale*` system property, auto-scaling of the `Canvas` that is hooked to the
+`*EditPartViewer*` is configured without auto-scaling. Instead, the listener hooks a `*ZoomChanged*` listener to this
+control and updates the internal `*monitorScale*` property whenever an event is fired. The `*RootEditPart*` listens to
+this property change and updates the scale of its figure accordingly.
```
akoch-yatta, line 1083: Some remarks for this part:
- How you describe the feature naming the flag `*draw2d.disableAutoscale*` sounds wrong to me. Yes, we disable it in SWT, but we enable a special scaling mechanism for draw2d, so I would propose the adapt the naming to something with enable like `*draw2d.enable...*`

MY proposal to extend the explanation here a bit

By setting the `*draw2d.disableAutoscale*` system property, the auto-scaling capability of SWT for the `Canvas` that is hooked to the `*EditPartViewer*` is disabled. Instead the scaling is done by Draw2d to prevent rounding errors. Therefor, the viewer hooks a `*ZoomChanged*` listener to the`Canvas` and updates the internal `*monitorScale*` property whenever an event is fired. The `*RootEditPart*` listens to
this property change and updates the scale of its figure accordingly.

azoitl: I think I'm confused now. I as user do I need to do anything or would it work correctly out of the box?

If something needs to be done I would suggest to change one of the examples (most probably the Logic Editor) so that we have a sample for users of GEF Classic.

ptziegler: Based on experience, I can say that people never read the changelog. Making the feature opt-in sounds reasonable, otherwise it will likely never be used. Regardless, the documentation is based on the PR, where it is still disabled by default.

I've also updated the documentation to emphasize that the scaling provided by _SWT_ is disabled by this property and instead done by _Draw2D_.

PR #805, opened 2025-10-04
Enable Draw2D auto-scaling by default

ptziegler:
This renames the "draw2d.disableAutoscale" environment variable introduced with 1bd28834343bce2bf73cbf66d9cb2accc54382f0 to "draw2d.enableAutoscale". In addition, this property now needs to be explicitly disabled.

By doing so, we drastically increase the visibility of this feature and reduce the changes that need to be done by consumer projects.


ptziegler:
The Zest test failures are fixed by #806. The ScalableLayeredPane that is used inside the PopupHelper introduces some weird little quirks that are causing issues when determining the actual popup figure. While it would be possible to adapt this in the test suite, I think it's better to clean the implementation up a little.

PR #806, opened 2025-10-05
Simplify configuration for Draw2D auto-scaling mode

ptziegler:
Enabling the Draw2D auto-scaling mode currently requires three steps:

1) Disable the native SWT auto-scaling
2) Hook a ZoomChanged listener to the underlying canvas
3) Update the zoom of the ScalableFigure

All of these steps are currently possible via individual methods in the InternalDraw2dUtils. With this change, those methods are combined into a single method, greatly reducing the amount of changes required in the calling classes.

PR #807, opened 2025-10-05
Simplify creation of PopupHelper LWS when Draw2D autoscaling is enabled

ptziegler:
When the auto-scaling property is set, a ScalableLayeredPane is injected
between the root figure and its contents.

The issue with this approach is that the contents is also stored as a
local variable of the LightweightSystem class, which is no longer
updated and therefore always null.

This pane is now embedded in a custom RootFigure subclass.


ptziegler:
Depends on #806

PR #813, opened 2025-10-07
Only configure auto-scaling for the canvas when supported by SWT

ptziegler:
The auto-scaling support in Draw2D relies on the existence of the SHELL_ZOOM key, in order to calculate the zoom factor of the widget, which was only added with the 2025-12 SWT release. Because auto-scaling is enabled by default, Draw2D always paints figures at 100% zoom when using an older Eclipse version, even if the monitor has a higher zoom level.

By checking whether this key is set, this feature is only available with the latest Eclipse release. With older versions, the native scaling functionality provided by SWT is used.

Closes https://github.com/eclipse-gef/gef-classic/issues/812

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
@@ -45,7 +45,8 @@ public class InternalDraw2dUtils {
 	private static final String DATA_AUTOSCALE_DISABLED = "AUTOSCALE_DISABLED"; //$NON-NLS-1$
 
 	private static final boolean enableAutoScale = "win32".equals(SWT.getPlatform()) //$NON-NLS-1$
-			&& Boolean.parseBoolean(System.getProperty(DRAW2D_ENABLE_AUTOSCALE, Boolean.TRUE.toString()));
+			&& Boolean.parseBoolean(System.getProperty(DRAW2D_ENABLE_AUTOSCALE, Boolean.TRUE.toString()))
+			&& SWT.getVersion() >= 4971; // SWT 2025-12 release or higher
 
 	public static boolean isAutoScaleEnabled() {
 		return enableAutoScale;


HeikoKlare:
Just a small note on this: the `NATIVE_ZOOM` data has been introduced for 2025-06 (https://github.com/eclipse-platform/eclipse.platform.swt/pull/2145). In older versions of SWT, the feature will not work at all. But in general one should be aware that using disabled autoscaling for specific contols on older releases than Eclipse 2025-12 will be highly experimental and will likely not work well for specific use cases. We are still working of finalizing proper support of the autoscale disablement in SWT, so that we will have a good state for 2025-12. For example, drag&dropping elements into autoscale-disabled controls from somewhere else does not work well (but will be fixed soon).

ptziegler:
I read this as "it would be better to only enable this feature for 2025-12 or higher"? I think consumers that care about the latest state of the HiDPI support also make sure to use the latest Eclipse release. Otherwise they seem to be fine with the rounding errors and thus should be happy with the old behavior.

There are some other glitches I noticed (figures only appearing when hovering over them, borders/connections not being drawn etc). If all of this is due to the experimental state of the `AUTOSCALE_DISABLED` key, then this sounds like the best solution.

ptziegler:
> There are some other glitches I noticed (figures only appearing when hovering over them, borders/connections not being drawn etc)

This is the reason. Weirdly enough, at 150% zoom, but not 125% zoom. This was tested with the 2025-06 but not with the 2025-12. 

```
java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4926)
	at org.eclipse.swt.SWT.error(SWT.java:4860)
	at org.eclipse.swt.SWT.error(SWT.java:4831)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1044)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:994)
	at org.eclipse.draw2d.SWTGraphics.drawImage(SWTGraphics.java:428)
	at org.eclipse.draw2d.ScaledGraphics.drawImage(ScaledGraphics.java:257)
	at org.eclipse.draw2d.Graphics.drawImage(Graphics.java:168)
	at org.eclipse.draw2d.Label.paintFigure(Label.java:506)
	at org.eclipse.draw2d.Figure.paint(Figure.java:1255)
	at org.eclipse.draw2d.Figure.paintChildren(Figure.java:1303)
	at org.eclipse.draw2d.Figure.paintClientArea(Figure.java:1336)
	at org.eclipse.draw2d.Clickable.paintClientArea(Clickable.java:325)
	at org.eclipse.draw2d.Figure.paint(Figure.java:1257)
	at org.eclipse.draw2d.Figure.paintChildren(Figure.java:1303)
	at org.eclipse.draw2d.Figure.paintClientArea(Figure.java:1336)
	at org.eclipse.draw2d.Figure.paint(Figure.java:1257)
```

PR #815, opened 2025-10-08
Override remove in PopupHelper.ScalableRootFigure #814

azoitl:
As this class is managing its children differently all methods manipulating and accessing children from the outside need to be overriden. remove() was missing resulting in acceptions.

Fixes: https://github.com/eclipse-gef/gef-classic/issues/814


ptziegler (approved):
I'm still thinking about whether there is a better way to integrate the scaling into the RootFigure...

One other issue with injecting the ScalableLayeredPane between root and contents is that the `getParent()` method for the contents (a child of the root figure) returns the scalable figure.

In any case, that's independent from this change.

ptziegler:
If there is nothing else, I'll do the respin later this evening.

PR #819, opened 2025-10-15
Introduce dedicated autoScaleLayer

azoitl:
If clients want to wrap the inner layers in other figures the fixed autoscaleeditpart listener is to limiting.

For example in 4diac IDE we are adding two more figures arround the inner layers so that we get an incremental expanding canvas similar to DrawIO. As the property listener is hard wired to the inner layer when moving 4diac IDE windows to different screens the wrong figure was scale adjusted. And as such broke our drawings. Here an example:

image

As you can see there are lines going outside the drawing area where normally elements are at the border.

With this PR I added a dedicated layer for just handling the autoscale. This would allow me to hook my two additional figures between the autoscale and the inner layer. 

@akoch-yatta  will this break the design? Does this have other issues?

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/Viewport.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/Viewport.java
@@ -27,7 +27,10 @@ public class Viewport extends Figure implements PropertyChangeListener {
 
 	/** ID for the view location property */
 	public static final String PROPERTY_VIEW_LOCATION = "viewLocation"; //$NON-NLS-1$
-	private IFigure view;
+	/**
+	 * @since 3.21
+	 */
+	protected IFigure view;
 
 	private boolean useTranslate = false;
 	private boolean trackWidth = false;
@@ -206,11 +209,11 @@ public class Viewport extends Figure implements PropertyChangeListener {
 	 * @since 2.0
 	 */
 	protected void readjustScrollBars() {
-		if (getContents() == null) {
+		if (view == null) {
 			return;
 		}
-		getVerticalRangeModel().setAll(0, getClientArea().height, getContents().getBounds().height);
-		getHorizontalRangeModel().setAll(0, getClientArea().width, getContents().getBounds().width);
+		getVerticalRangeModel().setAll(0, getClientArea().height, view.getBounds().height);
+		getHorizontalRangeModel().setAll(0, getClientArea().width, view.getBounds().width);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


org.eclipse.gef/src/org/eclipse/gef/editparts/FreeformGraphicalRootEditPart.java
```diff
@@ -88,7 +88,7 @@
  */
 public class FreeformGraphicalRootEditPart extends SimpleRootEditPart implements LayerConstants, LayerManager {
 
-	private ScalableFreeformLayeredPane innerLayers;
+	private LayeredPane innerLayers;
```
ptziegler: Shouldn't the same change also be done to the `ScalableRootEditPart`?

azoitl:
> I wonder if it would also be an option to create a `ScalableFreeformViewport` similar to what's already happening in #820 (or the PopupHelper in general)? Unless there are any reports, I would keep it simple and keep the logic (a FreeformViewport with a ScalableLayeredPane as its only child).

After looking at your proposal for the PopupHelper I find this idea quite nice. Especially as it would keep the structure as it was for existing clients. 
> 
> > Having a better API would be cool. But I don't know how this API would look like.
> 
> Any customization could be done via a `createViewport()` method, similar to what you already have with the `ScalableRootEditPart`. The only issue would appear if subclasses override the createViewport() method. But perhaps then one could add a fallback?

I'm not sure if I understand you here. The API I was talking about was that I can hook figures between the viewport and the `innerLayer`. A createViewport method would not help me.

But maybe my use-case is very special and I'm the only one doing such strange things.

ptziegler:
> The API I was talking about was that I can hook figures between the viewport and the innerLayer. A createViewport method would not help me.

I don't think it makes sense to expose API that lets you add anything between those figures. Instead I would try to make the viewport scalable to get rid of this issue. Any customization (such as hooking additional figures) would in principle be possible via a viewport subclass.

azoitl:
I reworked my pull request with a dedicted AutoscaleFreeformViewport.

 @ptziegler what do you think of it? Should we also extend it for ScalableRootEditPart?

org.eclipse.draw2d/src/org/eclipse/draw2d/Viewport.java
```diff
@@ -27,7 +27,10 @@ public class Viewport extends Figure implements PropertyChangeListener {
 
 	/** ID for the view location property */
 	public static final String PROPERTY_VIEW_LOCATION = "viewLocation"; //$NON-NLS-1$
-	private IFigure view;
+	/**
+	 * @since 3.21
+	 */
+	protected IFigure view;
```
azoitl: I had to make view somehow accessible for the subclass FreeformViewport so that the layout there works. This does not feel good.

azoitl:
I'm also very unhappy with the API extension. I would need to look into `FreeformViewport` if then scrolling is working as expected. However I would still have the problem that my use-case for which we introduced this API would not work and I would need to do very ugly hacks to hook my figures into the figure hierarchy.

Issue #818, opened 2025-10-15
Popups do not show any content on Windows


azoitl:
With the latest GEF nightly build we don't see any output in the pop ups.

When we set `draw2d.enableAutoscale=false` then pop-ups are working again as the should. 

However I have no clue how the recent changes are causing this.

CC @ptziegler @HeikoKlare @akoch-yatta

ptziegler:
If it's related to the pop-up, it is almost certainly also related to the [PopupHelperLightweightSystem ](https://github.com/eclipse-gef/gef-classic/blob/efc73d2d6c2c076ad38190de2b8c0908a0537d18/org.eclipse.draw2d/src/org/eclipse/draw2d/PopUpHelper.java#L225). Is the pop-up correctly sized, just without content? Because that could indicate an issue with the layouts.

ptziegler:
> But this would probably reintroduce the problem shown in https://github.com/eclipse-gef/gef-classic/pull/768#issuecomment-3168160441, right?

I don't think it would. You already have a similar problem in the `ScalableLayeredPane` and `ScalableFreeformLayeredPane`, where the scaling logic has simply been copy-pasted. In principle, the problem can be fixed by copy-pasting the same logic to the `ScalableRootFigure`, thus removing the need for the `scalablePane`. But I'd like to see whether there's a better approach.

azoitl:
@ptziegler we can confirm that pop-ups are now shown with contents again.

PR #820, opened 2025-10-16
Combine ScalableRootFigure and ScalableLayeredPane #818

ptziegler:
For the PopUpHelper, a custom LWS is created when auto-scaling is enabled, which injects a ScalableLayeredPane between the root figure and its contents. The problem here is that this pane is invisible to the layout manager and therefore the figure, and all its children, have an empty size.

Rather than copy-pasting the same logic that is used in both the ScalableFreeformLayeredPane and the ScalableLayered pane, the code is instead moved to the IScalablePaneHelper. The subclasses still have to override the relevant methods, but at least the implementation is now centralized.

Closes https://github.com/eclipse-gef/gef-classic/issues/818


ptziegler:
I would merge it now. I have extracted the LWS subclass as @akoch-yatta suggested and marked it as `@noreference`. It sadly can't be moved to the internal package, otherwise it wouldn't be able to access the package-private methods of the `IScalablePaneHelper`. But it should avoid the code-duplication in #822

PR #821, opened 2025-10-16
Fix shared cursors not properly scaled to monitor zoom

ptziegler:
The shared cursors are always drawn at 100% zoom, regardless of the monitor zoom.


HeikoKlare:
> Yes, even with the changes on the SWT side, the cursors still end up blurry due to the 100% zoom image-data. Thanks for testing those changes as well! They match exactly what I saw, so this looks like the best solution to this.

Just for the record: this change is in my opinion not only the best solution but actually the right thing to do 👍 The previous implementation can never have produced non-blurry results, as always the 100% image data were used and blurry scaled.

PR #823, opened 2025-10-23
Convert if/else if/else chain to switch

ptziegler:


azoitl (approved):
I was not aware that I have activated such an aggressive save action. But I like the result :-)

PR #824, opened 2025-10-23
PoC - Introduce dedicated autoScale layer

ptziegler:
This builds upon the new AutoScaleFreeformViewport added with
bbac50002c390ceca3e8da7b8ce8aea390f78f49 by perform the following
changes:

- The AutoscaleFreeformViewport is marked with @noreference
- The "view" field in the Viewport class is changed back to private.
  Instead, the FreeformFigure can be accessed via a getFreeformFigure()
  method.
- The AutoscaleFreeformViewport is only created when
  Draw2D-based auto-scaling is enabled.

### DIFF
--- a/org.eclipse.draw2d/META-INF/MANIFEST.MF
+++ b/org.eclipse.draw2d/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.draw2d;singleton:=true
-Bundle-Version: 3.20.200.qualifier
+Bundle-Version: 3.21.0.qualifier
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.draw2d,
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/AutoscaleFreeformViewport.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/AutoscaleFreeformViewport.java
@@ -1,0 +1,63 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Johannes Kepler University Linz and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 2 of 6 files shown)


ptziegler:
@azoitl I played around with this issue as well and ended up with a similar situations to yours. So instead I thought about how we minimize the exposed API and this is the result. Wdyt?

As a side-note: Both your changes and my adaptations break the feedback layer and the LED figure.
Screenshot 2025-10-25 205136

ptziegler:
@azoitl So what should we do here? Either way there are issues in some editors. I can try to brush things up so that this can be merged into the M2 or we delay everything to the M3, while I try to figure out what's wrong with the Logic editor.

> With these changes we could even get rid of the view variable as a whole or?

I don't see how that's possible. You still need it to track the contents of a plain Viewport.

azoitl:
@ptziegler I think we should still investigate a bit more. I would really love to know why these breaks are happening. Maybe we can prevent them or if not provide guidelines for how to fix it.

> >   With these changes we could even get rid of the view variable as a whole or?
>
> I don't see how that's possible. You still need it to track the contents of a plain Viewport.

`ViewPort` is adding the view as child like any other figure. `view` is just a double storage and would not be needed.

ptziegler:
Calling `scale(...)` on an `SWTGraphics` object automatically enables advanced mode. This is what's breaking the logic editor.

https://github.com/eclipse-gef/gef-classic/blob/3d8f41dbf2d80b6a9357be40994188fdd7e82b3a/org.eclipse.draw2d/src/org/eclipse/draw2d/IScalablePane.java#L47-L55

azoitl:
This now breaks all clients how set the scaled layer pane of a `ScalableFreeformLayeredPane` to false. That was the reason I set it here per default to false so that clients could choose for their own.

ptziegler:
Then how about this: We don't use scaled graphics by default and for the `ScalableFreeformGraphicalRootEditpart`, we simply pass through the `usedScaledGraphics` flag. This fixes the Logic editor and also respects the decision to only use SWTGraphics.

ptziegler:
@azoitl I think this is the best compromise we can get. Usage of the `ScaledGraphics` is disabled, unless you're using the `ScalableFreeformRootEditPart`, in which case it depends on the constructor argument. For the Logic editor, this is set to true, which is why the editor is no longer broken. For all other editors, it's whatever the consumer decided.

I've also merged both of our commits and added myself as `Co-authored-by`. I hope that's OK.

PR #825, opened 2025-10-24
Remove obsolete ExternalToolBuilder from example projects

ptziegler:
The Ant-based scripts that transform those projects into zip archives have long been superseded by Tycho, so there is no benefit in keeping them.

Together with those scripts, the .gitignore files in those projects are also deleted, which are used to exclude those generated zip files.


ptziegler:
For some time now I noticed an error, whenever I'm trying to launch the Eclipse from the GEF workspace. Turns out it's because some of the .launch files have already been deleted and the builder crashes when those projects are rebuilt while starting the IDE.

image

org.eclipse.gef.doc.isv/.externalToolBuilders/org.eclipse.oomph.version.VersionBuilder.launch
```diff
@@ -1,9 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<launchConfiguration type="org.eclipse.ant.AntBuilderLaunchConfigurationType">
-<booleanAttribute key="org.eclipse.ui.externaltools.ATTR_BUILDER_ENABLED" value="false"/>
-<stringAttribute key="org.eclipse.ui.externaltools.ATTR_DISABLED_BUILDER" value="org.eclipse.oomph.version.VersionBuilder"/>
-<mapAttribute key="org.eclipse.ui.externaltools.ATTR_TOOL_ARGUMENTS">
-<mapEntry key="release.path" value="/org.eclipse.gef.releng/oomph_version_mgmt/release.xml"/>
-</mapAttribute>
-<booleanAttribute key="org.eclipse.ui.externaltools.ATTR_TRIGGERS_CONFIGURED" value="true"/>
-</launchConfiguration>
```
ptziegler: I've never really heard about the Oomph Version Management builder. But I think it servers a similar purpose as the PDE API tool? The other launch configurations were removed with d40a6827ad9da9dd4e3c72c00f2ae57a1afcf1cc, so I think this simply as an oversight.

azoitl (approved):

Issue #831, opened 2025-10-28
"Widget is disposed" exception in FigureUtiltities.getFontMetrics


ptziegler:
Regression introduced by https://github.com/eclipse-gef/gef-classic/commit/efc73d2d6c2c076ad38190de2b8c0908a0537d18. Clients may call and cache `FigureUtilities.getFontMetrics()`. But if the Shell is disposed in the background, this instance is invalidated.

<img width="899" height="602" alt="Image" src="https://github.com/user-attachments/assets/0581cfea-f37e-4993-8c1a-759ecb492ad4" />

HeikoKlare:
One note before you or anyone starts working on this: we also experienced issues with the recreation of GCs in `FigureUtilities`. For us it's about testing with Sleak for detecting memory leaks. Since new GCs get created throughout diagram lifecycle (and the last of them survives until the end of the application lifecycle), one such `GC` is always detected as a memory leak during a test case run. One can of course try to work around that, but it's a bit messy. And it shows that the current solution is somewhere between a proper auto-closable GC only used temporarily for some `getTextExtents()` call or the like and the permanent, non-disposed GC we had before.

I thought about whether it's wouldn't be best to restore the old behavior (one static, never-disposed GC) and solve the memory leak issue on Windows which we wanted to fix with that (creating more and more SetFont operations on that GC) differently. As a concrete idea, I thought about a (performance) improvement in `GC`, which makes a `set...` operation just replace the previous one if there was no other operation in between. E.g., calling `GC#setFont()` multiple times without anything in between is not necessary to store. Only the last `setFont()` call is relevant. That would be a slight (performance) improvement for SWT/GC and implicitly resolve the memory leak issue with the `GC` in `FigureUtilities`, as the only operation called on that GC is `setFont()` and every call to it would then replace the stored operation for the previous one.

ptziegler:
I think it's a weird little quirk of SWT. You can try it with the following snippet:
```java
public static void main(String[] args) {
    Shell s = new Shell();
    GC gc = new GC(s);
    s.dispose();
    gc.dispose();
}
```

This throws the same exception when calling `gc.dispose()`... But it feels like it shouldn't...

HeikoKlare:
> But as you mentioned, this is likely something that needs to be revisited.

I started to work on a solution in SWT:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2689

[...]

PR #832, opened 2025-10-28
Dispose GC in FigureUtilities if Shell is destroyed #831

ptziegler:
If the invisible Shell created by the FigureUtilities class is disposed, any successive calls to GC.dispose() fail with a "Widget is disposed" exception.

To avoid this, the checks in getShell() and setFonts() are hardened to make sure that the previously created Shell/GC instances are still valid.

For the Shell, dispose listener is added that explicitly invalidates any previously created GC object.

Closes https://github.com/eclipse-gef/gef-classic/issues/831

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
@@ -14,6 +14,7 @@ package org.eclipse.draw2d;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import org.eclipse.swt.graphics.Color;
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
Looks sound to me as well and should properly prevent calls to `GC#dispose()` when the underlying shell is already disposed.
I have to admit that I do not fully understand how the shell can be disposed at all, but as the referenced issue shows, it can obviously happen.

ptziegler:
> I have to admit that I do not fully understand how the shell can be disposed at all, but as the referenced issue shows, it can obviously happen.

While it's a slightly artificial use-case where this is happening, you can iterate over `Display.getShells()` and then dispose the invisible one created by `FigureUtilities`.

PR #834, opened 2025-10-30
Prevail rounding precision in ScaledGraphics

akoch-yatta:
With multiple scaled layers with scaled graphics some methods will do destructive rounding before passing it to the next layer.  This PR addresses one of the affected methods, `drawLine`. 

The effect is detectable with the standard grid, where grid can alternate between different sizes:
Screenshot 2025-10-30 152148
Screenshot 2025-10-30 152156

Each grid cell should be 124px instead, because the grid size in this example is 25pt. Layer one has 2.5 scale and layer two has 2.0 scale, so 25 * 2.5 * 2.0 = 125px - 1 px for the grid line = 124px

The tests in this PR test a scenario like this. Only missing case in this scenario is if `ScaledGraphics#drawLine(Point, Point)` is called with a normal `Point` instead of a `PreecisionPoint`. I did not want to always wrap it hear, but would propose to use something like `Translatable#toPrecision` from [this branch](https://github.com/eclipse-gef/gef-classic/compare/master...vi-eclipse:gef-classic:precision-translation) and adapt this place then.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ScaledGraphicsTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ScaledGraphicsTest.java
@@ -1,0 +1,1266 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.draw2d.test;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Arrays;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.Path;
+import org.eclipse.swt.graphics.PathData;
+import org.eclipse.swt.widgets.Display;
+
+import org.eclipse.draw2d.SWTGraphics;
+import org.eclipse.draw2d.ScaledGraphics;
+import org.eclipse.draw2d.geometry.Point;
+import org.eclipse.draw2d.geometry.PointList;
+import org.eclipse.draw2d.geometry.Rectangle;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class ScaledGraphicsTest {
+
+	static Stream<Arguments> singleValueTestCombinations() {
+		int[] inputs = { 5, 7, 10, 17, 20 };
+		int[] monitorZooms = { 100, 125, 150, 175, 200 };
+		int[] diagramZooms = { 100, 150, 200, 250, 300, 400 };
+
+		return Arrays.stream(inputs).boxed()
+				.flatMap(source -> Arrays.stream(monitorZooms).boxed().flatMap(monitorZoom -> Arrays
+						.stream(diagramZooms).mapToObj(diagramZoom -> Arguments.of(source, monitorZoom, diagramZoom))));
+	}
+
+	@Test
+	@SuppressWarnings("static-method")
+	public void testDrawLineForRegression() {
+		RecordingSwtGraphics swtGraphics = executeTranslatedWithOneLayer(200, 250,
+				scaledGraphics -> scaledGraphics.drawLine(new Point(5, 5), new Point(5, 5 + 20)));
+		validateDrawLine(swtGraphics, new Point(30, 30));
+	}
+
+	@ParameterizedTest
+	@MethodSource("singleValueTestCombinations")
+	@SuppressWarnings("static-method")
+	public void testDrawLineWithInt(int source, int monitorZoom, int diagramZoom) {
+		ScaledGraphicsValidation validation = new ScaledGraphicsValidation(monitorZoom, diagramZoom);
+		validation.execute(scaledGraphics -> scaledGraphics.drawLine(source, source, source, source + 5));
+	}
+
+	@ParameterizedTest
+	@MethodSource("singleValueTestCombinations")
+	@SuppressWarnings("static-method")
+	public void testDrawLineWithIntTranslated(int source, int monitorZoom, int diagramZoom) {
+		ScaledGraphicsValidation validation = new ScaledGraphicsValidation(monitorZoom, diagramZoom);
+		validation.executeTranslated(scaledGraphics -> scaledGraphics.drawLine(source, source, source, source + 10));
+	}
+
+	@ParameterizedTest
+	@MethodSource("singleValueTestCombinations")
+	@SuppressWarnings("static-method")
+	public void testDrawLineWithPoint(int source, int monitorZoom, int diagramZoom) {
+		ScaledGraphicsValidation validation = new ScaledGraphicsValidation(monitorZoom, diagramZoom);
+		validation.execute(
+				scaledGraphics -> scaledGraphics.drawLine(new Point(source, source), new Point(source, source + 15)));
+	}
+
+	@ParameterizedTest
+	@MethodSource("singleValueTestCombinations")
+	@SuppressWarnings("static-method")
+	public void testDrawLineWithPointTranslated(int source, int monitorZoom, int diagramZoom) {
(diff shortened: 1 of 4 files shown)


azoitl:
For me `ScaledGraphics` is deprecated. It has so many problems and I would love to get rid of it and not further patch it.

org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ScaledGraphicsTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+		ImageGcDrawer imageGcDrawer = (gc, width, height) -> {
+			SWTGraphics graphics = new SWTGraphics(gc);
+			ScaledGraphics scaledGraphics = new ScaledGraphics(graphics);
+			scaledGraphics.scale(2);
+			ScaledGraphics scaledGraphics2 = new ScaledGraphics(scaledGraphics);
+			scaledGraphics2.scale(2.5);
+			scaledGraphics2.drawLine(15, 0, 15, 20);
+			scaledGraphics2.drawLine(20, 0, 20, 20);
+		};
```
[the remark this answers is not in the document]
akoch-yatta: Yes, probably. Saw the test fails of course, but didn't want to adapt until we agree on a solution. But I will have a look at those tests.

akoch-yatta:
> But I'm not a huge fan of removing the final modifier from `drawLine()`, because now you're able to have a different implementation for `drawLine(Point, Point)`, and `drawLine(int, int, int, int)`. Even though I'm more concerned about the inverse order of invocation.

We could move final to the methods in ScaledGraphics and SWTGraphics.

What worries you about the inverse invocation order? People who have custom subclasses, call the (Point, Point) method, but have an overwritten implementation for the (int, int, int, int) method? Could that break somehow?

> 
> Wouldn't the issue exist for all draw methods? Because from my understanding, the issue is caused by the rounding error when having two nested `ScalableGraphics`. So even if the rounding is fixed for `drawLine`, you would still have it for e.g. `drawOval`, wouldn't you?

Yes, definitely. I could have stated that in the PR description that this will be necessary... So if we have a mergeable for drawLine, we need to apply this pattern to the other places as well. I started with `drawLine` because there is this existing observable issue caused by it.

ptziegler:
> With the other fixes for rounding problems that we have done could we now per default deactivate usage of ScalableGraphics in the AutoscaleLayer? Then ScalableGraphics would only handle the last leg and the impact would be the same as before.

Generally yes, but not without additional adaptations. The problem with `SWTGraphics` as opposed to `ScaledGraphics` is that it requires advanced mode. This can and will cause issues with e.g. transparency, so I don't think it's a good idea for the M3.

ptziegler:
> What worries you about the inverse invocation order? People who have custom subclasses, call the (Point, Point) method, but have an overwritten implementation for the (int, int, int, int) method? Could that break somehow?

The `Graphics` class shows a clear intent that `drawXxx(Point)` is supposed to call `drawXxx(int)`. I can't predict whether this could break something and I'm therefore hesistant to violate such an assumption.

Moreover, it sounds like this issue only appears when working with nested `ScaledGraphics`. So I'd like to keep a fix limited to this class, rather than having it also leak into the `Graphics` class.

Though I tend to agree with @azoitl that it's worth considering to [deprecate](https://github.com/eclipse-gef/gef-classic/wiki/Bugzilla-%E2%80%90-Draw2D#bug-442442---deprecate-scaledgraphics-make-native-scaling-the-default) (not remove!) this class, rather than to fix its numerous issue.

azoitl:
> > Why not do both - at least fix, what is somehow easily fixable? But if you deprecate ScaledGraphics, it cannot be the default anymore like it currently is.
> 
> Deprecating `ScaledGraphics` and fixing those rounding errors are not mutually exclusive. As said: I don't have a strong opinion on these changes. As long as it doesn't modify the `Graphics` class, they are fine by me.

You are right as long as we have `ScaledGraphics` we also need to ensure that any changes we do are not breaking it. Or as in this case result in bad render output. I will in parallel prepare a PR for deprecating `ScaledGraphics`.

akoch-yatta:
I adapted the PR "a little". I would add an additional commit for each method in ScaledGraphics I adapt - I can squash that of course, if you prefer that.

I extended the ScaledGraphicsTests class to do the following:
- add one regression test with fixed values to check that the existing scaling with one layer is not be affected by the changes
- execute parameterized tests with a lot of value/monitor zoom/diagram zoom combinations for each methods, that compare the result if diagram zoom and monitor zoom would be on one layer vs. on two layers. If you remove the adaptions in ScaledGraphics you will see some of the tests will fail because of the rounding errors.

org.eclipse.draw2d/src/org/eclipse/draw2d/ScaledGraphics.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 Yatta and others.
```
ptziegler: You can't remove someone elses copyright: https://www.eclipse.org/projects/handbook/#legaldoc-faq
> How does the copyright header change when we accept a change to an existing file?
> 
> You have some options.
> 
> If the current copyright statement indicates copyright of a single entity, and the contribution comes from a different entity, then you can generally append "and others" to the existing copyright statement.
> ```
> /********************************************************************************
>  * Copyright (c) 2007 Milo Bloom and others
>  *
>  ...
> ```
> Or, you can add a separate line for the new copyright holder (with the year of the new contribution).
> ```
> /********************************************************************************
>  * Copyright (c) 2007 Milo Bloom
>  * Copyright (c) 2023 Mike Binkley
>  *
>  ...
> ```
> Both options are acceptable, so you can pick whichever works best for the situation.

org.eclipse.draw2d/src/org/eclipse/draw2d/ScaledGraphics.java
```diff
@@ -985,6 +1063,14 @@ private Rectangle zoomRect(int x, int y, int w, int h) {
...
 	}
 
+	private Rectangle zoomPrecisionRect(double x, double y, double w, double h) {
+		tempRECT.x = (int) (Math.floor(x * zoom + fractionalX));
+		tempRECT.y = (int) (Math.floor(y * zoom + fractionalY));
+		tempRECT.width = (int) (Math.floor(((x + w) * zoom + fractionalX))) - tempRECT.x;
+		tempRECT.height = (int) (Math.floor(((y + h) * zoom + fractionalY))) - tempRECT.y;
+		return tempRECT;
+	}
+
```
ptziegler: You're doing this calculation quite often. Can't you just use a `PrecisionRectangle` here and then access the `x`, `y`, `width`, `height` fields for the int-based operations and `preciseX()`, `preciseY()`, `preciseWidth()` and `preciseHeight()` methods for the double-based operations?

This also makes sure that the rounding from double to int is consistent across the board.

org.eclipse.draw2d/src/org/eclipse/draw2d/ScaledGraphics.java
```diff
@@ -927,6 +1086,14 @@ private Rectangle zoomFillRect(int x, int y, int w, int h) {
...
 	}
 
+	private Rectangle zoomPrecisionFillRect(double x, double y, double w, double h) {
+		tempRECT.x = (int) (Math.floor(x * zoom + fractionalX));
+		tempRECT.y = (int) (Math.floor(y * zoom + fractionalY));
+		tempRECT.width = (int) (Math.floor((x + w - 1) * zoom + fractionalX)) - tempRECT.x + 1;
+		tempRECT.height = (int) (Math.floor((y + h - 1) * zoom + fractionalY)) - tempRECT.y + 1;
+		return tempRECT;
+	}
+
```
ptziegler: This method is identical to `zoomFillRect`. You can just remove `zoomFillRect` and update the method name in `fillGradient`. Java will do the conversion from `int` to `double` automatically.

ptziegler:
I think it's easier if I say it like this, rather than commenting on the individual changes. I'm not a huge fan of all the code duplication when calculating the double-precision coordinates. 

Instead, what about this: You have to temporary rectangles in ScaledGraphics: `tempRECT` of type `Rectangle` and `tempPrecisionRECT` of type `PrecisionRectangle`. When calling e.g. `zoomClipRect`, `tempRECT` is filled and when calling `zoomClipPrecisionRect`, `tempPrecisionRECT` is filled. 

Based on whether the local `graphics` instance is of type `ScaledGraphics` either one or the other is used. For example:
```java
private void clipRect(double x, double y, double width, double height) {
    if (graphics instanceof ScaledGraphics scaledGraphics) {
        Rectangle z = zoomClipPrecisionRect(x, y, width, height);
        scaledGraphics.clipRect(z.preciseX(), z.preciseY(), z.preciseWidth(), z.preciseHeight());
    } else {
        graphics.clipRect(zoomClipRect(x, y, width, height));
    }
}
```

For reference, this is what I ended up with:
[ScaledGraphics.java.txt](https://github.com/user-attachments/files/23589724/ScaledGraphics.java.txt)

What do you think?

ptziegler:
I also ran the test locally but I'm running into failures for `testDrawPath` and `testFillPath`. Am I missing something? Because I don't see why they should be affected by this change.

image

PR #835, opened 2025-10-30
Revert GC recreation in FigureUtilities

HeikoKlare:
With commit efc73d2d6c2c076ad38190de2b8c0908a0537d18, GC disposals and recreations were introduced in FigureUtilities. They used to fix a memory leak caused by a long-living GC, as the operations executed on such a GC were permanently stored in the GC instance (on Windows). That change introduced two issues:
- In case a subclass of FigureUtilities performs further modifications to the used GC, those modifications will be lost upon next call of FigureUtilities#setFont(). This may be considered a breaking API change. While creating such subclasses may be questionable, it's also the only reason for keeping a static GC at all instead of using auto-disposed GCs for the duration of the individual method calls.
- Resource leak tests became more difficult, as every test for a diagram will usually leave behind a new GC (handle), precisely the last GC created by the FigureUtilities upon setting a font during diagram usage. That GC is detected as a leak and must somehow be excluded from leak detected, such as using Sleak.

With a recent improvement to SWT, the GC implementation on Windows is not prone to a memory leak when using a long-living GC on which only a single operation (in this case setFont()) is performed repeatedly anymore. Due to the mentioned issue and because the fix became unnecessary with SWT enhancements, this reverts the previous change to FigureUtilties.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
@@ -33,7 +33,6 @@ public class FigureUtilities {
 
 	private static final float RGB_VALUE_MULTIPLIER = 0.6f;
 	private static GC gc;
-	private static Shell shell;
 	private static Font appliedFont;
 	private static FontMetrics metrics;
 	private static Color ghostFillColor = new Color(null, 31, 31, 31);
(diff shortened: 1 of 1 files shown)


HeikoKlare:
> if I understand right if by saying subclasses may perform modifications to GC wouldnt this also include operations being applied, if subclasses performs operations this might still cause the memory leak. as this violates the case for the newly introduced ReplaceableOperation.

Yes, that's absolutely correct. But it's probably better if it works at all (even with a small leak) than if it would not work at all. And as you point out: this is probably not a reasonable use case anyway.
So for me the focus of this PR is on simplifying the behavior again by not having freqeuently replaces static resources, which make leak detection difficult. The breaking change I pointed to may rather be considered a side note.

arunjose696:
The change makes sense to  allow memory leak detection. Approving this as its just a revert to earlier behaviour and the memory leak responsible for the original change is now worked around with latest improvement in SWT.

PR #840, opened 2025-11-03
Add unit test for rounding errors with scaled layers

ptziegler:
Translating coordinates in GEF is done using integer-based arithmetic, which inevitably leads to rounding errors when combined with fractional scaling.

This test case simulates such an environment and tests the rounding error when using the basic geometric classes "Point", "PointList", "Rectangle" and "Dimension".

Instances of those classes are converted from relative to absolute and then back to relative coordinates, which should produce the input argument.

Relates to https://github.com/eclipse-gef/gef-classic/issues/829


ptziegler:
I've added a separate commit where the Figure is storing its bounds as a `PrecisionRectangle`. The `getSize()` and `getLocation()` methods have been overridden to return a `PrecisionDimension` and  `PrecisionPoint`, respectively.

ptziegler:
@azoitl I think this PR is effectively obsolete and superceeded by #846 and #849

PR #844, opened 2025-11-05
Harmonize calculation of int-width and int-height in PrecisionRectangle

ptziegler:
This is a follow-up to 116c3dd2d5c73db11d2d3702820dc881d274a6b8, due to which the scale() method was overriden, similar to the performScale() method.

When scaling a PrecisionRectangle, we need to make sure that its int-representation is identical to a plain Rectangle, which be off by one pixel compared to its double-representation.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionRectangleTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionRectangleTest.java
@@ -13,10 +13,6 @@
 
 package org.eclipse.draw2d.test;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


ptziegler:
Test failure documented in https://github.com/eclipse-gef/gef-classic/issues/784

ptziegler:
> What we could do here (as an intermediate solution) is to implement the modified rounding behavior only in the Figure#fromPreciseShape method (for the precision of translateToAbsolute/translateToRelative in case the original types are wrapped), so that for this specific (and completely hidden) use case we have the "correct" behavior and for everything else we for now stick to the existing behavior.

I thought about that too, but I really don't want to add such an implementation detail to the `Figure` class. At least for now I would just wait and see whether this causes actual issues before putting more effort into this.

PR #845, opened 2025-11-05
Override getSize() and getLocation() in PrecisionRectangle

ptziegler:
Both methods should return a precise Dimension and Point, respectively, to avoid rounding errors when transforming them.


ptziegler:
> Apart from the changed behavior that consumers will face, there are actual contracts that I would expect to hold that are broken now.

And I think that's the important part which is why I agree that this is was not the correct approach. Though note that even the current implementation has broken contracts, which led to this PR in the first place.

For example what is covered by the test case:
`r.getPreciseCopy().getLocation().scale(1.153), r.getPreciseCopy().scale(1.153).getLocation()`

If I have a `PrecisionRectangle` and want to scale its (precise) `x` and `y` coordinates, I would expect the scaling to be done with double-precision, not integer-precision (i.e. via a `PrecisionPoint` and not a normal `Point`). But fixing such an assumption should obviously not break other assumptions.

This should be addressed by https://github.com/eclipse-gef/gef-classic/pull/858

PR #857, opened 2025-11-17
Restore PrecisionRectangle substitutability for Rectangle

HeikoKlare:
Recent changes to PrecisionRectangle adapting how the integer-based values are calculated and how size and location of a rectangle is derived broke the substitution principle with respect to the Rectangle superclass and may also break existing consumers due to the changed behavior of API methods.

This change reverts the changes to the necessary extent such that the existing behavior is restored. It also adds a test for the substitutability of PrecisionRectangles with Rectangles.

The added test methods testScaleLocation() and testScaleDimension() are removed because scaling results for context-free rounded sizes or locations are obviously different from scaling a rectangle and then investigating its context-dependent rounded size and location values, so the different results are expected.

See also:
- https://github.com/eclipse-gef/gef-classic/pull/844#issuecomment-3541357947
- https://github.com/eclipse-gef/gef-classic/pull/844#issuecomment-3541877516
- https://github.com/eclipse-gef/gef-classic/pull/845#issuecomment-3541826688

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionRectangleTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionRectangleTest.java
@@ -119,7 +119,7 @@ public class PrecisionRectangleTest extends BaseTestCase {
 		assertEquals(91, r.preciseWidth(), 0);
 		assertEquals(91, r.preciseHeight(), 0);
 		// Different rounding behavior between Rectangle and PrecisionRectangle.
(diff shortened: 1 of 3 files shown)


HeikoKlare:
> Awesome! Thanks a bunch 😄

Thank you, @ptziegler (and of course also @azoitl), for taking care of all the reports and proposals so quickly and frequently! That's highly appreciated 🥇 
So great to see GEF Classic moving foward, allowing us to contribute to and consume GEF almost without any necessity for us to patch it for our needs.

PR #858, opened 2025-11-17
Don't override getSize() and getLocation() in PrecisionRectangle

ptziegler:
The PreciseDimension/PrecisePoint might have a different rounding behavior. Therefore a call to e.g. r.getSize().width might be different from r.width.

Instead, two new preciseSize() and preciseLocation() methods have been added.

See https://github.com/eclipse-gef/gef-classic/pull/845#issuecomment-3541826688

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
@@ -261,30 +261,12 @@ public final class PrecisionRectangle extends Rectangle {
 	}
 
 	/**
-	 * @see org.eclipse.draw2d.geometry.Rectangle#getLocation()
-	 * @since 3.21
-	 */
-	@Override
-	public PrecisionPoint getLocation() {
-		return new PrecisionPoint(preciseX(), preciseY());
-	}
-
-	/**
 	 * Returns a precise copy of this.
 	 *
 	 * @return a precise copy
 	 */
 	public PrecisionRectangle getPreciseCopy() {
 		return new PrecisionRectangle(preciseX(), preciseY(), preciseWidth(), preciseHeight());
-	}
-
-	/**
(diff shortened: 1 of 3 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionRectangle.java
```diff
@@ -260,15 +260,6 @@ public Rectangle getCopy() {
...
 	}
 
-	/**
-	 * @see org.eclipse.draw2d.geometry.Rectangle#getLocation()
-	 * @since 3.21
-	 */
-	@Override
-	public PrecisionPoint getLocation() {
-		return new PrecisionPoint(preciseX(), preciseY());
-	}
```
HeikoKlare: It would probably be fine to keep the location as precise type here, as (usually) rounding behavior for the location of a rectangle and rounding a point representing its topleft corner will usually be the same. Still it's probably best to just properly split it up and preserve the existing methods as they are.

ptziegler: The problem is that we don't really know all situations in which the point may now be used.

When you just call `getLocation()`, the `x` and `y` coordinates are almost guaranteed to be the same (with an accuracy of 1e-9 because one rounds using `Math.floor(...)`, the other using `PrecisionGeometry.doubleToInteger(...)`).

But what if you transform this point? For example `getLocation().scale(0.2).scale(5)` will behave differently for a `Point` and a `PrecisionPoint` because former discards the fractional value. It's effectively the same issue you experienced in https://github.com/eclipse-gef/gef-classic/issues/829.

So late into development it's simply too risky.

HeikoKlare (approved):
Gave this some testing together with https://github.com/eclipse-gef/gef-classic/pull/857 in our product (checking some of the tests I saw failing because of the recent adaptations) and with those two changes applied, the tests work fine again.

PR #863, opened 2025-11-19
Make IFigure#translate...() results backward compatible for Rectangle

HeikoKlare:
The wrapping of Rectangles into their precision representations when calling IFigure#translate...() methods leads to results that are incompatible with pre-existing behavior when only a single scaled layer is involved. The reason is the different rounding applied in a basic Rectangle compared to a PrecisionRectangle. Recent attempts to adapt the PrecisionRectangle behavior failed because they broke backward compatibility. But without streamlining the rounding of Rectangle and PrecisionRectangle, the results of IFigure#translate...() become incompatible.

For that reason, this change adapts the unwrapping of PrecisionRectangle to Rectangle in IFigure#translate...() to use the same rounding as applied in Rectangle to ensure backward compatibility of the method's behavior.

This conforms to what I already proposed in https://github.com/eclipse-gef/gef-classic/pull/844#issuecomment-3543090778. I considered it an optional opportunity for improvement there but turns out that without that change we break backward compatibility.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionTests.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionTests.java
@@ -18,8 +18,8 @@ import org.eclipse.draw2d.FigureUtilities;
 import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.ScalableLayeredPane;
 import org.eclipse.draw2d.geometry.PointList;
+import org.eclipse.draw2d.geometry.PrecisionDimension;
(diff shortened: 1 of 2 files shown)


pcdavid:
> > FWIW, I've not yet been able to dig into it, but something in recent GEF milestones broke several [GMF Runtime tests](https://ci.eclipse.org/gmf-runtime/job/gmf-runtime-all).
> 
> The tests execute a `translateToAbsolute` at the line where they fail, so likely it is related the issue we discuss here.

This PR seems to have fixed the issues I saw in the GMF Runtime tests: https://ci.eclipse.org/gmf-runtime/job/gmf-runtime-all/270/

They are now all passing when using the latest GEF nightly.
Thanks.

PR #864, opened 2025-11-19
Only use double-precision translation when HighDPI scaling is enabled

ptziegler:
When the "draw2d.enableAutoscale" system property is enabled, additional (scalable) layers are injected into the IFigure hierarchy to take the monitor zoom into consideration.

As a result of this, translating geometric shapes may cause additional rounding errors, which are the result of translating integer-values with an additional, fractional zoom level.

To work around this problem, the "useDoublePrecision()" method was added with 5cfe5b7587123e68c2077b6dbec4e5657316d458, which converts the geometric shapes to their precise variants. This conversion is generally not backwards compatible.

To make sure clients can fully opt-out of this new behavior, this conversion should only be done when the "draw2d.enableAutoscale" is set to true, meaning only when these new layers are added, that require this extended translation logic.

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionTests.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/PrecisionTests.java
@@ -13,6 +13,8 @@
 
 package org.eclipse.draw2d.test;
 
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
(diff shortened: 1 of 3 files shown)


ptziegler:
@azoitl I'd like to also include this one in the 3.26.1. If there is anything else we've missed in the precision classes, it allows us to also turn off the "precise" translation. It also isn't needed for Linux/MacOS as there we don't have fractional scaling.

PR #869, opened 2025-11-20
Embed tracker for dragging flyout palette composite into control

HeikoKlare:
The tracker for dragging the flyout palette is currently instantiated on the display rather than the control itself. This leads to offsets with monitor-specific scaling on Windows, as the display is lacking the monitor-specific zoom context.

This change adapts the tracker to operate on the flyout palette control rather than the display. It simplifies the implementation as the control/display coordinate conversions can be removed.

Fixes https://github.com/eclipse-gef/gef-classic/issues/868

### Before
flyout_broken

### After
flyout_fixed

My only concern is that I wonder why the implementation used the (more complex) logic via the display at all and did not just embed it into the view.

Very seldom I see that palette is drawn over the tracker border
image
which then may also leave behind an artifact of the border aftwards
image
which will just disappear when any redraw is triggered (e.g., by hovering over it).

Maybe that was a reason or maybe the implementation even goes back to a time where trackers did not work (well) on controls.

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
@@ -951,8 +951,8 @@ public class FlyoutPaletteComposite extends Composite {
 			if (paletteContainer.getVisible()) {
 				bounds = bounds.union(paletteContainer.getBounds());
 			}
-			final Rectangle origBounds = Display.getCurrent().map(flyout, null, bounds);
-			final Tracker tracker = new Tracker(Display.getDefault(), SWT.NULL);
+			final Rectangle origBounds = bounds;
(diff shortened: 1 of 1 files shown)


ptziegler:
> My only concern is that I wonder why the implementation used the (more complex) logic via the display at all and did not just embed it into the view.

Originally, that's how the tracker was used, but this was later changed with 2c1baba3d9ddda6bb9eb63724182e5edd41e4ad7. 

Apparently as part of a fix for [Bug 66201](https://bugs.eclipse.org/bugs/show_bug.cgi?id=66201) - Palette state is not maintained when switching from the view to the flyout

> The order in which notification is fired has changed.  Earlier, the workbench 
used to fire perspectiveChanged first and dispose the page later.  Now, it 
disposes the page before firing perspectiveChanged notification.  As a result, 
the palette state from the view is not transferred to the palette (although it 
works the other way around).

But this is about the palette view. I'm not sure how or why this has anything to do with the drag & drop feedback.

Issue #872, opened 2025-11-25
Custom cursors are scaled twice again


ptziegler:
https://github.com/user-attachments/assets/36ddbd50-0258-4495-895f-70a90b6ff034

HeikoKlare:
It seems like Windows is not linearly scaling the OS cursor. For me, the OS cursor has the same size at 100% and 125%. It becomes larger at 150%. The custom cursors in SWT scale linearly with the monitor zoom.
I was not aware of this behavior of OS cursors. I always assumed them to scale linearly. 

If we want to fix this, we would have to implement a specific cursor autoscaling mechanism in SWT that adheres to this specific scaling behavior of the OS.

ptziegler:
In a similar fashion, you can also change the cursor size directly. Which is not supported at all by SWT.

https://github.com/user-attachments/assets/f5319dbf-c2c7-4aab-9818-7808b72c9ffa

HeikoKlare:
> You're right. The cursor stays the same size, both at 100% and 125% zoom. So it's only updated in 50% increments? That is really weird.

That's how I understand the behavior and, yes, that's really weird. And I would bet that such a behavior is not even consistent throughout Windows versions.

> In a similar fashion, you can also change the cursor size directly. Which is not supported at all by SWT.

That's strange as we explicitly implemented support for this recently:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2493

I am quite sure that we also tested that with GEF (and in particular with GEF as that's our primary use case for custom cursors).
Seems like something is missing in the current SWT implementation.

HeikoKlare:
I created:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2811

I think we should bring it into the upcoming release as in my opinion the change is safe and would improve user experience in particular for (but not exclusive to) GEF. @ptziegler do you agree?

HeikoKlare:
I see. One question that comes up for me is how this behavior is on the other operating systems. Do they just scale according to `DPIUtil#getDeviceZoom()` so the line width here always needs to be 1 unless you have an accessibility scale factor on Windows? We could probably provide a method such as `getPointerSizeScaleFactor()` as API, but you would have to have quite some knowledge about how the cursors behave to make proper use of it. And there are probably other consumers that want to draw cursors on their own and then need to know how to properly scale line widths.

[...]

PR #873, opened 2025-11-25
Create custom cursor for 100%, 150% and 200% zoom #872

ptziegler:
At least on Windows and Linux, the operating system doesn't seem to linearly scale the system cursor.

Instead, predefined cursor images are used for:

- ( ..., 150%) - Cursor at 100% monitor zoom
- [150%, 200%) - Cursor at 150% monitor zoom
- (200%,  ...) - Cursor at 200% monitor zoom

SWT is linearly scaling the cursor icon. So if at e.g. 125% zoom, the cursor inside the widget will be scaled to 125% zoom but is not scaled outside the widget.

To account for this, the SWT-based cursor is replaced with three cursor SVGs, one for each of those zoom levels. Note that the images are split up to make sure the line width is exactly one for each cursor.

Closes https://github.com/eclipse-gef/gef-classic/issues/872


ptziegler:
It's only a visual problem that doesn't affect usability. I'm also running some tests right now, so I think it's better suited for the next release.

PR #877, opened 2025-12-01
Fix PrecisionRectangle#getLeft #getRight inconsistency

akoch-yatta:
Similar to #870, the x/y value of PrecisionRectangle#getLeft and PrecisionRectangle#getRight can return a different x or y value in comparision to the behavior of the other directional methods in PrecisionRectangle as #getBottomLeft. This PR overrides #getLeft and #getRight in PrecisionRectangle.


akoch-yatta:
We currently don't have any issues with that, but you could get into a similar situation we had with PrecisionRectangle#getBottom. Still, I would not consider this critical.

PR #880, opened 2025-12-02
Internalize deprecated fields in precision geometry classes

ptziegler:
The public fields of the PrecisionPoint, PrecisionRectangle and PrecisionDimension have been annotated with @noreference with 7865d7b918e456cb110405030e31e059ea045f38 and deprecated with 390e597e7523d4739154be895934b19997931f0f, over 15 years ago.

With this change, those fields are finally made private.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionDimension.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PrecisionDimension.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2024 IBM Corporation and others.
+ * Copyright (c) 2003, 2025 IBM Corporation and others.
  *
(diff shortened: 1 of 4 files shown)


ptziegler:
@HeikoKlare What about the following compromise: We keep the public fields for the time being but already adapt the getters to the new logic. This way you can still fall back to the current implementation if there are any issues and you wouldn't have to be concerned about any further changes once everything works.

One thing I noticed that might cause issues. If you update the integer-field, the fractional part will be cleared. This can't be done without those double-precision fields and something that would be a behavioral change.

Out of interest, how is this handled in SWT? From my understanding, a `Point(1, 1)` will internally be represented as a `Point.ofFloat(1.5, 1.5)` if at 150% zoom. If I update the `x` value to `2`, would the internal representation then be a `Point.ofFloat(2.5, 1.5)` as opposed to a `Point.ofFloat(3, 1.5)`?

PR #881, opened 2025-12-03
Adapt Logic editor to use "org.eclipse.ui.menus" extension point

ptziegler:
This replaces the deprecated "org.eclipse.ui.popupMenus" popup-menu contribution for the Logic editor with a "org.eclipse.ui.menus" contribution.

The LogicPrintAction has been renamed to LogicPrintHandler and now extends the AbstractHandler class.


ptziegler:
Merging, as this only affects the example editor.

PR #882, opened 2025-12-04
Correct mapping of tracker size for FlyoutPaletteComposite movement

HeikoKlare:
When dragging a flyout palette, a tracker on the Display is instantiated to show a preview of the drag result. When different zooms are involved (such as when using Draw2d autoscaling for the monitor-specific scaling capability on Windows), the tracker preview has a wrong size on non-100% monitors. This is caused by multiple Display#map() methods being applied on the same rectangle, which leads to the zoom being applied multiple times.

This change adapts the preview size calculation to prevent that the map operation (and thus the zoom) is not applied multiple times to the width and height of the preview rectangle. The change is backward compatible as without monitor-specific scaling capabilities and Draw2d autoscaling there are no zoom transformations involved, which is why the multiple map operations applied did not lead to any issues.

Fixes https://github.com/eclipse-gef/gef-classic/issues/868

For the tracker preview to work properly, a recent SWT state is also required, as an SWT fix is also necessary to make the tracker work properly:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2799

This supercedes and thus closes #https://github.com/eclipse-gef/gef-classic/pull/869
It is less invasive and thus better preserves existing behavior and reduces the risk of regressions.

### Before
flyout_tracker_broken

### After
flyout_tracker_fixed

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
@@ -951,6 +951,7 @@ public class FlyoutPaletteComposite extends Composite {
 			if (paletteContainer.getVisible()) {
 				bounds = bounds.union(paletteContainer.getBounds());
 			}
+			final Point origSize = new Point(bounds.width, bounds.height);
 			final Rectangle origBounds = Display.getCurrent().map(flyout, null, bounds);
 			final Tracker tracker = new Tracker(Display.getDefault(), SWT.NULL);
 			tracker.setRectangles(new Rectangle[] { origBounds });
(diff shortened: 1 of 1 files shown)


ptziegler:
> @ptziegler to have this work properly, you need a fix in SWT that was merged after the 2025-12 release:

You even said so in your initial commit message. Apologies... 🤦In that case I have to wait for the next SWT build, because I currently get this error message:

image

As the latest SWT build is not in https://download.eclipse.org/eclipse/updates/4.39-I-builds/, thus I'm missing https://github.com/eclipse-platform/eclipse.platform.swt/commit/5071736aa4ec89d3696a84b52ebb1b701256ad7a.

Wouldn't it be better to have a solution that works on as many SWT versions, rather than the upcoming one?

HeikoKlare:
> Wouldn't it be better to have a solution that works on as many SWT versions, rather than the upcoming one?

Thank you for the pointer. Even though the dialog does not pop up unexpectedly with the latest SWT version (i.e., latest I-Build and upcoming release), it could be an issue if you combined the Eclipse platform (namely the `Workbench`) with an older SWT version. So we might either make the behavior depend on the available SWT version or bump the lower bound for the SWT dependency. Since we access methods that have recently been introduced anyway, a version bump is probably the right thing to do (and is actually erroneously missing now). I will take of it.

HeikoKlare:
> The question was more related towards the change itself 😅. Given that #869 also works with older SWT versions, while this one requires the upcoming release.

Oh, that's even how I first read the comment and then I somehow moved over to applying it to the dialog only 😆 

So yes, the other solution would provide better backward compatibility, but at the same time it is a more invasive and thus more prone to unexpected errors. Some additional thoughts why it should not really matter and the benefit would be really limited:
- The pre-existing code works fine in case you do not use monitor-specific scaling; so without the involved SWT and GEF change, everything should be fine without monitor-specific scaling anyway
- Monitor-specific scaling is only well-supported for GEF when using Eclipse 2025-12 and above. With 2026-03, the according fix will be available. And for GEF any of the fixes will only become available with 3.27.

This sums up to the other solution only being better/available when using the combination of Eclipse 2025-12 and GEF Classic in 3.27 and above. Every other combination will work equally good or bad.

PR #885, opened 2025-12-06
Update all features and plugins for 2026-03 release cycle

ptziegler:


ptziegler:
I don't know enough about SonarQube to understand what this error is about:

> No modules found in the project or none of them are eligible for A3S analysis.

But given that it popped up all of a sudden, I assume it's either caused by an update to the Jenkins plugin or the SonarQube server.

ptziegler:
> But given that it popped up all of a sudden, I assume it's either caused by an update to the Jenkins plugin or the SonarQube server.

I think this is a new validation error that is thrown when the scanner is run on a non-Java project. I've updated the pom.xml file to skip all modules that don't contain a MANIFEST.MF file.

PR #895, opened 2025-12-09
Revert "Prevail rounding precision in ScaledGraphics operations"

ptziegler:
This reverts commit 88b83cf3a9793efa9c8f71ebd091a84586618ddf. Due to 1a9909791fabff667abf1626dc2a22e615330534, a ScaledGraphics instance always has a non-ScaledGraphics instance as parent, rendering this commit obsolete.


ptziegler:
@akoch-yatta fyi

azoitl (approved):

ptziegler:
~~I've also just created https://github.com/eclipse-platform/eclipse.platform.swt/pull/2867 to avoid this problem in the future. SWT is not supposed to throw such an exception.~~ Wrong PR

PR #896, opened 2025-12-09
Catch ServiceConfigurationError when JSVG is not in classpath

ptziegler:
When Draw2D is used in a pure Java environment, it might happen that the JSVG library, which provides the SVG support, is not in the classpath.

The call to isSvgSupported() should return `false` in such a situation, rather than throwing a ServiceConfigurationError. This problem can be observed when e.g. running any of the widgets that access an SVG.


azoitl (approved):

ptziegler:
I've also just created https://github.com/eclipse-platform/eclipse.platform.swt/pull/2867 to avoid this problem in the future. SWT is not supposed to throw such an exception.

PR #898, opened 2025-12-10
Allow signing builds from the maintenance branch

ptziegler:
The Jenkinsfile currently only allows signing and therefore also deploying the artifacts from the main branch to our update site. In order to allow the same for our maintenance branch, this rule needs to be relaxed.

Instead of requiring the branch to have the name 'master', it may also have the form 'Rx_yy_maintenance', where 'x' and 'y' are integers.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,6 +17,7 @@ pipeline {
   environment {
     BUILD_TIMESTAMP = sh(returnStdout: true, script: 'date +%Y%m%d%H%M').trim()
     MAIN_BRANCH = 'master'
+    MAINTENANCE_BRANCH = '^R\\d_\\d{2}_maintenance'
   }
 
   parameters {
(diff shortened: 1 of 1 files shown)


merks:
Yes I made it quite restrictive mostly so PRs are never signed.  Using something that includes maintenance branches is good. 

I wonder if you know about using Replay on a job to modify the pipeline quickly for testing changes?

ptziegler:
And for comparision https://ci.eclipse.org/gef/job/build-classic/view/change-requests/job/PR-866/6, which is not signed, even though I set it to TRUE in the pipeline.

image

merks:
Yes it’s really strongly intended to prevent a PR’s changes to be signed.  So you cannot actually test signing via changes via a PR.  You NEVER want a contributor to create signed results from some arbitrary change that you have not reviewed and approved. 

If you want to sign a branch, you should start such a build of a real branch and use Replay to apply changes until the build can sign that branch.

ptziegler:
> If you want to sign a branch, you should start such a build of a real branch and use Replay to apply changes until the build can sign that branch.

That's exactly what I did. ;D The screenshot was taken from a build where this change was applied. And the flag remains FALSE, even though it was enabled on Jenkins. Which is exactly how it should behave.

PR #901, opened 2025-12-13
Deprecate ScaledGraphics

azoitl:
All scaling needs for Draw2d can be handled by SWT natively and there are several issues in the implementation of ScaledGraphics.

FYI @HeikoKlare, @akoch-yatta

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/PrinterGraphics.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/PrinterGraphics.java
@@ -28,6 +28,7 @@ import org.eclipse.swt.widgets.Display;
  *
  * @author danlee
  */
+@SuppressWarnings("removal")
 public class PrinterGraphics extends ScaledGraphics {
 
 	Map<Image, Image> imageCache = new HashMap<>();
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableFreeformLayeredPane.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableFreeformLayeredPane.java
@@ -24,17 +24,25 @@ public class ScalableFreeformLayeredPane extends FreeformLayeredPane implements 
 
 	private double scale = 1.0;
 
+	/**
+	 * @deprecated will be deleted after the 2028-03 release (see
+	 *             {@link ScaledGraphics}).
+	 */
+	@Deprecated(forRemoval = true, since = "2026-03")
 	private final boolean useScaledGraphics;
 
 	public ScalableFreeformLayeredPane() {
-		this(true);
+		this(false);
 	}
 
 	/**
 	 * Constructor which allows to configure if scaled graphics should be used.
 	 *
 	 * @since 3.13
+	 * @deprecated will be deleted after the 2028-03 release (see
+	 *             {@link ScaledGraphics}).
 	 */
+	@Deprecated(forRemoval = true, since = "2026-03")
 	public ScalableFreeformLayeredPane(boolean useScaledGraphics) {
 		this.useScaledGraphics = useScaledGraphics;
 	}
@@ -97,7 +105,10 @@ public class ScalableFreeformLayeredPane extends FreeformLayeredPane implements 
 
 	/**
 	 * @since 3.13
+	 * @deprecated will be deleted after the 2028-03 release (see
+	 *             {@link ScaledGraphics}).
 	 */
+	@Deprecated(forRemoval = true, since = "2026-03")
 	@Override
 	public boolean useScaledGraphics() {
 		return useScaledGraphics;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@
 
 - Added new PrecisionPointList class. This class was created during the previous
   release and has now been finalized.
+- **Deprecated ScaledGraphics**  
+  *To be removed after 2028-03 release*   
+  All scaling needs for Draw2d can be handled by SWT natively and there are several issues in the implementation of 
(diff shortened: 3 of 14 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/PrinterGraphics.java
```diff
@@ -27,11 +27,16 @@
  * A scalable graphics object used to print to a printer.
  *
  * @author danlee
+ * @deprecated will be deleted after the 2028-03 release (see
+ *             {@link ScaledGraphics}).
  */
+@Deprecated(forRemoval = true, since = "2026-03")
 public class PrinterGraphics extends ScaledGraphics {
```
ptziegler: The `PrinterGraphics` class creates all resources using a `Printer`, rather than a `Display`. Wouldn't this cause issues when using a plain `SWTGraphics` object?

org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableFreeformLayeredPane.java
```diff
@@ -24,17 +24,25 @@ public class ScalableFreeformLayeredPane extends FreeformLayeredPane implements
...
+	/**
+	 * @deprecated will be deleted after the 2028-03 release (see
+	 *             {@link ScaledGraphics}).
+	 */
+	@Deprecated(forRemoval = true, since = "2026-03")
 	private final boolean useScaledGraphics;
 
 	public ScalableFreeformLayeredPane() {
-		this(true);
+		this(false);
```
ptziegler: I remember there being issues with the Logic editor when scaling with a plain SWTGraphics object. I'd have to double-check if this is still the case, but I would prefer to not change the default behavior when this already causes issues for our examples.

org.eclipse.draw2d/src/org/eclipse/draw2d/PrinterGraphics.java
azoitl: Most probably. I was very unsure about this one. I mainly did it because of the deprecated methods around the PrinterGraphics usage.

org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableFreeformLayeredPane.java
azoitl: I remember too. That is also my hesitation in my comment above. However I thought we could change it now early so that we find such problems and can address them and if we can not address them we can still change it back in M3 or RC1. Similar to the native scaling flag we introduced in the last cycle.

org.eclipse.draw2d/src/org/eclipse/draw2d/PrinterGraphics.java
ptziegler: I just tried out the PrinterExample with a plain SWTGraphics object and unsurprisingly, it doesn't work. Left it how it looks like, right is how it's supposed to look like.

image

Phillipus:
> Good point. Will remove this change and later look into how @Phillipus is doing this in Archi. Because this was the driver of the original API change around `PrinterGraphics`.

Mine is a bit of hack, see https://github.com/archimatetool/archi/blob/a033f986cf68ff3f5659ab7d9c5c3a11631ff3b2/org.eclipse.draw2d/src/org/eclipse/draw2d/PrintOperation.java#L31

(I need to align my hacked fork of GEF/Draw2d with the latest version sometime next year)

HeikoKlare:
I currently try to adapt our product to `SWTGraphics` being used by default. This means, I have to adapt all places that now default to `SWTGraphics` instead of `ScaledGraphics` to explicitly use `ScaledGraphics` instead (until we have eventually migrated to `SWTGraphics`). I run into some complications as not all consumers of the classes with changed defaults can easily be reconfigured. This includes:
- `GraphicalViewerImpl`/`ScrollingGraphicalViewer` create a `ScalableRootEditPart` in `createDefaultRoot()`. Every instantiation of those classes would have to be overwritten with a subclass that replaces this method with one that creates a `ScalabelRootEditPart` with `ScaledGraphics`. Every existing subclass needs to overwrite that method additionally
- `ScalableLightweightSystem` is instantiated at several places that are not easy to replace, such as `RulerViewer` or `PopupHelper`, which in turn are instantiated at other places that would have to be overwritten as well.
- `SimpleAutoscaledRootEditPart` creates a `ScalableLayeredPane` in `createFigure()`. The class is internal and currently only consumed by the `PaletteViewer` for which using `SWTGraphics` should hopefully be fine. Still I wanted to mentioned as we have a consumer in out product (comparable to the `PaletteViewer`), which probably needs to configure the pane to use `ScaledGraphics`.

The first one produces some effort and for the second I am not even sure how one could use the `ScalableLightweightSystem()` and its consumer with `ScaledGraphics` if desired.

@ptziegler the Changelog file mentiones the classes for which the default configuration has been changed but does not propose a strategy how to switch back to `ScaledGraphics` for them, in particular including all the non-configurable consumers of them. Is there any strategy how to properly switch back or do consumers potentially have to overwrite all consumers classes (such as `GraphicalViewerImpl` and the like) to preserve the existing behavior?

ptziegler:
I remember bringing up the same concern. My understanding was that we try out switchting to `SWTGraphics` by default and  reconsider if there are complications. 

https://github.com/eclipse-gef/gef-classic/pull/901#discussion_r2617433035

> I remember too. That is also my hesitation in my comment above. However I thought we could change it now early so that we find such problems and can address them and if we can not address them we can still change it back in M3 or RC1. Similar to the native scaling flag we introduced in the last cycle.

 To be honest, I haven't put much thought into the migration step, primarily because I don't really have dog in this fight.

PR #905, opened 2025-12-16
Make the creation of the PopUpHelper configurable

ptziegler:
This introduces a new `IToolTipHelperFactory` that is used as an OSGi service in the `DomainEventDispatcher` to allow consumers to inject their own `ToolTipHelper`.

It is currently very difficult to customize the tool-tips shown by an edit-part viewer in any meaningful ways, as one has to:

a) Subclass the event dispatcher to use a custom tool-tip helper b) Subclass the LWS to use this custom event dispatcher c) Subclass the viewer to use this custom LWS

A similar customization was already done in the past via e9839e7588032a97ede7c1fbdb8503d7a0b908aa. The advantage of using a service is that it allows implementations to customize the tool-tip of a specific viewer, rather than all viewers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ Reason for this change is the `TreeViewer`, where the visual part of an `EditPar
 realized as a `Widget` and not an `IFigure`. For the `findObjectAtExcluding` method
 specifically, the `TreeViewer` is using `EditPart`'s as exclusion set.
 
+* Added new `IToolTipHelperFactory` interface that is used as factory in the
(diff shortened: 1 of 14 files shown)


ptziegler:
For my use case:

The description of the palette entries in our editor is done as raw HTML. Therefore our tool-tip provider doesn't use a `FigureCanvas`, but rather an SWT `Browser`.

image

Getting this to work in GEF is a real pain because I can't modify the tool-tip created by the palette without subclassing all of the internal edit parts and I can't even inject the tool-tip helper without subclassing the `DomainEventDispatcher`.

Issue #903, opened 2025-12-16
Generic type in EditPartViewer.getVisualPartMap is wrong


ptziegler:
As the documentation says: The visual part may be an `IFigure` **OR** a `TreeItem` (or perhaps something completely different).

https://github.com/eclipse-gef/gef-classic/blob/9a17c909d459b7b2c02a9d4d4add7c37326789a6/org.eclipse.gef/src/org/eclipse/gef/EditPartViewer.java#L412-L422

PR #906, opened 2025-12-16
Further refinement of the return type of getVisualPartMap() #903

ptziegler:
This is a continuation of c5b0306ff812205be4d9dd6b554c476b0fd61095. Due to this change, consumers may now experience a compilation error when calling `getViewer().getVisualPartMap().put(...)`.

When calling `getVisualPartMap()`, a wildcard generic map is returned, which does not allow the addition of values. To get around this, one has to cast the map to a specific type, which is not something that should be done on consumer side. But the keys also can't be plain `Object`s, because this prevents overriding this method in sub-classes.

To get around this limitation, the following approach is taken: The signature in the `EditPartViewer` interface is changed so that this method returns a generic map. The generic type is not specified, so calling `getVisualPartMap()` returns a `Map<Object,EditPart`, which allows the addition of entries.

But when inside the AbstractEditPartViewer, this method can be overridden to return a `Map<?,EditPart`, which can then be overridden again for the `GraphicalEditPartViewer` and the `TreeEditPartViewer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ## GEF
 
 * Due to an oversight, the generic types of the following type had to be adapted:
-  * Map<~~IFigure~~?, EditPart> EditPartViewer.getVisualPartMap()`
+  * <T> Map<~~IFigure~~T, EditPart> EditPartViewer.getVisualPartMap()`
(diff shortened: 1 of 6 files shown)


azoitl (approved):
This is a very smart solution. One of the reasons I love to participate in GEF Classic is that I regularly learn more about the possibilities of Java.

PR #907, opened 2025-12-18
Add missing `@SuppressWarnings` where deprecated types are used

ptziegler:
With the 4.38 Eclipse release, JDT revised the way deprecations are handled. As a result, referencing a deprecated method/class now produces a deprecation warning, even when used inside another deprecated method/class.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4572


ptziegler:
Luckily, the deprecation warnings are only related to the JDT changes. The build is green again now that all missing annotations have been added.

PR #914, opened 2025-12-18
Generify PaletteViewerKeyHandler.getNavigationSiblings() #155

ptziegler:
The generic type is already defined by the PaletteViewerKeyHandler. This requires several type-casts, as the call to `EditPart.getParent()` is generally not guaranteed to return a `GraphicalEditPart`.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/155

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/parts/PaletteViewerKeyHandler.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/parts/PaletteViewerKeyHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


ptziegler:
> Should a getParent of a GraphicalEditPart always be a GraphicalEditPart?

That's should be the case for a `PaletteEditPart`. At least that's how I understand the documentation of the `PaletteEditPart`.

https://github.com/eclipse-gef/gef-classic/blob/16d9b027a9aa155f45ecbb047fa2c614c3d6077a/org.eclipse.gef/src/org/eclipse/gef/ui/palette/editparts/PaletteEditPart.java#L52-L53

The root is generally a `SliderPaletteEditPart` and the children of a `PaletteEditPart` are also `PaletteEditPart`s. My only concern is that we never really enforce this contract. But then I can simply override `getParent()` in `PaletteEditPart` and `getFocusEditPart()` in `PaletteViewerKeyHandler`.

azoitl:
> > The root is generally a SliderPaletteEditPart and the children of a PaletteEditPart are also PaletteEditParts
> 
> But the parent of the `SliderPaletteEditPart` is not. So that already doesn't work :/

But they would be GraphicalEditParts or?

ptziegler:
> But they would be GraphicalEditParts or?

Outside of the palette root, yes. But that's something I don't want to rely on. The only problematic part is the `buildNavigationList` method, where I decided to do the explicit type cast. Which is safe, given that it is only done for palette models, which are expected to be realized by a `PaletteEditPart`.

azoitl (approved):

PR #923, opened 2025-12-21
Implement zoomTo(Rectangle)

azoitl:
The missing rectangle zoom is implemented such that the rectangle is maximized in the viewport and that the center of the rectangle is in the center of the viewport.

Implements: https://bugs.eclipse.org/bugs/show_bug.cgi?id=172463

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
     - ScalableRootEditPart
     - ScalableLightweightSystem
 - Added `Rectangle.getBounds()`
+- Implemented AbstractZoomManager.zoomTo(Rectangle): This allows to implement features 
+  as zoom to selection or any given bounding box (e.g., Marquee selected box). 
+  This feature was originally request in [Bug 172463](https://bugs.eclipse.org/bugs/show_bug.cgi?id=172463)   
 
 ## GEF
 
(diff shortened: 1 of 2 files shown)


ptziegler:
You're right. If I move one of the nodes to the bottom right corner, the other nodes are centered when double-clicking on them. But the scrollbar in the Zest viewer behaves really and definitely not in a way I would expect. But that is completely unrelated to this change.

azoitl:
> You're right. If I move one of the nodes to the bottom right corner, the other nodes are centered when double-clicking on them. But the scrollbar in the Zest viewer behaves really and definitely not in a way I would expect. But that is completely unrelated to this change.

In 4diac IDE we faced a similar problem. Our solution is to have around the pure drawing canvas a passe-partout. Similar how draw IO is doing it:
image

I have on my roadmap to upstream this once. The only thing with my passe-partout implementation is that I think it works not with all figure canvases. I have an incremental growing one for this it fits perfectly.

If you want to see how what I did these are the two classes:
 - The passe-partout: https://github.com/eclipse-4diac/4diac-ide/blob/release/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/BackgroundFreeformFigure.java
- The incremental growing canvas: https://github.com/eclipse-4diac/4diac-ide/blob/release/plugins/org.eclipse.fordiac.ide.gef/src/org/eclipse/fordiac/ide/gef/figures/ModuloFreeformFigure.java

One feedback that we got from our users on the passe-partout based canvas is that they clearly see where the canvas ends.

PR #926, opened 2025-12-22
Implement getAbsoluteScale for SWT Graphics

azoitl:
As SWT Graphics allows different scale factors in x and y, rotations, and shear which is all stored in the transformation matrix it is not as easy to get a single value for the current absolute scale. This implementation derives an approximation from the determinate of the matrix. This represents an average scale factor. For non rotated uniform x/y scaling it is the exact scale factor.

Even if it is an approximation it is suitable for level of detail checking code.

Implements: https://bugs.eclipse.org/bugs/show_bug.cgi?id=267462

An example of how to use this can be seen in https://github.com/azoitl/gef-classic/commit/7522bbc202b7abd26fb313e5218ea7ff75f51dfb

This is especially interesting for outline figures of large diagrams. I could see a significant improvement when testing that with 4diac IDE.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java
@@ -680,6 +680,26 @@ public class SWTGraphics extends Graphics {
 		return ((currentState.graphicHints & AA_MASK) >> AA_SHIFT) - AA_WHOLE_NUMBER;
 	}
 
+	private final float[] transformElements = new float[6];
+
+	/**
+	 * Returns a stable, direction-independent estimate of the current graphics
+	 * scaling, computed as the square root of the absolute determinant of the
+	 * transformation matrix.
+	 *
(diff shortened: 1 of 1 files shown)


ptziegler (approved):
Using the determinat of the transformation matrix as scaling factor absolutely makes sense.

azoitl:
> Using the determinat of the transformation matrix as scaling factor absolutely makes sense.

@ptziegler I would nonetheless not use it for image scaling as it gives you an average of both scales. Here the max of x and y scaling may be better.

ptziegler:
> Yes, absolutely. I've thought about adding a getter for a Draw2D [Transform](https://github.com/eclipse-gef/gef-classic/blob/master/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Transform.java) or something similar, which would give you the horizontal and vertical scaling.

Related: [Bug 130339](https://bugs.eclipse.org/bugs/show_bug.cgi?id=130339) - Graphics: Expose transform api in SWT GC to allow for advanced transformations

PR #931, opened 2025-12-23
[Zest] Fix paint logic in GraphLabel

ptziegler:
Because SWT doesn't provide any API to draw a rectangle with round color and a color gradient, the `GraphLabel` has split up its paint operation into three parts.

1) The upper and lower part of the figure are painted, in a solid color and with round edges.

2) The center part is painted as a rectangle with gradient color.

3) The border is painted, again in a solid color and with round edges.

There are several problems with this approach:

- The round corners don't perfectly align with the rectangle in the center, leading to visual artifacts that become especially noticeable when no border is painted.

- The way the corners are painted leads breaks, if high arc widths are chosen.

- The arc of the corner is sharper than the arc of the border, which causes the inner part to stick out at higher zoom levels.

As mentioned, the main reason for this approach is a lack of API on the SWT side. However, this API is not needed. By using a `Path`, a clipping can be created in such a way that it perfectly encloses a round rectangle. Within, one can then simply paint a rectangle with gradient color. A round border is then painted around this rectangle.

Because the border and the rectangle have the same arc length, no overlap occurs.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/internal/GraphLabel.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/internal/GraphLabel.java
@@ -15,6 +15,7 @@ package org.eclipse.zest.core.widgets.internal;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.Path;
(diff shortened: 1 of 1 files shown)


ptziegler:
This is how the label currently looks like when zoomed in:

image

Again without border and an arc-width of 16:

image

And now with this change and also an arc-width of 16:

image

azoitl (approved):

PR #932, opened 2025-12-23
Add new `drawImage()` methods to `Graphics` for drawing scaled images

ptziegler:
This adds wrapper methods for the `drawImage(Image,int,int,int,int)` method that was introduced in SWT 1.312. Unlike the other `drawImage` methods, it is able to properly scale SVG images if the `Image` is backed by an `ImageDataProvider`.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/922


ptziegler:
A side-by-side comparison of all `drawImage()` methods. Only the `drawImage(Image,int,int,int,int)` method is able to properly scale SVG images.

image

The test image is also my attempt to recreate the GEF logo, since I only found a PNG. I can provide the original 120x120px file if there is interest.

azoitl:
The GEF 5 repo has one with 350x350 and I thought there was something on the old web-page. However I couldn't find the web-page in gitlab anymore, not sure where it went.

Phillipus:
Why not get the bounds from the Image with `Image.getBounds()`?

```java
public void drawImage(Image image, int destX, int destY, int destWidth, int destHeight) {
	ImageData src = image.getImageData(100);
	drawImage(image, src.x, src.y, src.width, src.height, destX, destY, destWidth, destHeight);
}
```

HeikoKlare (commented):
Thank you for adding that method! We intended to provide a PR after adding that method to SWT (as we specifically require it for GEF and currently have that method patched into our Draw2d bundle) but were not sure regarding backwards compatibilty to older SWT versions yet (e.g., performance impact of using reflection), but you now were faster anyway 🙂

Even though `ScaledGraphics` is deprecated, could we add a proper implementation to that class as well? We already have one patched into out GEF/Draw2d bundle as we require that method and it would be nice for us (and probably other remaining `ScaledGraphics` users) to benefit from the new functionality without patching. 
This is how we implemented the method in `ScaledGraphics`:
```java
	public void drawImage(Image srcImage, int x, int y, int width, int height) {
		graphics.drawImage(srcImage, (int) (Math.floor((x * zoom + fractionalX))),
				(int) (Math.floor((y * zoom + fractionalY))), (int) (Math.floor((width * zoom + fractionalX))),
				(int) (Math.floor((height * zoom + fractionalY))));
	}
```
And this is the derived one in `PrinterGraphics`:
```java
	public void drawImage(Image srcImage, int x, int y, int width, int height) {
		super.drawImage(printerImage(srcImage), x, y, width, height);
	}
```

Phillipus:
> We intended to provide a PR after adding that method to SWT (as we specifically require it for GEF and currently have that method patched into our Draw2d bundle)

I added the `drawImage(Image,int,int,int,int)` methods to our fork of Draw2d.

They work fine and scaled up SVG images are sharp...but only on Windows, not Mac or Linux. See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2917

HeikoKlare:
> They work fine and scaled up SVG images are sharp...but only on Windows, not Mac or Linux. See [eclipse-platform/eclipse.platform.swt#2917](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2917)

As mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2917, I can confirm the limitations with MacOS and Linux (not supporting sharp scaling of images initialized with plain file name or InputStream). We intend to fix the issue for MacOS/Linux in the 4.39 release.

So you might consider guarding the usage of the new `drawImage` method in Draw2d's `Graphics` implementations to the upcoming SWT version, in order to avoid that someone adopts the method and tests it with SWT from Eclipse release 4.39 or newer and then in production combines it with a 4.38 release of Eclipse which already contains the method but don't provide the same capabilities on MacOS and Linux yet. You might also argue that it's still fine to use the method with a 4.38 version of Eclipse as the results will usually not be worse than using the existing draw method as that one is also not capable of sharply rendering SVGs, but I still wanted to mention that it could be something to consider when adopting that new method.

Phillipus:
> We already have one patched into out GEF/Draw2d bundle as we require that method and it would be nice for us (and probably other remaining `ScaledGraphics` users) to benefit from the new functionality without patching.

Thanks for that. I added it to our fork (with credits to @HeikoKlare for this one and @ptziegler for the other implementations)

ptziegler:
> (e.g., performance impact of using reflection)

Regarding the potential performance impact: I can just call the SWT method using a `MethodHandle` instead of reflection. They're supposed to be the high-performance replacement anyway...

HeikoKlare:
Great, thank you! I wasn't even aware of the `MethodHandle`; good to know about that technique.

PR #935, opened 2025-12-23
Directly provide positions in label drawings

azoitl:
PaintFigure is heavily using Graphics#translate. This is an overkill and cost more performance then what it gains in readability and maintainability.

Especially when having many labels (a few 1000) and with SWT Graphics this can have a significant memory (mainly garbage) and performance gain.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/Label.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/Label.java
@@ -493,19 +493,19 @@ public class Label extends Figure implements PositionConstants {
 			super.paintFigure(graphics);
 		}
 		Rectangle bounds = getBounds();
-		graphics.translate(bounds.x, bounds.y);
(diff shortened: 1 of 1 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/Label.java
```diff
@@ -493,19 +493,18 @@ protected void paintFigure(Graphics graphics) {
 			super.paintFigure(graphics);
 		}
 		Rectangle bounds = getBounds();
-		graphics.translate(bounds.x, bounds.y);
 		if (icon != null) {
-			graphics.drawImage(icon, getIconLocation());
+			graphics.drawImage(icon, bounds.x, bounds.y);
```
ptziegler: Wouldn't this mess with subclasses that override getIconLocation()/getTextLocation()?

ptziegler (approved):

PR #940, opened 2026-01-01
Fix improper bounds calculation for MoveHandleLocator when zoomed in

ptziegler:
The `MoveHandleLocator`shrinks the size of the target figure by one, handles the insets and then expands it by one again. The problem is that the first operation is done before translating the bounds to absolute coordinates. Instead of shrinking the size by one, it is therefore shrunk by one times the zoom factor.

As a result, the bounds of the target figure don't fully enclose the bounds of the figure owned by the locator.

To reproduce: Open the Logic editor and select e.g. a Gate. When selecting the element, the selection box doesn't fully enclose the right and bottom of the figure.

### DIFF
--- a/org.eclipse.gef.tests/src/org/eclipse/gef/test/GEFTestSuite.java
+++ b/org.eclipse.gef.tests/src/org/eclipse/gef/test/GEFTestSuite.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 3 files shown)


ptziegler:
For illustration, I've updated the Gate figure to draw a simple rectangle. As you can see, the selection box is too small:

image

The initial change was done with 7a7906105540be07e60db5f5d9729684c3bfef2b and e7a9fdab57776bc44a8212744c473f6537eac7c1. But due to the poor commit message, I can't say anything about _why_ it was done like this.

azoitl:
> I don't see why it was added but I also can't predict the side-effects that would be introduced by removing them. My current guess is that it likely avoids some rounding errors when scaling the figures.
>
> All I can say is that it not a +1/-1, but rather a +1/(-1*zoom), which is causing these artifacts.

Me neither. I only fear that now that you changed it to PrecisionRectangle that we have now more or different rounding issues and that the +1/(-1*zoom) adds to that. However I have no clue how to find that out.

PR #941, opened 2026-01-01
Cleanup Gate figures in Logic Editor example

ptziegler:
This updates the AND, OR and XOR figures so that:

1) The feedback figures are transparent again, similar to the other figures of the logic editor. Because the editor is now using an `SWTGraphics` instance for painting (with advanced mode), the transparency needs to be explicitly set. This requires that the individual shapes of the figures don't overlap one another.

2) The figures are now properly centered so that they look good when selected.

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/AndGateFigure.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/AndGateFigure.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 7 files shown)


org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/AndGateFigure.java
```diff
@@ -57,22 +57,28 @@ protected void paintFigure(Graphics g) {
...
+		g.drawLine(r.right(), r.y + CORNER_ARC / 2, r.right(), r.bottom());
 
 		// draw and outline the arc
+		r.y += 6;
 		r.height = 18;
-		r.y += 7;
+		g.setAlpha(getAlpha());
 		g.fillArc(r, 180, 180);
-		r.width--;
-		r.height--;
```
ptziegler: That's a pattern I see quite often, but which doesn't make sense to me. Why would the filled arc be larger than its outline? Even then, it should be shrunk by a multiple of two, to account for both sides instead of just one...

azoitl (commented):
This is much cleaner and looks very nice. 
What do you think about making also the outline transparent. This would make the code for drawing the shapes much cleaner. And if we increase the alpha a bit (e.g, 150 -170) we get a nice feedback figure.

azoitl (approved):

azoitl:
> @azoitl Should I still try out different alpha values? The value of 100 was effectively chosen at random.

@ptziegler when I tried your PR with the opaque border the alpha value of 100 looked really good. Only when the border was not opaque I needed some higher value. Therefore I would leave it as it for now.

azoitl:
This is strange. Is this in SWT? A further reason we need to get the the Skia Canvas up and running. I really hope that then we are less OS dependent. I haven't yet found time to test the main Draw2d examples with it.

PR #947, opened 2026-01-07
Execute GEF tests in the UI thread

ptziegler:
The GEF tests are currently run in a separate thread to simplify the SWTBot integration. Main motivation being that the tests can be executed without blocking the UI updates that would occur when e.g. clicking on a palette item.

However, this leads to very ugly tests because all tests that directly work with a widget now needs to explicitly synchronized.

The SWTBot tests have been explicitly updated to wait for UI events where necessary, which allows us to execute them in the UI thread and avoid all of this overhead.


ptziegler:
It's interesting that this exception wasn't logged before. It is caused by this command here, which references an SVG image that doesn't yet exist in the 2024-09.

https://github.com/eclipse-gef/gef-classic/blob/4463d9d0ce13c19d10fbe769d29f28b7345a8cba/org.eclipse.gef.examples.logic/plugin.xml#L43-L66

azoitl:
This is pretty cool. I always thought SWTBot calls must not be on the ui thread. Did I miss read the docs?

azoitl:
Good to know. But we are on junit 5 and do not inherit from ' SWTBotTestCase' anymore.

PR #952, opened 2026-01-10
Add SVG-based Graphics class backed by Batik

ptziegler:
This introduces two new classes; The `AWTGraphics` and `SVGGraphics`. Former is an implementation of the  `Graphics` class using Swing/AWT as rendering toolkit and which is extended by the latter, to transform the paint operations into valid SVG using the Batik library.

To avoid adding a new dependency to `org.eclipse.draw2d`, the `SVGGraphics` class is contained by the new `org.eclipse.draw2d.svg` pluginm, while `AWTGraphics` is embedded directly in the base plugin.

The `GraphicsTest` and `GraphicsClipping` tests have been adapted to now also execute the test methods using an `AWTGraphics`, rather than just an `SWTGraphics` and/or `ScaledGraphics` instance.


ptziegler:
> An SVG Figure that would draw an SVG using gc primitives or even turn it into figures/editparts?

So you want to parse the SVG and then create  "SVG" figures for all elements in this document? Conceptually, this shouldn't be a problem. But to me that sounds like an "SVG Editor" and not something that really fits into this PR. Perhaps this is something that could be turned into a new example editor?

The "Shapes" editor does something similar, right? You'd "only" need to swap out the persistence layer and add some SVG code generation for the shapes and then you're already on a good track. At least this part should fit into the scope of a GSoC project.

WindowBuilder does something similar, except that we have a Swing/SWT container as source. So it's definitely possible.

ptziegler:
@azoitl What do you think of the changes to the `SWTGraphics` class? Rather than duplicating the entire implementation, I moved everything to a common base class, so that the class only contains the SWT-specific components. Given the amount of changes, I'd definitely postpone this for the M1.

I will then also split this up into two PRs. One for the `SWTGraphics` restructuring and one for the `SVGGraphics`.

azoitl:
> @azoitl What do you think of the changes to the `SWTGraphics` class? Rather than duplicating the entire implementation, I moved everything to a common base class, so that the class only contains the SWT-specific components. Given the amount of changes, I'd definitely postpone this for the M1.
 
@ptziegler I find this split a very good idea. Maybe this gives us in the future an option to simplify and optimize this code. 

I personally wouldn't postpone it to then next M1. The earlier we have it merged the earlier we get feedback.

ptziegler:
I'm really not sure if/how this can be made smaller. Another open topic that we should keep in mind is the interaction with Draw2D and SWT. When painting in the Event Dispatcher thread, one may get errors such as:

`
Exception in thread "AWT-EventQueue-0" org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4934)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:598)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:280)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:271)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:221)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:156)
	at org.eclipse.draw2d.FigureUtilities.getGC(FigureUtilities.java:100)
	at org.eclipse.draw2d.FigureUtilities.getTextDimension(FigureUtilities.java:122)
	at org.eclipse.draw2d.FigureUtilities.getTextExtents(FigureUtilities.java:163)
	at org.eclipse.draw2d.TextUtilities.getTextExtents(TextUtilities.java:60)
	at org.eclipse.draw2d.Label.calculateTextSize(Label.java:231)
	at org.eclipse.draw2d.Label.getTextSize(Label.java:457)
	at org.eclipse.draw2d.Label.getPreferredSize(Label.java:320)
	at org.eclipse.draw2d.Figure.getPreferredSize(Figure.java:892)
	at org.eclipse.draw2d.Label.getSubStringText(Label.java:350)
	at org.eclipse.draw2d.Label.paintFigure(Label.java:508)
	at org.eclipse.draw2d.Figure.paint(Figure.java:1275)
`

ptziegler:
Compilation errors are expected. See https://github.com/eclipse-gef/gef-classic/pull/1181.


> [ERROR] The Java feature 'Pattern Matching in Switch' is only available with source level 21 and above
> [ERROR] D:\a\gef-classic\gef-classic\org.eclipse.draw2d.tests\src\org\eclipse\draw2d\test\utils\Graphics2DToGraphicsAdapter.java:[142] 
> [ERROR] 	case Path2D path -> fillPath(path);
> [ERROR] 	     ^^^^^^^^^^^

pcdavid:
I have not read your exchanges in details, but it looks like there might be an overlap with what GMF does [here](https://github.com/eclipse-gmf-runtime/gmf-runtime/tree/master/bundles/org.eclipse.gmf.runtime.draw2d.ui.render.awt/src/org/eclipse/gmf/runtime/draw2d/ui/render/awt/internal/svg).

Sirius Desktop has its own (quite intricate and not very satisfying) support for both [rendering SVG images on a diagram](https://github.com/eclipse-sirius/sirius-desktop/blob/master/plugins/org.eclipse.sirius.diagram.ui/src-diag/org/eclipse/sirius/diagram/ui/tools/api/figure/SVGFigure.java) and [exporting a diagram as an SVG file](https://github.com/eclipse-sirius/sirius-desktop/tree/master/plugins/org.eclipse.sirius.diagram.ui/src-diag/org/eclipse/sirius/diagram/ui/tools/internal/render).

Obviously, if all 3 projects could converge on a single implementation directly in GEF that'd be great.

Also note that for the actual SVG rendering, recent versions of Eclipse have https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/bundles/org.eclipse.swt.svg based on com.github.weisj.jsvg (might lighter/simpler than Batik), though I have not yet found time to look at it closely.

ptziegler:
@pcdavid  If I understand your examples correctly, those are not only about exporting, but also about painting an SVG image on the SWT canvas. The use case here is really just about providing a (hopefully) simple way of exporting a diagram to an SVG file.

In addition to this PR, there was also [this](https://github.com/eclipse-gef/gef-classic/discussions/939) discussion, where a similar [implementation](https://gitlab.eclipse.org/eclipse/graphiti/graphiti/-/blob/master/plugins/org.eclipse.graphiti.export.batik/src/org/eclipse/graphiti/export/batik/SVGExporter.java) from Graphiti was mentioned, which internally relies your GMF Graphics2Graphics2D adapter (or rather, I think it was copy-pasted).

The main motivation behind this PR was that Graphiti itself is very much on life-support and you don't really want to pull in such a juggernaught, just for creating a simple SVG file.

From working on it, I can tell that it's a real nightmare to get any new implementation of the `Graphics` class to behave similarly to the `SWTGraphics` class. And I would expect any other implementations to face a similar challenge.

Ideally, it probably would've been for the best if one of those SVG implementations would've been contributed back to the main project, in order to avoid all of this overlap. But at that point, the GEF project likely already shifted to JavaFX.

pcdavid:
This one supports exporting a Sirius diagram as an SVG file: https://github.com/eclipse-sirius/sirius-desktop/tree/master/plugins/org.eclipse.sirius.diagram.ui/src-diag/org/eclipse/sirius/diagram/ui/tools/internal/render, with a fork of GMF Runtime's GraphicsToGraphics2DAdaptor.

But feel free to ignore these links if they don't fit your need, I only mentioned them in case you could find useful code there.

merks:
FYI, for the p2 aggregator we are using some internal from GMF to export the dependency diagram to SVG:

https://github.com/eclipse-cbi/p2repo-aggregator/blob/537f7c21fab9ca93c2f7b332f9d5874e64b8f9ad/plugins/org.eclipse.cbi.p2repo.aggregator.analyzer.editor/src/org/eclipse/cbi/p2repo/aggregator/analyzer/presentation/AnalyzerEditor.java#L4152

It's be very happy with APIs that do this.

ptziegler:
> I only mentioned them in case you could find useful code there.

I appreciate te suggestion and I remember using them to cross-check my implementation of e.g. the conversion between SWT Paths and AWT Paths, together with what the coding agent suggested. Having an existing code-base definitely helped a lot.

> It's be very happy with APIs that do this.

Just at a quick glance, I think the same behavior can be achieved with this `Graphics` class and it should look something like:

```java
Dimension surfaceSize = normalizer.getBounds().getSize();

SVGGraphics g = new SVGGraphics();
g.setSVGCanvasSize(surfaceSize);
g.setClip(new Rectangle(0, 0, surfaceSize.width, surfaceSize.height));

root.paint(g);

normalizer.reset();

try (StringWriter stringWriter = new StringWriter()) {
	g.stream(stringWriter);
	stringWriter.flush();
	return stringWriter.toString();
} catch (IOException e) {
	throw new RuntimeException(e);
}
```

Note that the clip needs to be explicitly set (unless a custom SVGGraphics object is used). In SWT, the clip is always initialized with the bounds of e.g. the `Image` or `Composite` from which the GC instance was created. That concept just doesn't exist for SVGs.

However, this `SVGGraphics` object doesn't expose the internal `SVGraphics2D` instance, because that would allow you to completely bypass the `Graphics` API. You can provide your own instance, but the Batik API for that is ugly and I haven't come up with a cleaner way here.

PR #959, opened 2026-01-16
Add ScaledGraphics tests for drawing texts/strings

ptziegler:
Make sure that the font height returned by `graphics.getFontMetrics()` and `FigureUtilities.getFontMetrics()` get reasonable values.

See https://github.com/eclipse-gef/gef-classic/issues/945


ptziegler:
Curious how this didn't fail with 2024-09 :thinking:

Issue #945, opened 2026-01-06
SWTGraphic scaling inconsistent across platforms


ptziegler:
This was brought up [here](https://github.com/eclipse-gef/gef-classic/pull/901#discussion_r2651243866) and [here](https://github.com/eclipse-gef/gef-classic/pull/901#discussion_r2656544265). When a plain `SWTGraphic` is used, scaling is done internally using an SWT `Transform`. This class seems to behave differently on different platforms.

For example the following snippet:
```java
package test;

import org.eclipse.draw2d.SWTGraphics;
import org.eclipse.draw2d.ScaledGraphics;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.graphics.GC;
import org.eclipse.swt.graphics.Pattern;
import org.eclipse.swt.graphics.Transform;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class Test {
	public static void main(String[] args) {
		// System.setProperty("swt.autoScale.updateOnRuntime", "true");
		Shell shell = new Shell();
		shell.setSize(400, 400);

		Display d = shell.getDisplay();

		Color white = new Color(255, 255, 255);
		Color gray = new Color(128, 128, 128);

		shell.addPaintListener(event -> {
			GC gc = event.gc;

			for (int y = 0 ; y < shell.getSize().y; y += 50) {
				gc.setBackground(y % 20 == 0 ? white : gray);
				gc.fillRectangle(0, y, shell.getSize().x, 50);
			}
			
			SWTGraphics g1 = new SWTGraphics(gc);
			g1.translate(50, 100);
			g1.scale(2.5f);
			g1.setBackgroundColor(white);
			g1.drawText("Hello World", 0, 0);
			g1.dispose();

			gc.setTransform(null);
			

			ScaledGraphics g2 = new ScaledGraphics(new SWTGraphics(gc));
			g2.translate(50, 100);
			g2.scale(2.5f);
			g2.setBackgroundColor(white);
			g2.drawText("Hello World", 0, 0);
			g2.dispose();
		});

		shell.open();
		
		while(!shell.isDisposed()) {
			while(!d.readAndDispatch()) {
				d.sleep();
			}
		}
	}
}
```

<img width="383" height="392" alt="Image" src="https://github.com/user-attachments/assets/489bf2ac-99d6-494b-9b45-9d5448f28bea" />

ptziegler:
[...]

PR #961, opened 2026-01-16
Fix zoom error for fonts in ScaledGraphics

ptziegler:
The native scaling in the internal `Shell` of the `FigureUtilities` class is always disabled when running on Windows. This means that calling `FigureUtilities.getFontMetrics()` always returns the metrics of an *unscaled* font.

This method is called in the `ScaledGraphics` to determine, whether the internal font has been changed and calculates the zoomed location by comparing the local with the target font. The target font is using the font metrics of the internal `Graphics` object, which might be of a *scaled* font.

It is therefore possible that the same font has different heights (depending on the graphics context), which leads to an incorrect placement.

To avoid this a new `FigureUtilities.getFontMetrics(Font, Graphics)` method is added, which returns the font metrics with respect to the internal graphics context of the `ScaledGraphics` instance.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/945

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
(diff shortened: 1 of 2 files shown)


ptziegler:
I've also tested this with #959 at 125% zoom and with the snippet in #945. The tests are successful and the snippet looks the same on Windows as it does on Linux.

PR #963, opened 2026-01-17
[Win32] Fix incorrect font scaling when Transform is used

### LINKED ISSUE #945: SWTGraphic scaling inconsistent across platforms

ptziegler:
On Windows, the font sizes returned by `getFontMetrics()` are incorrectly scaled by the Transform and are drawn larger than they are supposed to. This is an issue in SWT. As a workaround, the Transform can be cleared, the font metrics flushed and then the Transform re-applied again.

To reproduce, open the Logic editor on Windows, add an LED figure and change the editor zoom to something != 100%. The LED text should then no longer be aligned properly.

Closes https://github.com/eclipse-gef/gef-classic/issues/945

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AdvancedGraphicsTests.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AdvancedGraphicsTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2025 IBM Corporation and others.
+ * Copyright (c) 2004, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -19,6 +19,7 @@ import java.util.Deque;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
+import org.eclipse.swt.graphics.FontMetrics;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
@@ -32,9 +33,12 @@ import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 
 import org.eclipse.draw2d.ColorConstants;
+import org.eclipse.draw2d.FigureUtilities;
 import org.eclipse.draw2d.SWTGraphics;
 
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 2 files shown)


ptziegler:
Before:

Before

After:

After

The problem itself probably existed since forever. But only with certain fonts and when using an SWTGraphics instead of a ScaledGraphics, which is probably why noone complained up until now.

azoitl (commented):
@ptziegler shold we wait for some feedback from the SWT people first, or merge it as workaround for now? If we do the latter I guess it would be good to refere the PRs and issues with each other to keep track on what is going on.

azoitl:
@ptziegler I'm a bit concerned about the performance impact of your workaround.

Given the feedback of @HeikoKlare should I improve the Logic editor example to better use fonts? E.g., in 4diac IDE we use the theming api and make the fonts even adjustable for users.

azoitl:
> > I'm a bit concerned about the performance impact of your workaround.
> 
> Wouldn't this only be the case if you update the font constantly? For the majority of the time, this should be a no-op operation.

I have shapes with three different fonts and in one diagram a few hundreds of them. Therefore I'm concerned.

> 
> > Given the feedback of @HeikoKlare should I improve the Logic editor example to better use fonts?
> 
> So we should just ignore the problem, instead of fixing it?

We should definitely also provide a fix. 

However I see the Logic Editor as our best practices collection. Therefore I think we should also show there how you would do it in a platform neutral way correctly.

ptziegler:
Closing, as handling of invalid fonts should be done on the SWT side. For the Logic editor, we now use a valid font, so the original problem has already been solved.

Issue #965, opened 2026-01-17
Palette width is different on Windows 100% and 200% scaling


Phillipus:
- Eclipse 4.38
- Targeting latest commit of GEF/Draw2d
- Windows 11
- Display resolution 3840x2160
- Display scale 100% and 200%

1. Set display scaling to 100%
2. Set palette width by dragging the sash
3. Close app
4. Set display scaling to 200%
5. Open app
6. Palette width is different

At 100% scaling:

<img width="1407" height="609" alt="Image" src="https://github.com/user-attachments/assets/f62e5cea-4301-43d2-bc69-48d691d19a3d" />

At 200% scaling:

<img width="2817" height="1222" alt="Image" src="https://github.com/user-attachments/assets/771359ec-a292-4f3d-a705-377e658581f8" />

PR #969, opened 2026-01-19
Fix: Palette width is different on Windows 100% and 200% scaling #965

ptziegler:
Due to 671e2488d076ca2bc5bb1224fdf462536700c398, the flyout palette is no longer scaled with the monitor zoom. This means the width has to be explicitly scaled when performing the layout. Otherwise the palette size no longer matches once the monitor zoom is changed.

Closes https://github.com/eclipse-gef/gef-classic/issues/965

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
@@ -64,7 +64,7 @@ public class InternalDraw2dUtils {
 		}
 		control.setData(InternalDraw2dUtils.DATA_AUTOSCALE_DISABLED, true);
 		control.addListener(SWT.ZoomChanged, e -> zoomConsumer.accept(e.detail / 100.0));
-		zoomConsumer.accept(InternalDraw2dUtils.calculateScale(control));
+		zoomConsumer.accept((double) InternalDraw2dUtils.calculateScale(control));
 	}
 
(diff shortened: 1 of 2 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
```diff
@@ -64,7 +64,7 @@ public static void configureForAutoscalingMode(Control control, Consumer<Double>
 		}
 		control.setData(InternalDraw2dUtils.DATA_AUTOSCALE_DISABLED, true);
 		control.addListener(SWT.ZoomChanged, e -> zoomConsumer.accept(e.detail / 100.0));
-		zoomConsumer.accept(InternalDraw2dUtils.calculateScale(control));
+		zoomConsumer.accept((double) InternalDraw2dUtils.calculateScale(control));
```
ptziegler: I'll probably update this consumer to accept a float instead of double. We don't really need double-precision for zoom changes and floats have the advantage that they return an int when used with `Math.round()`.

ptziegler:
> Before I got a hi-res monitor I would still set the display scaling to higher settings > 100%. Yes, things look a bit huge but it can work for a quick test.

That's what I'm currently doing :) It's just that the only Windows I have is running on my laptop, which also only supports up to  175% monitor zoom. So a lot of the "change zoom to 200%" problems are something I can only try out in a limited capacity. :shrug:

Phillipus:
Should this fix affect Windows only? On Mac if I switch between a version of GEF after this change and then switch to a version before this change the palette width is 2x what it was.

PR #972, opened 2026-01-21
Use SWT Path for painting And/OR and XOR Gate Figures

ptziegler:
The Gate figures are created by combining arcs and lines to create round corners. Doing so makes the figures susceptible to rounding errors, especially when anti-alias is enabled. These errors are visible when painting the images.

Instead of using a combination of shapes, the outline of those figures is represented with an SWT Path. The advantage is that this path describes the entire figure, thus making the structure more readable.

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/OrGateFigure.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/OrGateFigure.java
@@ -13,29 +13,29 @@
 package org.eclipse.gef.examples.logicdesigner.figures;
 
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Path;
 
 import org.eclipse.draw2d.Graphics;
 import org.eclipse.draw2d.geometry.Dimension;
-import org.eclipse.draw2d.geometry.PointList;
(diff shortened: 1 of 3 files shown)


ptziegler:
Side-by-side comparision of the updated OR gate and the XOR gate.

image

org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/OrGateFigure.java
```diff
@@ -60,33 +60,20 @@ public Dimension getPreferredSize(int wHint, int hHint) {
 	@Override
 	protected void paintFigure(Graphics g) {
 		Rectangle r = getBounds().getCopy();
-		r.translate(5, 4);
-		r.setSize(22, 18);
 
-		g.setAntialias(SWT.ON);
```
ptziegler: The anti-alias is what's causing those ugly gaps on Windows. Don't see any other way to fix this, other than disabling it.

ptziegler:
I think this is as good as it can get. There are still some minor artifacts but I don't see how this can be fixed on the Draw2D side.

Screenshot 2026-02-09 201927

azoitl (approved):
I like this improvement a lot. I think we can live for now with the small gap.

PR #976, opened 2026-01-21
Use "menu background selected" color for Marquee selection rectangle

ptziegler:
Instead of using XOR to create a constract between the marquee rectangle and the background, we now use one of the "selection" colors defined by the GEF theme. In light mode, this will be a dark color and a light color, when using dark mode.

On Windows, XOR mode doesn't work reliably when using it together with the advanced mode, leading to an almost invisible rectangle because its outline as painted white.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/964

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/tools/MarqueeSelectionTool.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/tools/MarqueeSelectionTool.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


ptziegler:
> how about changing it to SWT.COLOR_LIST_SELECTION

That's the constant associated with this color. I can't tell you why we call it something else here.

https://github.com/eclipse-gef/gef-classic/blob/5a3c5a1fa1236094b124b62b573728d6558caddc/org.eclipse.draw2d/src/org/eclipse/draw2d/BasicColorProvider.java#L70-L73

Phillipus:
> That is puzzeling me. For 4diac IDE XOR mode is not working since we switched to SWTGraphics around 2021.

We switched to SWTGraphics in 2020. My old fork of GEF/Draw2d is from 2021 and XOR has always worked on all platforms. It's working now in the current version of Archi.

For things like the marquee tool XOR mode will guarantee that it shows whatever the background colour is. But, for reasons due to scaling, advanced mode, and so on it seems that it cannot be used on Windows going forward, even if it does work on Mac and Linux.

PR #978, opened 2026-01-23
Use new SWT autoscale disabledment API

akoch-yatta:
This PR uses the new autoscale disablement API added for SWT in the upcoming version 3.133. To be backwards compatible, it will use reflection and fall back to the previous implementation if an older version of SWT used. This can be cleaned up when GEF uses SWT 3.133 as minimal requirement.

The behavior should not be changed by this PR but just work as before. I utilizes `Shell.getZoom` and `Control.setAutoscalingMode`.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
@@ -34,22 +34,6 @@ public class InternalDraw2dUtils {
 	 */
 	private static final String DRAW2D_ENABLE_AUTOSCALE = "draw2d.enableAutoscale"; //$NON-NLS-1$
 
-	/**
-	 * Internal flag for fetching the shell zoom
-	 */
-	private static final String DATA_SHELL_ZOOM = "SHELL_ZOOM"; //$NON-NLS-1$
-
-	/**
-	 * Data that can be set to scale this widget at 100%.
-	 */
-	private static final String DATA_AUTOSCALE_DISABLED = "AUTOSCALE_DISABLED"; //$NON-NLS-1$
-
-	/**
-	 * Data that can be set to make a control not propagate autoScale disabling to
-	 * children.
-	 */
-	private static final String DATA_PROPOGATE_AUTOSCALE_DISABLED = "PROPOGATE_AUTOSCALE_DISABLED"; //$NON-NLS-1$
-
 	private static final boolean enableAutoScale = "win32".equals(SWT.getPlatform()) //$NON-NLS-1$
 			&& Boolean.parseBoolean(System.getProperty(DRAW2D_ENABLE_AUTOSCALE, Boolean.TRUE.toString()))
 			&& SWT.getVersion() >= 4971; // SWT 2025-12 release or higher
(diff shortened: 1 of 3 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
```diff
@@ -62,16 +103,36 @@ public static void configureForAutoscalingMode(Control control, Consumer<Double>
...
+			// only set the mode, if SWT autoscaling is enabled
+			if (SET_AUTOSCALINGMODE_HANDLE != null) {
+				try {
+					SET_AUTOSCALINGMODE_HANDLE.invoke(control, AUTOSCALING_MODE_DISABLED_INHERITED);
+				} catch (Throwable e) {
+					throw new SWTException(e.getMessage());
+				}
+			} else {
+				control.setData(InternalDraw2dUtils.DATA_AUTOSCALE_DISABLED, true);
+			}
```
ptziegler: Similarly, this can be simplified to `setAutoscalingMode(control, AUTOSCALING_MODE_DISABLED_INHERITED)` if the logic is extracted to a common method.

org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
```diff
@@ -33,27 +38,63 @@ public class InternalDraw2dUtils {
...
+	static {
+		try {
+			// Introduced with SWT 3.133
+			Class<?> autoscalingModeEnumClass = Class.forName("org.eclipse.swt.graphics.AutoscalingMode"); //$NON-NLS-1$
+			if (autoscalingModeEnumClass.isEnum()) {
+				AUTOSCALING_MODE_DISABLED_INHERITED = Enum.valueOf((Class<? extends Enum>) autoscalingModeEnumClass,
+						"DISABLED_INHERITED");
+				assert AUTOSCALING_MODE_DISABLED != null;
+				AUTOSCALING_MODE_DISABLED = Enum.valueOf((Class<? extends Enum>) autoscalingModeEnumClass, "DISABLED");
+				assert AUTOSCALING_MODE_DISABLED_INHERITED != null;
```
ptziegler: Afaik, the `assert` check needs to be explicitly enabled. I'd either check this instance via `Objects.requireNonNull(...)` or skip it altogether.

org.eclipse.draw2d/src/org/eclipse/draw2d/internal/InternalDraw2dUtils.java
```diff
@@ -33,27 +38,63 @@ public class InternalDraw2dUtils {
...
+		}
+	}
+
+	private static MethodHandle SET_AUTOSCALINGMODE_HANDLE;
+	private static Object AUTOSCALING_MODE_DISABLED_INHERITED;
+	private static Object AUTOSCALING_MODE_DISABLED;
+	static {
+		try {
+			// Introduced with SWT 3.133
+			Class<?> autoscalingModeEnumClass = Class.forName("org.eclipse.swt.graphics.AutoscalingMode"); //$NON-NLS-1$
```
ptziegler: This way you don't have to do the type-cast below. You still have to suppress the "unchecked warnings" and add the `//$NON-NLS-1$` comment, though.
```suggestion
			@SuppressWarnings("rawtypes")
			Class<? extends Enum> autoscalingModeEnumClass = Class.forName("org.eclipse.swt.graphics.AutoscalingMode") //$NON-NLS-1$
					.asSubclass(Enum.class);
```

PR #995, opened 2026-01-31
Fine Tuned Drop Shadow

azoitl:
- Made Drop Shadow a bit Softer
 - RectangleDropShadow correctly considers the corner radius for the start of the shadow

### DIFF
--- a/org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/shadows/ShadowsExample.java
+++ b/org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/shadows/ShadowsExample.java
@@ -1,0 +1,114 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Johannes Kepler University Linz
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alois Zoitl - initial API and implementation
+ *******************************************************************************/
+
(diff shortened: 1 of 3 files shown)


ptziegler:
Would it make sense to add a small example, so that people can easily see what these borders look like? This would also make it much simpler to see the before-and-after.

azoitl:
Yes I definitely want to do that for the 206-03 release. I was not yet sure if I use the logic editor or something else.

ptziegler:
> Yes I definitely want to do that for the 206-03 release. I was not yet sure if I use the logic editor or something else.

To be honest, I think a snippet in `org.eclipse.draw2d.examples` would be more than enough. I'm a little hesitant to squeeze everything in the example editors. Otherwise they'll just get way too complicated.

azoitl:
> To be honest, I think a snippet in `org.eclipse.draw2d.examples` would be more than enough. I'm a little hesitant to squeeze everything in the example editors. Otherwise they'll just get way too complicated.

Then I'll do that for just showing the shadows. But they really shine when applied to an editor. I'm in the process of migrating my pure drop shadows in Eclipse 4diac to the new ones and the difference is large.

ptziegler (approved):

azoitl:
I'm currently experiementing a lot with it, as I'm not yet happy with the default values. More tuning is to be expected. 

But before that I want to add some LoD awareness, draw more lines for scaling factors >1 and for very small (e.g. outline views) draw only one darker, for performance reasons.

azoitl:
@ptziegler this is how t looks in a real-world application:
image

The outer shadow is a bit larger then the default and with a stronger exponetial fall-of. The shadow of the inner container is the default shadow. My next step is if I can with the given infrastructure also build a drop shadow for the small blocks. Given their complex shape this will be a bit hard but a a nice test for the AbstractDropShadow class.

PR #997, opened 2026-02-02
Suppress deprecation warnings regarding ScaledGraphics

ptziegler:
The way ScaledGraphics is used is perfectly valid in the IScalablePane and PrinterGraphics and can only be resolved when removing this class.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableLayeredPane.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableLayeredPane.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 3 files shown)


azoitl (approved):

ptziegler:
> I guess we can ignore the Eclipse compiler warnings or?

The warnings are correct. I think it makes more sense if I enable the deprecation warnings now as part of this commit, so that I can then fix them with #998 and an additional one for `fireMoved()`.

org.eclipse.draw2d/src/org/eclipse/draw2d/ScalableLayeredPane.java
```diff
@@ -100,6 +100,7 @@ public boolean optimizeClip() {
 	 * @param newZoom The new zoom level
 	 */
 	@Override
+	@SuppressWarnings("deprecation")
```
ptziegler: I'm not really sure what to do about the `fireMoved()` method, so I think ignoring the deprecations is the safest option for now. We can't really adapt the code as long as this method exist and I don't think it makes sense to mark it for removal before we have a good understanding of the potential implications or even just how to refactor our code.

azoitl: I totally agree with you here.

PR #998, opened 2026-02-02
Mark `VirtualNode` for removal

ptziegler:
This class was deprecated with fb1a1a272e28d3f710a31be7c2d69ee318d9a780 and should be treated exactly as any other Node. By marking it as "for removal", it can be made package-private after the 2028-03 release.


ptziegler:
Included in #997

Issue #999, opened 2026-02-04
Create new ImagePrintOperation class


ptziegler:
Based on https://github.com/eclipse-gef/gef-classic/issues/977#issuecomment-3843327582, I think it would make sense to create an `ImagePrintOperation` class, which allows the user to print an IFigure to an image. When Draw2D-based scaling is enabled, painting an `IFigure` on an `Image` may introduce clipping, if not at 100% monitor zoom. This isn't something the user should be forced to handle on their own.

This class should return an `Image` backed by an `ImageDataProvider` and the `ImageData` is created similar to this:

```java
public Image print(IFigure figure) {
	return new Image(null, new ImageDataProvider() {
		@Override
		public ImageData getImageData(int zoom) {
			return printAtZoom(figure, zoom / 100.0);
		}
	});
}

private ImageData printAtZoom(IFigure figure, double scale) {
	Rectangle bounds = figure.getBounds().getCopy().scale(scale);
	Image image = new Image(null, bounds.width, bounds.height);

	GC gc = new GC(image);
	image.getImageData(100);
	
	SWTGraphics graphics = new SWTGraphics(gc);
	graphics.scale(scale);
	figure.paint(graphics);
	graphics.dispose();
	gc.dispose();
	
	ImageData imageData = image.getImageData();
	image.dispose();
	return imageData;
}
```

ptziegler:
Done via #1001

PR #1001, opened 2026-02-04
Add utility functions for printing figures and graphical viewers #999

ptziegler:
This adds the new utility classes `ImagePrintFigureOperation` and `ImagePrintGraphicalViewerOperation` which can be used to paint the contents of a Figure/GraphicalViewer onto an SWT Image.

When capturing a figure, clipping may occur because text sizes are not consistent across different zoom levels. When Draw2D-based scaling is enabled, the `FigureCanvas` is configured to always draw as if at 100% zoom and then scale its contents to match the actual monitor zoom.

However, when painting on an Image, the Image GC may use the monitor zoom, leading to inconsistencies. To avoid this, the GC must be forced into "100% mode" by first calling `Image.getImageData(100)`. Then the SWTGraphics instance must be scaled by the monitor zoom so that the Figure can be painted.

The result is an exact representation of the source figure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
   This feature was originally request in [Bug 172463](https://bugs.eclipse.org/bugs/show_bug.cgi?id=172463)   
 - Implemented `SWTGraphics.getAbsoluteScale()`
   This feature was originally requested in [Bug 267462](https://bugs.eclipse.org/bugs/show_bug.cgi?id=267462)
+  Added `ImagePrintFigureOperation` for painting a Figure on an Image.
 
 ## GEF
 
@@ -65,6 +66,8 @@ specifically, the `TreeViewer` is using `EditPart`'s as exclusion set.
   contribute their own implementation via an OSGi service.
   Alternatively, clients may inject a custom `ToolTipHelper` instance directly
   by overriding `createToolTipHelper()` in `SWTEventDispatcher`.
+  Added `ImagePrintGraphicalViewerOperation` for painting the printable layer
+  of a GraphicalViewer on an Image.
 
 ## Zest
 
--- a/org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/printing/ImagePrintExample.java
+++ b/org.eclipse.draw2d.examples/src/org/eclipse/draw2d/examples/printing/ImagePrintExample.java
@@ -1,0 +1,96 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.draw2d.examples.printing;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Shell;
+
+import org.eclipse.draw2d.FigureCanvas;
+import org.eclipse.draw2d.ImagePrintFigureOperation;
+import org.eclipse.draw2d.SWTGraphics;
+import org.eclipse.draw2d.ScalableLightweightSystem;
+import org.eclipse.draw2d.text.FlowPage;
+import org.eclipse.draw2d.text.ParagraphTextLayout;
+import org.eclipse.draw2d.text.TextFlow;
+
+/**
+ * This example shows how to use the {@link ImagePrintFigureOperation}.
+ */
+public class ImagePrintExample {
+
+	public static void main(String[] args) {
+		Shell shell = new Shell();
+		shell.setSize(400, 200);
+		shell.setLayout(new GridLayout(3, true));
+
+		TextFlow textFlow = new TextFlow("Hello World"); //$NON-NLS-1$
+		textFlow.setLayoutManager(new ParagraphTextLayout(textFlow, ParagraphTextLayout.WORD_WRAP_SOFT));
+
+		FlowPage flowPage = new FlowPage();
+		flowPage.add(textFlow);
+
+		Label original = new Label(shell, SWT.NONE);
+		original.setText("Original:"); //$NON-NLS-1$
+		original.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
+
+		Label imagePrinter = new Label(shell, SWT.NONE);
+		imagePrinter.setText("ImagePrinter:"); //$NON-NLS-1$
+		imagePrinter.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
+
+		Label image = new Label(shell, SWT.NONE);
+		image.setText("Image:"); //$NON-NLS-1$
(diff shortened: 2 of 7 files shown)


ptziegler:
Side-by-side comparison when drawing the image directly compared to the printer:

Screenshot 2026-02-04 203712

Screenshot 2026-02-04 204900

azoitl (approved):

ptziegler:
fyi @HeikoKlare @Phillipus

ptziegler:
> How could I use the ImagePrintFigureOperation in this case?

I'd simply adapt the  `preparePrintSource()` and `restorePrintSource()` method to also pass along the `SWTGraphics` instance used for painting. Then you can subclass the `ImagePrintFigureOperation` and add the pre-processing to the `preparePrintSource()` method.

Perhaps it makes sense to also add a `getBounds()` method to return the Image size? Otherwise injecting the minimum size would be tricky.

Phillipus:
> I'd simply adapt the preparePrintSource() and restorePrintSource() method to also pass along the SWTGraphics instance used for painting. 

I can't see how that can be done as `createImageAtCurrentZoom` and `printAtZoom` are private and uses its own `SWTGraphics ` instance

ptziegler:
> I can't see how that can be done as createImageAtCurrentZoom and printAtZoom are private and uses its own SWTGraphics  instance

By "I'd simply adapt" I mean that _I'll_ update this PR so that the internal `SWTGraphics` instance is accessible within the prepare/restore methods.

azoitl (approved):
I think this would also be good to have for 3.27, or?

ptziegler:
> Regarding the exception thrown using the Image in a Nebula Gallery Item, this only occurs if `zoom` < 100 at `getImageDataAtZoom`. A workaround is to return null for that one:
> 
> ```java
> protected ImageData getImageDataAtZoom(int zoom) {
> 	if (zoom < 100) {
> 		return null;
> 	}
> 
> 	Image image = createImageAtScale(zoom / 100.0);
> 	try {
> 		return image.getImageData(100);
> 	} finally {
> 		image.dispose();
> 	}
> }
> ```

I managed to reproduce this issue. It happens if you e.g. have an image of size 16x16 and try to get the image data at 1% zoom. The scaled size is (16 * 0.01, 16 * 0.01), which is rounded down to (0, 0) and therefore invalid.

Phillipus:
> I managed to reproduce this issue. It happens if you e.g. have an image of size 16x16 and try to get the image data at 1% zoom. The scaled size is (16 * 0.01, 16 * 0.01), which is rounded down to (0, 0) and therefore invalid.

Right. I'm not sure if my suggested workaround (return null on zoom < 100) is the right solution here.

ptziegler:
I played around with this problem and in the end, the only way I managed to reproduce the exact stack-trace is via this example:

```java
package org.eclipse.swt.tests.manual;

import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.layout.BorderData;
import org.eclipse.swt.widgets.Slider;

public class DrawGC {

		Shell shell = new Shell();
		shell.setSize(100, 100);

		Image image = createTestImage(display, 16);

			event.gc.drawImage(image, 2, 2, 16, 16, 0, 0, 16, 16);
		});

		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}

		image.dispose();
	}

	private static Image createTestImage(Display display, int size) {
		Image image = new Image(display, size, size);
		GC gc = new GC(image);
		gc.setBackground(display.getSystemColor(SWT.COLOR_RED));
		gc.fillRectangle(0, 0, size, size);
		gc.dispose();
		return image;
	}
}
```

The `computeSourceRectangle(...)` method is only called from within `drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight)` and the exception is thrown when

a) the source rectangle differs too much from the target rectangle
b) the monitor zoom is not 100%.

Until I see a reproducer, I would argue that this is a bug in whatever widget is painting the image. Because as far as I can tell, the `computeSourceRectangle` method only considers the source and destination size of the image.

Phillipus:
This is the more complete stack trace with the latest SWT build:

```
java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4915)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.computeSourceRectangle(GC.java:1251)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.lambda$0(GC.java:1223)
	at org.eclipse.swt.graphics.Image.executeOnImageHandleAtBestFittingSize(Image.java:946)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.draw(GC.java:1222)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.apply(GC.java:1204)
	at org.eclipse.swt.graphics.GC.storeAndApplyOperationForExistingHandle(GC.java:6102)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1134)
	at org.eclipse.nebula.widgets.gallery.DefaultGalleryItemRenderer.draw(DefaultGalleryItemRenderer.java:226)
	at org.eclipse.nebula.widgets.gallery.AbstractGridGroupRenderer.drawItem(AbstractGridGroupRenderer.java:267)
	at org.eclipse.nebula.widgets.gallery.NoGroupRenderer.draw(NoGroupRenderer.java:66)
	at org.eclipse.nebula.widgets.gallery.Gallery._drawGroup(Gallery.java:1355)
	at org.eclipse.nebula.widgets.gallery.Gallery.onPaint(Gallery.java:1186)
	at org.eclipse.nebula.widgets.gallery.Gallery.lambda$2(Gallery.java:585)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4363)
```

ptziegler:
The Gallery widget is working on the raw image size to determine some obscure width and height which is passed to the `drawImage(...)` call. Which means that this has nothing to do with the print-operation class. Or rather, it means that this class may produce an image whose bounds break the Gallery calculation.

But I will not go and try out every possible combination of width and height to see which one the widget doesn't like. So as said before: As long as there isn't a reproducer, I consider this to be a Nebula bug.

Phillipus:
> As long as there isn't a reproducer, I consider this to be a Nebula bug.

I tend to agree with that. My reports here are just to document the (potential) issue.

ptziegler:
I don't think there'll be any progress with the Nebula problem unless there's a simple reproducer. If you see this issue again, can you check what "best" size is calculated in the widget and what the initial the bounds are the renderer method is called with?

image

Phillipus:
> I don't think there'll be any progress with the Nebula problem unless there's a simple reproducer. If you see this issue again, can you check what "best" size is calculated in the widget and what the initial the bounds are the renderer method is called with?

Sure. I'll come back to this one at some point. Thanks for your work on this!

Phillipus:
@ptziegler I've implemented the `ImagePrintFigureOperation` in Archi so that a GEF diagram can be used as an `Image` from the printable layer figure of a `GraphicalViewerImpl` (saving to image file or displaying a thumbnail). This fixes the text clipping on Windows.

The problem I see is that if the Images are cached as in the Nebula `Gallery` widget memory usage is much higher. In Archi we can have 100s of thumbnails of diagram images. The increase in memory consumption is because of holding onto the `ImageData` in the `ScaledImageDataProvider`:

https://github.com/eclipse-gef/gef-classic/blob/681ad7a8af6f02a1a83b9d41d65cda96af459fe0/org.eclipse.draw2d/src/org/eclipse/draw2d/ImagePrintFigureOperation.java#L225-L232

Also, there is a performance hit because `Image.getImageData(zoom)` is called several times in  `ImagePrintFigureOperation`

ptziegler:
@Phillipus 

I briefly looking into this and I believe that caching the `ImageData` in the provider is redundant, as the same thing already happens on the SWT side. The main concern when I implemented it was performance, because I don't want to recapture the figure every time I want to paint it. But if that doesn't happen anyways, then this check can go.

image

PR #1002, opened 2026-02-05
Make default Graphics implementation configurable

HeikoKlare:
The default `Graphics` implementation to use was changed from `ScaledGraphics` to `SWTGraphics`. In order to better support the transition from using `ScaledGraphics` to `SWTGraphics`, this change makes the default Graphics implementation configurable. While the default is still changed to `SWTGraphics`, a system property is added that allows to change the default back to ScaledGraphics.

Also see the following and related comments in the `ScaledGraphics` deprecation PR:
- https://github.com/eclipse-gef/gef-classic/pull/901#issuecomment-3854316386

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
   
   As part of this deprecation also all methods and flags allowing to switch to between `SWTGraphics` and 
   `ScaledGraphics` are deprecated and the default behavior is changed to use `SWTGraphics`.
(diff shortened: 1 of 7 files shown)


azoitl (commented):
I like that approach. Could you also update the Changelog entry with this property.

azoitl (approved):
@HeikoKlare thx for the PR and the changelog update.

@ptziegler WDYT?

PR #1003, opened 2026-02-08
A HSL Color Representation

azoitl:
HSL color representations can be helpful for different color calculations (e.g., correct lighter or darker colors, contrast colors).

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
@@ -22,6 +22,7 @@ import org.eclipse.swt.graphics.FontMetrics;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.widgets.Shell;
 
+import org.eclipse.draw2d.colors.HSLColor;
 import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.draw2d.geometry.Rectangle;
 import org.eclipse.draw2d.internal.InternalDraw2dUtils;
@@ -31,7 +32,6 @@ import org.eclipse.draw2d.internal.InternalDraw2dUtils;
  */
 public class FigureUtilities {
 
-	private static final float RGB_VALUE_MULTIPLIER = 0.6f;
 	private static GC gc;
 	private static Font appliedFont;
 	private static FontMetrics metrics;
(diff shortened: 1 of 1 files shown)


ptziegler (approved):
You're probably right. Just because we don't work with HLS colors right now doesn't mean we won't in the future.

azoitl:
@ptziegler an idea that I had today is that I turn the HSL record class into a normal class. Then I could add a lazy created field with the RGB color. This would have the advantage that you could use HSL colors efficiently as drop in replacement for SWT Color. The what I'm not sure if it should be a value class (non modifyble HSL color fields) or if methods like lighter or darker should modify the HSL color and not create a new one. 

WDYT @ptziegler?

org.eclipse.draw2d/src/org/eclipse/draw2d/colors/HSLColor.java
```diff
@@ -0,0 +1,213 @@
...
+
+		if (hue < 60.0) {
+			retVal = m1 + (((m2 - m1) * hue) / 60.0);
+		} else if (hue < 180.0) {
+			retVal = m2;
+		} else if (hue < 240.0) {
+			retVal = m1 + (((m2 - m1) * (240.0 - hue)) / 60.0);
+		}
+
+		return (int) Math.round(MAX_RGB_VALUE * retVal);
```
ptziegler:
```suggestion
		return new RGB((float) h, (float) s, (float) l);
```

azoitl: There are way to many different color representations ;-) In 4diac IDE I also have an YUV color description. This is helpfull when you want to find out if two colors are distinguishable by humans.

ptziegler:
> `getColor()` would on demand create a color and cache it in the HSL color. With that users could apply HSL color modification means or generation means lighter, darker, mixing, as they are maybe used from CSS or similar tools.
> ```java
> void setForgoundColor(HSLColor hsl) { 
>  setForgroundColor(hsl.getColor);   
>}
>```

I don't like that idea at all, because now you're adding an implementation detail (how the color is encoded) to the interface. To me, the handling of those colors as descriptions makes the most sense, with the ability to convert them to `RGB` and subsequently to `Color`s so that it's compatible with the remaining API.

> getColor() would on demand create a color and cache it in the HSL color. With that users could apply HSL color modification means or generation means lighter, darker, mixing, as they are maybe used from CSS or similar tools.

Wouldn't it be simpler to store the converted SWT color as a local variable in the figure/edit-part?

PR #1004, opened 2026-02-10
Upgraded Eclipse Tycho to 5.0.2

azoitl:


azoitl:
> Looks like you were faster than me. This version uses the correct ECJ version by default, so this workaround with the system property is no longer needed. Should we remove it?
> Image

Yes I think this is a good idea. Will do it.

PR #1009, opened 2026-02-15
Introduce LED Font via Eclipse Theme API

azoitl:
In order to have a defined font on all supported OSes an own logic editor theme font for the font to used by the LED figure is introduced. The values are based on the JFace text editor. We cannot directly use the text colors as text editor zooming is changing the values and the default colors have the wrong size.

@ptziegler during the work on this PR I noticed that there are images for the numbers to be displayed by the LED figure in: https://github.com/eclipse-gef/gef-classic/tree/master/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/icons

I also checked the history and noticed that these where used before switching to the font. Now with SVG images in place we could reconsider switching back to the images. As the logic editor is for me a bit a show of of best practices and features of GEF Classic both would fine. WDYT?

### DIFF
--- a/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/edit/LEDEditPart.java
+++ b/org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/edit/LEDEditPart.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -19,7 +19,11 @@ import java.util.List;
 
 import org.eclipse.swt.accessibility.AccessibleControlEvent;
 import org.eclipse.swt.accessibility.AccessibleEvent;
+import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.Image;
+
+import org.eclipse.jface.resource.FontDescriptor;
+import org.eclipse.jface.resource.JFaceResources;
 
 import org.eclipse.draw2d.ConnectionAnchor;
 import org.eclipse.draw2d.IFigure;
@@ -30,7 +34,6 @@ import org.eclipse.gef.AccessibleEditPart;
 import org.eclipse.gef.EditPolicy;
 
 import org.eclipse.gef.examples.logicdesigner.LogicMessages;
-import org.eclipse.gef.examples.logicdesigner.figures.FigureFactory;
 import org.eclipse.gef.examples.logicdesigner.figures.LEDFigure;
 import org.eclipse.gef.examples.logicdesigner.model.LED;
 
@@ -41,6 +44,7 @@ public class LEDEditPart extends LogicEditPart {
 
 	private static Image LED_SEL_PRIM_BG;
 	private static Image LED_SEL_SECD_BG;
+	private static FontDescriptor ledDisplayFontDescriptor = null;
 
 	private static Image createImage(String name) {
 		try (InputStream stream = LEDFigure.class.getResourceAsStream(name)) {
@@ -80,7 +84,9 @@ public class LEDEditPart extends LogicEditPart {
 	 */
 	@Override
 	protected IFigure createFigure() {
-		return FigureFactory.createNewLED();
+		IFigure ledFigure = new LEDFigure();
+		ledFigure.setFont(getLEDDisplayFont());
+		return ledFigure;
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


ptziegler:
> I also checked the history and noticed that these where used before switching to the font. Now with SVG images in place we could reconsider switching back to the images.

I'm not sure if going back to images is really the way to go, because unlike fonts, you have no way to respond to themes. The numbers will then always be yellow, regardless of whether you're using the light or black theme.

org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/LEDFigure.java
```diff
@@ -35,7 +36,8 @@
 public class LEDFigure extends NodeFigure implements HandleBounds {
 
 	public static final Dimension SIZE = new Dimension(92, 71);
-	protected static final Font DISPLAY_FONT = new Font(null, "DialogInput", 24, SWT.NORMAL); //$NON-NLS-1$
+	protected static final Font DISPLAY_FONT = JFaceResources.getFontRegistry()
+			.get("org.eclipse.gef.examples.logic.theme.ledFont"); //$NON-NLS-1$
```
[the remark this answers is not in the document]
azoitl: > What about the `TEXT_FONT` then? That should internally be mapped to Monospace. The point I'm trying to make is: Why reinvent the wheel when JFace already provides those default fonts.

I would also be fine with this. The question I have is what do we want to show. How to use default Eclipse fonts or how to add your own font description using the theme api which can be adjusted by users. I went for the later as it took me some time to do it correctly in 4diac IDE.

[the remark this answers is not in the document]
ptziegler: > The question I have is what do we want to show. How to use default Eclipse fonts or how to add your own font description using the theme api which can be adjusted by users.

The existing constants are usually more than enough for almost all all use-cases. So I think they should be used as often as possible, instead of re-inventing the wheel every time. Using a custom font for something that is effective identical to the JFace font feels more like a _bad_ example, even if it is intended to show how the extension point works.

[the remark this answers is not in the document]
azoitl: I would see it a nice example if the font should be changeable by the user. But for the LED text I don't think this is something that should be done. Changed accordingly.

org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/edit/LEDEditPart.java
```diff
@@ -80,7 +84,7 @@ protected void createEditPolicies() {
 	 */
 	@Override
 	protected IFigure createFigure() {
-		return FigureFactory.createNewLED();
+		return new LEDFigure(getLEDDisplayFont());
```
ptziegler: Out of interest, how do you feel about simply calling `setFont()` after the figure has been created?

```
@Override
protected IFigure createFigure() {
    IFigure figure = FigureFactory.createNewLED();
    figure.setFont(getLEDDisplayFont());
    return figure;
}
```

The only text that is drawn are the digits and then you don't have to pass the font all the way down.

azoitl: Originally I thought that the LED font is used only by parts of the figure therefore I kept it separate. But now after reading your feedback twice I see that this is the much cleaner solution. Will change it.

ptziegler (approved):
Thanks! Should we merge it now or postpone this for the next release? It doesn't affect the core plugins but still adds unnecessary noise so late into the release.

org.eclipse.gef.examples.logic/src/org/eclipse/gef/examples/logicdesigner/figures/LEDFeedbackFigure.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000 IBM Corporation and others.
```
ptziegler: Any reason why you removed this copyright year?

ptziegler: That sounds like an incredibly poor decision... So unless you have the whole project history at hand, you have no clue whether a file is still actively worked on?

I suppose it makes sense in so far that people forget to update it 99% of the time. But that doesn't mean I like it.

PR #1010, opened 2026-02-16
Rename `HSLColor` to `HSL`

ptziegler:
This record represents a color description (similar to RGB) and not an actual color. Renaming the class avoids any confusion with the SWT `Color`.


azoitl (changes requested):
While I totally agree to call it HSL and not implement SWT Color features. I would really prefere to keep it a record. I find the idea with the public fields really not good.

PR #1013, opened 2026-02-17
Optimize Drop Shadow for small Zooms

azoitl:
When far zoomed out drawing the drop shadow in full detail is only wasting performance. Therefore we are now only drawing a single line darker to give some contour.


ptziegler (approved):
Looking good 👍 Do you want to merge it now or wait until the next release?

PR #1016, opened 2026-02-17
Fix erroneous autoscale disablement on disabling disablement propagation

HeikoKlare:
With the recent adoption of the SWT autoscale disablement API the behavior of the method `InternalDraw2dUtils.setPropagateAutoScaleDisabled()` was changed. While it only changed whether autoscale disablement is propagated _if_ autoscaling is disable it now directly disabled is.

This derivation in behavior is reverted by this change, which ensures that the intent to not propagate disablement is simply cached for the case that actually autoscaling is disabled. A slight remaining derivation in behavior is that autoscale propagation cannot be disabled after autoscaling has been disabled for a control, but that flow was not used before anyway.

Fixes https://github.com/eclipse-gef/gef-classic/issues/1014


ptziegler (approved):
The change makes sense and works as expected.

PR #1019, opened 2026-02-18
Deprecate XOR-related methods in `Shape`

ptziegler:
This is a continuation of f7a00b18852d83a98b4fb11aea0d9c6f4186cba5. XOR is unsupported on Windows when GDI+ (advanced mode) is enabled and returns a black shadow, rather than a blending of the figure and its background. Due to the support for fractional scaling (and the deprecation of `ScaledGraphics`), advanced mode is almost always active on Windows.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/FigureUtilities.java
@@ -219,6 +219,8 @@ public class FigureUtilities {
 	 * @param s the shape
 	 * @return the ghosted shape
 	 * @since 2.0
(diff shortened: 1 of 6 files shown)


azoitl:
For that problem. HSL shines. There are several options to create contrast or complementary colors, either by tweaking the hue part or by working with saturation and luminosity combined. 

In 4diac IDE we use now for quite some time the selection color for the border and fill with selection color with alpha. Even with colored background and in darkmode this works really good. We have a border of width 2 and the alpha is 50.

Phillipus:
Using a `RectangleFigure` leads to the same problem I reported in https://github.com/eclipse-gef/gef-classic/issues/1020 The rectangle line width is clipped on the top and left sides:

Image 003

Issue #886, opened 2025-12-06
[Umbrella] Text Editor Example


ptziegler:
This items primarily serves to keep track of everything that is wrong with our Text Editor:

- IllegalArgumentException upon opening the editor

When the editor is openend, the following exception is logged:

```
java.lang.IllegalArgumentException: Offset: -1 is invalid
	at org.eclipse.draw2d.text.TextFlow.getCaretPlacement(TextFlow.java:257)
	at org.eclipse.gef.examples.text.edit.TextFlowPart.getCaretPlacement(TextFlowPart.java:82)
	at org.eclipse.gef.examples.text.tools.TextTool.getCaretInfo(TextTool.java:514)
	at org.eclipse.gef.examples.text.tools.TextTool$CaretRefresh.refreshCaret(TextTool.java:880)
	at org.eclipse.gef.examples.text.tools.TextTool$CaretRefresh.run(TextTool.java:869)
```

- Bold/Italics/Underline actions can't be undone

Selecting a bold text and clicking the "bold" action adds invalid elements to the model, rather than removing the bold element.

<img width="160" height="98" alt="Image" src="https://github.com/user-attachments/assets/f4324498-f399-48ef-84e6-f4d6d3dd9cc0" />

- Bold/Italics/Underline actions should only be active when a text area is selected

Currently they are always active.

PR #1027, opened 2026-02-23
Fix IllegalArgumentException in TextEditor when opened twice #886

ptziegler:
When opening the Text editor, the caret is moved to the first character of the first word. But in order to calculate this location, the layout has to be applied, because it internally creates the text fragments of the underlying `TextFlow`.

This step is required because the Caret location is determined by the call to `getNextVisibleOffset()` in `TextFlow`, which returns -1, if no fragments exist. When the Caret is refreshed in the `TextTool`, this value is then passed via the `TextFlowPart` to the `TextFlow` which then throws the `IllegalArgumentException`.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/886

### DIFF
--- a/org.eclipse.gef.examples.text/src/org/eclipse/gef/examples/text/TextEditor.java
+++ b/org.eclipse.gef.examples.text/src/org/eclipse/gef/examples/text/TextEditor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2025 IBM Corporation and others.
+ * Copyright (c) 2004, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


ptziegler:
Related to this exception here:

```
java.lang.IllegalArgumentException: Offset: -1 is invalid
	at org.eclipse.draw2d.text.TextFlow.getCaretPlacement(TextFlow.java:257)
	at org.eclipse.gef.examples.text.edit.TextFlowPart.getCaretPlacement(TextFlowPart.java:82)
	at org.eclipse.gef.examples.text.tools.TextTool.getCaretInfo(TextTool.java:514)
	at org.eclipse.gef.examples.text.tools.TextTool$CaretRefresh.refreshCaret(TextTool.java:880)
	at org.eclipse.gef.examples.text.tools.TextTool$CaretRefresh.run(TextTool.java:869)
```

As a side-note: The caret isn't drawn at all on Linux. On Windows, it's shown when the editor is opened for the first time, but not when opened again, due to this error.

Issue #1030, opened 2026-02-24
Inherited methods in PrinterGraphics are marked as deprecated


ptziegler:
`PrinterGraphics` extends the deprecated `ScaledGraphics` class. Any method that isn't overridden in `PrinterGraphics` is therefore also marked as deprecated. See:

https://github.com/eclipse-gef/gef-classic/blob/2a97b978ad4ea0f908ea1705eab4f157f1a44d45/org.eclipse.draw2d.examples/sandbox/swt/bugs/Bug81467.java#L32-L37

...which causes the following warnings:

> The method drawString(String, int, int) from the type ScaledGraphics is deprecated
> The method scale(double) from the type ScaledGraphics is deprecated
> The method setFont(Font) from the type ScaledGraphics is deprecated

ptziegler:
I'll try to get this fixed for the RC2.

PR #1031, opened 2026-02-24
Undeprecate public methods in ScaledGraphics #1030

ptziegler:
The deprecated `ScaledGraphics` class is extended by the `PrintGraphics` class. If a method defined by the `Graphics` class hasn't been overridden in the `PrinterGraphics`, this deprecation warning is propagated, even though the `PrinterGraphics` is not deprecated.

Closes https://github.com/eclipse-gef/gef-classic/issues/1030


azoitl (approved):

ptziegler:
I don't think there is a way to suppress the "perhaps this member should be marked as deprecated, too?" warning specifically. So I simply ignore all warnings for the `ScaledGraphics` methods.

PR #1032, opened 2026-02-28
Fixed index out of Bound exception on Graph Demos

azoitl:
The shuffle operation used an random number generator which could also generate negative numbers.


ptziegler:
Does shuffling even _do_ anything here? Even after playing around with the example for a while, the graphs look identical, regardless of whether they're shuffled or not...

PR #1049, opened 2026-03-10
Extract "Custom Palette Color" snippet from Shapes editor

ptziegler:
With 6e5b5b94cf8e1e3cfee51e32f4b5389ad93df6a6, the Shapes editor was extend to switch the theming of the palette via an entry in the main menu. Given that this example has nothing to do with the editor, it is better suited as a standalone snippet.


ptziegler:
image

PR #1056, opened 2026-03-12
Require Java 21 for example and test projects

ptziegler:
When trying to run the tests and/or snippets using the 2026-03 release repository while having a Java 17 installed, it is no longer possible to execute them due to a "class has been compiled by a more recent version" error caused by the recent SWT plugin included in the classpath.


azoitl (approved):
I will not start the discussion of bringing everything to Java21 🙊 

But jokes aside I think this is a good step in the right direction. Especially it allows us to show how with modern Java some GEF Classic and Draw2d shall be consumed.

Phillipus:
Are there any plans now to bump the BREE of the projects to 21?

ptziegler:
> Are there any plans now to bump the BREE of the projects to 21?

Not right now, no. The BREE of the core plugins will likely only get updated if

a) We require at least Eclipse 2026-03
b) There are Java 21+ language features that drastically reduce the maintenance effort
c) There are changes that would break compatibility with Java 17.

Issue #1055, opened 2026-03-12
GraphViewer area is 2x too large on Windows with higher display scales


Phillipus:
Windows 11
Monitor display 200%

In Archi we use a `GraphViewer` to visualise objects and their relationships. When selecting an object in the UI the `GraphViewer` is updated. Normally all nodes and connections fit into the view area:

<img width="1466" height="268" alt="Image" src="https://github.com/user-attachments/assets/f5d2f315-5e71-412a-922a-20e87b1874ca" />

However, on Windows at monitor display scales > 100% the view area is larger causing the nodes and connections to be off screen and scroll bars required, i.e it doesn't fit into the given window bounds. Here it is at 200% monitor display:

<img width="2935" height="540" alt="Image" src="https://github.com/user-attachments/assets/01cca4f1-c0cf-46bb-9f8d-f63a358e4ebc" />

PR #1057, opened 2026-03-12
Consider monitor zoom for layout bounds in Graph/GraphViewer #1055

ptziegler:
The preferred size (which is used to describe the bounds for the layout algorithm) of the Graph must be scaled by the underlying monitor zoom, to make sure that the graph items are contained by the size of the Graph.

Closes https://github.com/eclipse-gef/gef-classic/issues/1055


Phillipus:
Sorry to report that this PR has the opposite effect on Mac. The bounds are quarter of what they should be.

ptziegler:
> On Mac InternalDraw2dUtils.calculateScale(this) also returns 2 on Mac with retina display but it's not the right factor to use here. I don't know if the check should be platform dependent?

It's effectively the same issue as in https://github.com/eclipse-gef/gef-classic/pull/969#issuecomment-3979526278. In short, the call should be Windows-only.

PR #1061, opened 2026-03-15
Expand getBounds() by 1 pixel in Polygon and Polyline decorations

Phillipus:
- See https://github.com/eclipse-gef/gef-classic/issues/1047


azoitl (changes requested):
LGTM. But @Phillipus coudl you please remove the archi comments.

Phillipus:
> But @Phillipus coudl you please remove the archi comments.

Yikes, thanks for pointing that out. Removed. :-)

PR #1084, opened 2026-04-08
[Zest] Fix memory-leak in GraphLabel when images are cached

ptziegler:
The Zest `GraphLabel` has a potential memory leak when the `NODES_CACHE_LABEL` graph style is used.

When this style is set, the `CachedLabel` will create an internal `Image` onto which the label of this figure is painted. This `Image` is not disposed when the figure is destroyed.

If such an image is created, it will now be registered to an `ImageRegistry`. The lifecycle of this registry is bound to the underlying `Graph`, which will make sure that the `Images` are destroyed, once the `Graph` is disposed.


ptziegler:
Now that I think about it, I wonder if I can just override `removeNotify()` to dispose the cached image... But that method won't be called when the Graph is disposed.

PR #1086, opened 2026-04-15
Remove redundant device parameter from Color constructors

Phillipus:
Since SWT version 3.115 it is not necessary to use the Color(Device, int, int, int) constructor

It may be that the old Color constructor will be deprecated in the near future


azoitl (approved):

Phillipus:
Added more instances with latest force push.

vogella:
LGTM, I handle the rest

PR #1088, opened 2026-04-17
feat: implement Iterable interface for PointList

gravemalte:
Allows `PointList` to be used in enhanced for-each loops. 

Added `PointListIterator` to handle internal primitive array traversal.

Note on each `next()` a `Point` object is created. Another way would be that we use a flyweight, but I'm unsure if it fits, because then we only get a reference and we delegate the saving to the user with our API then. We would then just fill the correct `x` and `y` into the flyweight.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -10,8 +10,12 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Alexander Shatalin (Borland) - Contribution for Bug 238874
+ *     Malte Grave - Added Iterator and Iterable API
  *******************************************************************************/
 package org.eclipse.draw2d.geometry;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 /**
  * Represents a List of Points. This class is used for building an
@@ -19,7 +23,7 @@ package org.eclipse.draw2d.geometry;
  * the client using {@link Point Points}. SWT uses integer arrays when painting
  * polylines and polygons.
  */
-public class PointList implements java.io.Serializable, Translatable {
+public class PointList implements java.io.Serializable, Translatable, Iterable<Point> {
 
 	private int[] points = {};
 	private Rectangle bounds;
@@ -55,6 +59,35 @@ public class PointList implements java.io.Serializable, Translatable {
 	 */
 	public PointList(int size) {
 		points = new int[size * 2];
+	}
+
+	/**
+	 * An {@link Iterator} implementation for {@link PointList}.
+	 * <p>
+	 * This iterator traverses the internal primitive array and returns each
+	 * coordinate pair as a new {@link Point} object.
+	 * </p>
+	 *
+	 * @see java.lang.Iterable#iterator()
+	 * @see org.eclipse.draw2d.geometry.PointList
+	 * @since 3.28
+	 */
+	class PointListIterator implements Iterator<Point> {
+		private int currentIndex;
+
+		@Override
+		public boolean hasNext() {
+			return currentIndex < size();
+		}
+
+		@Override
+		public Point next() {
+			if (!hasNext()) {
+				throw new NoSuchElementException();
+			}
+			return getPoint(currentIndex++);
+		}
+
 	}
 
 	/**
@@ -278,6 +311,22 @@ public class PointList implements java.io.Serializable, Translatable {
 			}
 		}
 		return false;
+	}
+
+	/**
+	 * Returns an iterator over the points in this list.
+	 * <p>
+	 * This implementation provides a {@link PointListIterator} that converts the
+	 * internal primitive coordinates into {@link Point} objects.
+	 * </p>
+	 *
+	 * @return an {@link Iterator} of {@link Point}s
+	 * @see java.lang.Iterable#iterator()
+	 * @since 3.28
+	 */
+	@Override
+	public Iterator<Point> iterator() {
+		return new PointListIterator();
 	}
 
 	/**


org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
```diff
@@ -57,6 +61,35 @@ public PointList(int size) {
...
+	 * This iterator traverses the internal primitive array and returns each
+	 * coordinate pair as a new {@link Point} object.
+	 * </p>
+	 *
+	 * @see java.lang.Iterable#iterator()
+	 * @see org.eclipse.draw2d.geometry.PointList
+	 * @since 3.28
+	 */
+	class PointListIterator implements Iterator<Point> {
+		private int currentIndex = 0;
```
merks: Initializing to 0 is pointless and generates pointless bytecode

gravemalte: Thanks for pointing that out. Fixed it.

merks: It's a tiny nit.  Thanks for doing it anyway.

gravemalte: Happy to learn.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2023, 2026 IBM Corporation and others.
```
azoitl: Please remove the 2023 we only have first year last year.

azoitl (changes requested):

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
```diff
@@ -57,6 +61,35 @@ public PointList(int size) {
...
+	 * @see java.lang.Iterable#iterator()
+	 * @see org.eclipse.draw2d.geometry.PointList
+	 * @since 3.28
+	 */
+	class PointListIterator implements Iterator<Point> {
+		private int currentIndex;
+
+		@Override
+		public boolean hasNext() {
+			return currentIndex < PointList.this.size * 2;
```
ptziegler:
```suggestion
			return currentIndex < size();
```

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
```diff
@@ -57,6 +61,35 @@ public PointList(int size) {
...
+		}
+
+		@Override
+		public Point next() {
+			if (!hasNext()) {
+				throw new NoSuchElementException();
+			}
+			return new Point(PointList.this.points[currentIndex++], PointList.this.points[currentIndex++]);
```
ptziegler:
```suggestion
			return getPoint(currentIndex++);
```

ptziegler (commented):
Accessing the `points` array directly means that you'll only be able to iterate over the integer-part of a point, even when using a `PrecisionPointList`. By calling `getPoint()`, you avoid this problem because this method is overridden to return a `PrecisionPoint`, including the fractional part.

> Another way would be that we use a flyweight, but I'm unsure if it fits, because then we only get a reference and we delegate the saving to the user with our API then. We would then just fill the correct x and y into the flyweight.

I've only seen iterators where `next()` returns a new object. And I would argue that, even if not explicitly documented, this is the general expectation. If you were to re-use the same object and write something like this:

```java
List<Point> points = new ArrayList<>();
for (Point point : pointList) {
    points.add(point);
}
```

... you end up with a list containing the same point over and over again. It might make sense to enable such a behavior for performance-critical operations,  but I think the current behavior should be the default.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
gravemalte: Done.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
gravemalte: Added it.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
gravemalte: I didn't see that method, thanks for pointing out!

gravemalte:
> I've only seen iterators where `next()` returns a new object. And I would argue that, even if not explicitly documented, this is the general expectation. If you were to re-use the same object and write something like this:
> 
> ```java
> List<Point> points = new ArrayList<>();
> for (Point point : pointList) {
>     points.add(point);
> }
> ```
> 
> ... you end up with a list containing the same point over and over again. It might make sense to enable such a behavior for performance-critical operations, but I think the current behavior should be the default.

Exactly that was my idea behind the flyweight idea, but I agree, this should be only used in very performance critical applications, if we want such a behavior which can be enabled by a factory.

gravemalte:
@azoitl and @ptziegler I just saw that I explicitly used `import` but they're a lot of place in GEF where we prefix the used class, is there a guideline?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
ptziegler: You don't need the * 2 anymore. This was only needed because previously, you used to increment `currentIndex` twice.

ptziegler:
> I just saw that I explicitly used import but they're a lot of place in GEF where we prefix the used class, is there a guideline?

As far as I remember, the fully qualified name is only used in the documentation (or if there is ambiguity). But I actually don't know _why_ it was done like this. Nonetheless, in almost all cases you want the import statement and then reference the classes by its simple name.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/PointList.java
gravemalte: Stupid mistake from my side, didn't check it properly.

azoitl (approved):

ptziegler (approved):

ptziegler:
@gravemalte Apologies. I updated your branch but forgot to merge the changes.

ptziegler:
> I just saw that I explicitly used import but they're a lot of place in GEF where we prefix the used class, is there a guideline?

I got cursious because I saw the fully-qualified name primarily for `java.io.Serializable`. There use to be a `sunw.io.Serializable`, so it was used to avoid ambiguity. My original point still stands though. Unless there is such an ambiguity, the type should be imported.

PR #1093, opened 2026-04-22
[PoC] Make Graph figures customizable

ptziegler:
This adds a new IFigureFactory interface which makes it possible to use custom figures for visualization.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=242918


ptziegler:
Most of the customization is already possible via e.g. the `IConnectionStyleProvider` or `IEntityStyleProvider` (when using the GraphViewer). When building the Graph with pure SWT, one can call the methods directly when creating the widgets. I don't think it makes sense to add a different way that runs orthogonal to the existing implementation.

PR #1094, opened 2026-04-22
Fix several issues with the Zest SubGraph feature

ptziegler:
1) If the sub-graph is opened and a node moved, a NullPointerException is thrown because the underlying NodeLayout has already been disposed. Rather than disposing this layout when removing it from the Graph, it is now disposed together with the node itself.

2) When opening the sub-graph while no layout algorithm is set, a NullPointerException is thrown when trying to set the layout context.

3) Rather than modifying the SWT structure of the graph, a sub-graph now only modifies the Zest layer. When switching to the sub-graph, all nodes of the graph are removed and replaced with the nodes of the graph container.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
@@ -31,12 +31,10 @@ import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Image;
-import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Item;
-import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swt.widgets.Widget;
 
 import org.eclipse.zest.core.viewers.internal.ZoomManager;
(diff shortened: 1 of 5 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/widgets/Graph.java
```diff
@@ -885,11 +885,10 @@ public void mouseDoubleClicked(org.eclipse.draw2d.MouseEvent me) {
 			if (itemUnderMouse instanceof GraphContainer container) {
 				// GraphContainer under this mouse
 
-				Display display = Display.getCurrent();
-				Shell shell = display.getActiveShell();
+				Shell shell = getShell();
```
ptziegler: @azoitl Isn't this approach fundamentally flawed? We can't just take the active Shell, replace the layout and add a new Graph. What if the graph is embedded into an RCP application, rather than being a stand-alone snippet?

The new Graph also shouldn't be added to the Shell, but to the parent of the old Graph, ideally inheriting the layout constraints. In addition, the two `new Shell()`s are never disposed, so we're just leaking resources everywhere...

ptziegler: We probably shouldn't be creating a new Graph in the first place... Instead the children of the graph should be replaced with the children of the container. In any case, you also have to consider the edge-case where the graph is disposed while the sub-graph is opened. In that case all removed items also need to be disposed.

ptziegler: Why even bother with changing the SWT structure. Wouldn't it be sufficient to "simply" remove the other nodes from the Zest layer?

ptziegler:
I've updated the PR so that switching to the sub-graph only updates the figures on the root layer and doesn't touch the SWT structure anymore. This should also boost performance as any tree modifications are incredibly slow on Linux.

Technically this can be further simplified by swapping out the root layer of the graph with the root layer of the graph container, but there I'm a little bit concerned about potential side-effects, as in some places we explicitly check against the graph root layer.

ptziegler:
Putting it as a draft because there is a special handling for connection figures in contains I need to further investigate.

ptziegler:
I figured it out. For containers, we may add connections directly to its figure rather than the connection layer. Which must be taken into consideration when moving into the sub-graph.

PR #1095, opened 2026-04-23
Explicitly erase drop shadow border together with figure

ptziegler:
The shadow of the drop-shadow border is drawn outside of the figure bounds. As a result, this overpaint is not removed when the figure is erased.

To fix this problem, a new `eraseBorder()` method is added to the `BackgroundBorder` interface, which is called within the `erase()` method of the figure. This method is overridden for the drop-shadow classes to also mark the shadow as dirty.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
@@ -379,6 +379,9 @@ public class Figure implements IFigure {
 		}
 
 		Rectangle r = new Rectangle(getBounds());
+		if (getBorder() instanceof BackgroundBorder border) {
+			border.eraseBackground(this, r);
+		}
 		getParent().translateToParent(r);
 		getParent().repaint(r.x, r.y, r.width, r.height);
 	}
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/shadows/AbstractDropShadowBorder.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/shadows/AbstractDropShadowBorder.java
@@ -220,10 +220,19 @@ public abstract class AbstractDropShadowBorder extends AbstractBackgroundBorder 
 	}
 
 	private void updateClip(Graphics graphics, Rectangle shadowRect) {
-		int shadowSize = Math.max(haloSize, dropShadowSize) + 1;
+		int shadowSize = getShadowSize();
 		shadowRect.expand(shadowSize, shadowSize);
 		graphics.setClip(shadowRect);
 		shadowRect.shrink(shadowSize, shadowSize);
 	}
 
+	@Override
+	public void eraseBackground(IFigure figure, Rectangle area) {
+		int shadowSize = getShadowSize();
+		area.expand(shadowSize, shadowSize);
+	}
+
+	private int getShadowSize() {
+		return Math.max(haloSize, dropShadowSize) + 1;
+	}
 }
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/BackgroundBorder.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/BackgroundBorder.java
@@ -19,6 +19,7 @@ import org.eclipse.draw2d.Border;
 import org.eclipse.draw2d.Graphics;
 import org.eclipse.draw2d.IFigure;
 import org.eclipse.draw2d.geometry.Insets;
+import org.eclipse.draw2d.geometry.Rectangle;
 
 /**
  * An interface for a special border which can paint both underneath and on top
@@ -45,4 +46,30 @@ public interface BackgroundBorder extends Border {
 	 */
 	void paintBackground(IFigure figure, Graphics graphics, Insets insets);
 
+	/**
+	 * Called when this background should be erased. If the background is drawn
+	 * outside of the figure bounds, the area needs to be expanded to avoid visual
+	 * artifacts. Does nothing by default.
+	 * <p>
+	 * Note: This method is called within the {@link IFigure#erase()} method before
+	 * {@link IFigure#repaint(int, int, int, int)}. Subclasses only need to extend
+	 * the rectangle to cover the area painted by the background.
+	 * </p>
+	 *
+	 * Example:
+	 *
+	 * <pre>
+	 * public void eraseBackground(IFigure figure, Rectangle area) {
+	 * 	area.expand(1, 1);
+	 * }
+	 * </pre>
+	 *
+	 * @param figure The figure whose representation is being erased. Never
+	 *               {@code null}.
+	 * @param area   The area of the figure that is being erased (in parent
+	 *               coordinates)
+	 */
+	default void eraseBackground(IFigure figure, Rectangle area) {
+		// nothing to erase
+	}
 }


ptziegler:
In response to https://github.com/eclipse-gef/gef-classic/issues/1090#issuecomment-4288363094. This is pretty much a prerequisite when using the drop-shadow border in movable figures.

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
```diff
@@ -381,6 +381,10 @@ public void erase() {
 		Rectangle r = new Rectangle(getBounds());
 		getParent().translateToParent(r);
 		getParent().repaint(r.x, r.y, r.width, r.height);
+		if (getBorder() instanceof BackgroundBorder border) {
+			getParent().translateToParent(r);
```
azoitl: why is a second translate to parent needed?

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
```diff
@@ -381,6 +381,10 @@ public void erase() {
...
+			border.eraseBackground(getParent(), r);
```
azoitl: what if eraseBackground would modify r? Then we could stick with the single repaint from line 383

azoitl (commented):
I like this approach a lot. This allows us to keep the shadow outside and fixes repaints on windows. Great!

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
ptziegler: The dirty area is intersected by the figure bounds in the `DeferredUpdateManager`:

https://github.com/eclipse-gef/gef-classic/blob/54b65c713fb668827991ee53b791c88d279ee436/org.eclipse.draw2d/src/org/eclipse/draw2d/DeferredUpdateManager.java#L279-L281

This means that any dirty area outside of the figure bounds is effectively ignored. So the area of the parent figure needs to also be marked as dirty and therefore,  the coordinates have to be translated again, this time relative to the grand-parent.

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
ptziegler: We have to mark both the current figure and the parent as dirty. So we're forced to do two repaints.

azoitl: I feel a bit stupid but all calls I see here are `getParent().repaint(..)`.Also the code inside `AbstractDropShadowBorder`. Where is the grandparent?

org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/shadows/AbstractDropShadowBorder.java
```diff
@@ -220,10 +220,22 @@ public void setSoftness(double softness) {
...
 		shadowRect.expand(shadowSize, shadowSize);
 		graphics.setClip(shadowRect);
 		shadowRect.shrink(shadowSize, shadowSize);
 	}
 
+	@Override
+	public void eraseBackground(IFigure parent, Rectangle area) {
+		int shadowSize = getShadowSize();
+		Rectangle parentArea = area.getCopy();
+		parent.translateToParent(parentArea);
```
azoitl: here you have another translate to parent. Are you sure about this?

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
ptziegler: The `eraseBackground()` method is called with the parent of the figure that is being erased. On this parent, we then call `translateToParent` inside the `AbstractDropShadowBorder` to translate the `parentArea` (i.e. the figure bounds) bounds relative to the grand-parent, so that we can then call `parent.repaint(parentArea)`.

azoitl: but this is then the third `translateToParent` call on the same parent figure and with the same original bounds. I still don't get why this has to be done three times after the first translate to parent it is already translated and as it is not changed why doing it 2 more times.

ptziegler: Ah, no, you're right. I didn't see the third `translateToParent` call and thought you were talking about the original one in the `erase()` method. The second call is wrong also also violates the JavaDoc (the coordinates are expected to be parent-relative not grandparent-relative).

org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/shadows/AbstractDropShadowBorder.java
ptziegler: That one is correct, the one in the `erase()` method is too much and has already been removed again. The idea of doing it here is to avoid any unnecessary overhead for background borders that don't paint outside of the figure bounds.

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
azoitl: As I thought more about it shouldn't we use `parent.getParent()` inside  `AbstractDropShadowBorder` or even do that already in the erase method so that implementers of eraseBackground can harder mess up?

org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/shadows/AbstractDropShadowBorder.java
azoitl: I'm still not fully convinced. Because area is already translated to the grand parent. Why do it here again.

ptziegler: You have every right to be confused. Because now that I look at it again, this doesn't make sense at all. I think I went under the assumption that we're calling `repaint(...)` from the current figure, but the `erase()` method is calling `getParent().repaint(...)`. So calling `translateToParent(...)` more than once is wrong. Moreover, we can then absolutely do it in such a way that `eraseBackground()` only modifies the rectangle, so that `repaint(...)` is only called once from within the `erase()` method.

ptziegler:
@azoitl Apologies for the misunderstanding. You were totally right that the initial PR was very weird. I've adapted everything so that `repaint(...)` is only called once from within the `erase(...)` method and also re-tested this with the Zest snippet and verified that no visual artifacts remain when moving the nodes around.

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
```diff
@@ -380,6 +380,9 @@ public void erase() {
 
 		Rectangle r = new Rectangle(getBounds());
 		getParent().translateToParent(r);
+		if (getBorder() instanceof BackgroundBorder border) {
```
azoitl, line 382: one last stupid comment. if the parent is scaling most probably it would be good to have the background border before the translateToParent or?

azoitl (commented):
@ptziegler thx for your patience with me. I use these kind of PRs as learning oportunity. The drawing and especially updatemanager parts is an area of draw 2d where I understand only a little. Happy that together we got a good solution

org.eclipse.draw2d/src/org/eclipse/draw2d/Figure.java
ptziegler: Technically it doesn't matter because `translateToParent()` doesn't care about scaling. But formally you're correct. Within the `erase()` method, the shadow should be considered a part of the figure bounds, so it only makes sense to call this method before translating the rectangle.

I've updated the PR and moved `eraseBorder()` before `translateToParent()`

azoitl: if the parent is an scale able pane then it would scale. So it is definitely better in this way.

azoitl (approved):

PR #1096, opened 2026-04-23
[PoC] Modernize Zest Graph styling

ptziegler:
- Use RectangleDropShadowBorder for GraphLabel.
- Increase line-width for connections to 2 and enable anti-alias.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@
 
 ## Zest
 
+- Updated  the styling and color theme of the Zest graph and its items.
+  
+  The color constants defines in the `Graph` class have been deprecated for removal in favor of the new 
+  `GraphColorProvider` class, which provides the colors used by the graph items. The highlight color has been adapted 
+  and is now configurable for both the foreground and the background color.
+  
+  Furthermore, the old gradient-based colors have been removed and the blue background color replaced with gray. For the
+  `GraphLabel`, the new `RectangleDropShadowBorder` is used to give the items a more modern look. As a side-effect, it
+  is therefore no longer possible to define the border color and width of this figure, which is why the
+  `IStyleableFigure` has been removed from this class.
+  
+  For the graph connections, the line width has been changed to 2 and anti-alias enabled, to provide a higher contrast
+  against the background.
+
 ## General
 
 
--- a/org.eclipse.zest.core/.settings/.api_filters
+++ b/org.eclipse.zest.core/.settings/.api_filters
@@ -114,4 +114,46 @@
             </message_arguments>
         </filter>
     </resource>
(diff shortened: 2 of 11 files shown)


ptziegler:
Before and after. The same styling also needs to be applied to the container and I have not yet tested how this affects other widgets that use the `GraphLabel`.

image


image

ptziegler:
An open question is the lack of outline. Previously, you were able to set the border width and color. I'm currently using blue selection color as shadow color, but there is no longer any width. Personally, I'm fine with that. But it's something we should keep in mind.

ptziegler:
I've also updated the ExpandGraphLabel. It doesn't make sense to use a drop-shadow here, because it would leak into the client area of the container. So I simply removed the gradient.

image

ptziegler:
I removed the `IStylableFigure` from the `GraphLabel`, as there isn't an outline to style anymore. @vogella @azoitl fyi

vogella:
Looks way better, now a nice gray instead of blue would make a huge difference IMHO.

azoitl:
I can recommend chatgpt for finding good colors. For me it worked best when giving it a screenshot and some constraints (e.g., colors that are given) or some wishes (modern look).

vogella:
I recommend Google Stich or Claude Design (beta) for design work.

ptziegler:
Now with black & gray. I've also deprecated the old color constants. Instead a `GraphColorProvider` is used, extending the `ColorProvider` from Draw2D.

image

ptziegler:
And here how it looks with the SimRel graph:

image

vogella:
Way better, yellow as selection color is debatable

azoitl:
> I recommend Google Stich or Claude Design (beta) for design work.

thx for the tip. will give it a try

ptziegler:
I've now also update the selection color to match the color in e.g. the marquee selection or just the color that is used everywhere else in the IDE.

image

azoitl:
> I recommend Google Stich or Claude Design (beta) for design work.

@vogella I found some time to look into these. These are one step ahead of what I meant when I spoke about design. I was more into developing a design language, affordances, color schemes, accent colors, also analyzing existing designs for consistency, calmness and focusing the users attention. Regarding the last twos here the latest light theme of the Eclipse platform is a major improvement. Congratulations btw. 

But all in all I'm talking about general UX work. I recently posted a full set of IMHO very good videos into the Eclipse IDE matrix channel: https://chat.eclipse.org/#/room/!ogLLxzRyyDnuBlTASe:matrix.eclipse.org/$cI6DDlGPlIsxDbLvtVI3Z63hjHEPsRMqiOoSNkadlqE?via=matrix.eclipse.org&via=matrix.org&via=grin.hu

Issue #1098, opened 2026-04-24
DropShadowBorder may draw outside of client area


ptziegler:
While doing some more testing with #1096, I noticed this problem:

<img width="405" height="424" alt="Image" src="https://github.com/user-attachments/assets/f0c64c5e-3efa-42b1-969f-9ee133a67867" />

...which is a side-effect of this logic here. Calling `setClip(...)` erases the clipping set by the container.

https://github.com/eclipse-gef/gef-classic/blob/68089e9646e867422fd5da3abbc1e21d042b9bd2/org.eclipse.draw2d/src/org/eclipse/draw2d/backgrounds/shadows/AbstractDropShadowBorder.java#L222-L227

azoitl:
I found the approach of explicitly setting the clip not very good. Any tips what I could do here? 

Maybe drawing the shadow outside of the figure bounds was not the best idea.

ptziegler:
Normally one would have to intersect the shadow rectangle with the client area of the parent. But that won't work here because Zest is using a freeform-layer, so the nodes can extend beyond the visible area.

Initially, I thought about using something like this:

```java
private void updateClip(Graphics graphics, Rectangle shadowRect) {
	int shadowSize = getShadowSize();
	CLIP_RECT_CACHE.expand(shadowSize, shadowSize);
	graphics.setClip(CLIP_RECT_CACHE);
	CLIP_RECT_CACHE.shrink(shadowSize, shadowSize);
}
```

But sadly, this only reduces the problem and doesn't actually solve it. So in short, I'm not quite sure yet. You would somehow need to determine the clip of the parent figure, to make sure that you won't exceed it.

ptziegler:
This problem only appears when working with nested `FreeformLayeredPanes`, so I don't think a "general" solution is needed.

What I think will work is a `DropShadowClippingStrategy` in combination with a `DropShadowContainer`, which has the figure containing the drop-shadow border as its only child and whose size is determined by the child figure + shadow size.

azoitl:
Hmh. I'm still thinking if it was a smart idea to draw outside of the figure. Shadows definitely look better and you have more flexibility. But is it worth the hassle?

ptziegler:
I don't think it makes sense to draw the shadow within the figure bounds, as they aren't really part of the figure. And for the majority of use-cases, the current implementation should work just fine.

[...]

PR #1099, opened 2026-04-24
PoC - Move clipping logic from AbstractDropShadowBorder to Figure #1098

ptziegler:
The drop-shadow border should not update overwrite the clip of the Graphics object, as this might cause the shadow to be painted outside of the client area. Instead, the clip of the parent figure needs to be adjusted, to account for the additional area occupied by the shadow.


ptziegler:
@azoitl Something like this might work, but I'm not really happy with it. Alternatively, I could move this logic to a new `DropShadowClippingStrategy`. The only problem is that this would then need to be set for the parent figure, not the figure containing the actual border.

PR #1103, opened 2026-04-27
Set line width to 2 for feedback figures

Phillipus:
This improves visibility of the outline especially on dark theme

Also change line style to LINE_DOT in LayoutEditPolicy as LINE_DASHDOT can look messy on dark theme


ptziegler (approved):
Increasing the line width is fine by me. We do that anyway for e.g. the marquee border.

azoitl (approved):

PR #1107, opened 2026-05-01
[PoC] Separate GEF components depending on the Eclipse 3.x API

ptziegler:
This determines the minimum amount of classes that need to be removed
from `org.eclipse.gef` in order to remove the Eclipse 3.x dependencies.


ptziegler:
This is just a quick demo. Visualization works out-of-the-box but the PaletteViewer will not work without adjustments because the default implementation references the `ScopedPreferenceStore` and the `IMemento`, which are both E3 concepts.

image

ptziegler:
As expected, the naive approach is absolutely brutal and would effectively remove the palette and ruler. The main problem with the palette is the dependency on the `IMemento`, which is an Eclipse 3.x exclusive concept which has been replaced by the `getPersistedState()` method of an `MPart`.

PR #1110, opened 2026-05-03
Fix RoundedRectangleAnchor

Phillipus:
Corner Dimension need to be translated to absolute of getOwner()

Use translated rectangle r instead of getOwner().getBounds()

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
@@ -59,6 +59,18 @@ public class RoundedRectangleAnchor extends ChopboxAnchor {
 	}
 
 	/**
+	 * Return the corner dimensions used to calculate the anchor location in
+	 * {@link #getLocation}. Sub-classes may over-ride this to return a different
+	 * value.
+	 *
+	 * @return The corner dimensions
+	 * @since 3.23
+	 */
+	public Dimension getCornerDimensions() {
+		return getOwner() instanceof RoundedRectangle roundRect ? roundRect.getCornerDimensions() : dimension.getCopy();
+	}
+
+	/**
 	 * Calculates the position with ChopboxAnchor#getLocation() and if the anchor is
 	 * not at the rounded corners, the result is returned. If the anchor point
 	 * should be at a corner, the rectangle for the ellipse is determined and
@@ -69,10 +81,8 @@ public class RoundedRectangleAnchor extends ChopboxAnchor {
 	 */
 	@Override
 	public Point getLocation(final Point ref) {
-		Dimension corner = dimension;
-		if (getOwner() instanceof RoundedRectangle) {
-			corner = ((RoundedRectangle) getOwner()).getCornerDimensions();
-		}
+		final Dimension corner = getCornerDimensions();
+		getOwner().translateToAbsolute(corner);
 		final Point location = super.getLocation(ref);
 		final Rectangle r = Rectangle.SINGLETON;
 		r.setBounds(getOwner().getBounds());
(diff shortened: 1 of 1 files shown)


Phillipus:
Without this fix:

At 100% zoom:

Image

At 400% zoom:

Image

org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
```diff
@@ -69,10 +69,13 @@ public RoundedRectangleAnchor(final Figure figure, final Dimension corners) {
 	 */
 	@Override
 	public Point getLocation(final Point ref) {
-		Dimension corner = dimension;
+		Dimension corner;
 		if (getOwner() instanceof RoundedRectangle) {
```
azoitl:
```suggestion
		if (getOwner() instanceof RoundedRectangle roundRect) {
```

org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
```diff
@@ -69,10 +69,13 @@ public RoundedRectangleAnchor(final Figure figure, final Dimension corners) {
...
 			corner = ((RoundedRectangle) getOwner()).getCornerDimensions();
```
azoitl:
```suggestion
			corner = roundRect.getCornerDimensions();
```

org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
```diff
@@ -69,10 +69,13 @@ public RoundedRectangleAnchor(final Figure figure, final Dimension corners) {
 	 */
 	@Override
 	public Point getLocation(final Point ref) {
-		Dimension corner = dimension;
+		Dimension corner;
```
azoitl: I would suggest to use the ternary operator here with direct assignment of the result. Then corner could be final.

azoitl (changes requested):

org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
Phillipus: Yes, good idea. Done.

azoitl (approved):
Thx for the fix.

Phillipus:
I have another suggestion for this class.

The problem I have in my use of it in the constructor `RoundedRectangleAnchor(final Figure figure, final Dimension corners)` is that `corners` is statically set. In one of my figures `corners` is dynamic and needs to be re-evaluated in `getLocation`.

If the following change was made then sub-classes can over-ride `getCornerDimensions()` with a value:

```java
// Add this method
public Dimension getCornerDimensions() {
	return dimension;
}

// Change this line
final Dimension corner = getOwner() instanceof RoundedRectangle roundRect ? roundRect.getCornerDimensions()
		: getCornerDimensions().getCopy();
```

And then can be used like this:

```java
RoundedRectangleAnchor anchor = new RoundedRectangleAnchor(figure, dynamicVal) {
    @Override
    public Dimension getCornerDimensions() {
        return dynamicVal;
    }
};
```

azoitl:
@Phillipus I like this proposal. Not only does it help you it makes the class in total cleaner.

Phillipus:
>  I like this proposal. Not only does it help you it makes the class in total cleaner.

Great!

What's the best practice here for a comment on the new method and what version is it for `@since` in Draw2d?

azoitl:
> > I like this proposal. Not only does it help you it makes the class in total cleaner.
> 
> Great!
> 
> What's the best practice here for a comment on the new method and what version is it for `@since` in Draw2d?

I always try to provide a reasonable Javadoc description of a new method (not that I'm good in that). for the `@since` version you need to take the current version of the plugin's manifest the class is contained in. In your case this would be `3.23`. If you would be the first to enhance a plugin then it may mean that you also have to update the plugins version number. The PDE API tools would tell you that.

Phillipus:
> What's the best practice here for a comment on the new method and what version is it for `@since` in Draw2d?

I made a stab at it in latest commit.

org.eclipse.draw2d/src/org/eclipse/draw2d/RoundedRectangleAnchor.java
```diff
@@ -69,10 +81,9 @@ public RoundedRectangleAnchor(final Figure figure, final Dimension corners) {
 	 */
 	@Override
 	public Point getLocation(final Point ref) {
-		Dimension corner = dimension;
-		if (getOwner() instanceof RoundedRectangle) {
-			corner = ((RoundedRectangle) getOwner()).getCornerDimensions();
-		}
+		final Dimension corner = getOwner() instanceof RoundedRectangle roundRect ? roundRect.getCornerDimensions()
```
azoitl: should we move the whole statement into the getCornerDimension method?

Phillipus: Do you mean:

```java
public Dimension getCornerDimensions() {
	return getOwner() instanceof RoundedRectangle roundRect ? roundRect.getCornerDimensions() : dimension.getCopy();
}

public Point getLocation(final Point ref) {
    final Dimension corner = getCornerDimensions();
```

Phillipus: Done.

azoitl: Yes, as this would be the corner dimension to be used. But I'm not sure.

Phillipus: It has the advantage of always returning a copy of the `Dimension` as `roundRect.getCornerDimensions()` already returns a copy.

azoitl (approved):
I like the new method.

PR #1113, opened 2026-05-08
Fix ConnectionEndpointLocator co-ordinates and add useOwnerOffset method

Phillipus:
- Connection owner bounds should be translated to correct co-ordinate system

- Add a new constructor with an option to not use an offset when calculating the location in relation to the owner

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ConnectionEndpointLocator.java
@@ -195,8 +195,10 @@ public class ConnectionEndpointLocator implements Locator {
 		IFigure connOwner = getConnectionOwner();
 
 		int quadrant;
-		if (connOwner != null) {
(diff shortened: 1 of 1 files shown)


ptziegler:
Hm... as you said, you're probably the only one who's using this new constructor. So I wonder whether that's a little bit too much.

What about instead adding a protected `useOwnerOffset()` method that you can subclass?

Phillipus:
> What about instead adding a protected useOwnerOffset() method that you can subclass?

Excellent idea! I've changed it.

Thanks for looking at this.

ptziegler (approved):
Thanks!

azoitl (approved):
Thx for finding this issue and the fix. Sorry for currently not being as responsive.

Phillipus:
> Thx for finding this issue and the fix. Sorry for currently not being as responsive.

Hey, not need to apologise. The thanks goes to you and @ptziegler for the amazing work you do on this project. Without you guys this project would have not be as great as it is. 👏  👏  👏  👏  👏  👏

PR #1117, opened 2026-05-12
Added Helper functions to Vector and Point

azoitl:
Added a normalize function to Vector and allow Vector as translation input for Point and PrecisionPoint.

### DIFF
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Point.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Point.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -350,6 +350,18 @@ public class Point implements Cloneable, java.io.Serializable, Translatable {
 	}
 
 	/**
+	 * Creates a new Point which is translated by the values of the provided
+	 * {@link Vector}.
+	 *
+	 * @param vec {@link Vector} which provides the translation amounts.
+	 * @return A new Point
+	 * @since 3.23
+	 */
+	public Point getTranslated(Vector vec) {
+		return getCopy().translate(vec);
+	}
+
+	/**
 	 * Creates a new Point with the transposed values of this Point. Can be useful
 	 * in orientation change calculations.
 	 *
@@ -547,6 +559,18 @@ public class Point implements Cloneable, java.io.Serializable, Translatable {
 	 */
 	public Point translate(Point p) {
 		return translate(p.x(), p.y());
+	}
+
+	/**
+	 * Shifts the location of this Point by the input {@link Vector} along each of
+	 * the axes, and returns this for convenience.
+	 *
+	 * @param vec {@link Vector} to which the origin is being shifted.
+	 * @return <code>this</code> for convenience
+	 * @since 3.23
+	 */
+	public Point translate(Vector vec) {
+		return translate(vec.x, vec.y);
 	}
 
 	/**
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Vector.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Vector.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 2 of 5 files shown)


org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Vector.java
```diff
@@ -21,6 +21,8 @@
  */
 public class Vector {
 
+	private static final double LENGHT_EPSILON = 0.001;
```
Phillipus: Is that a typo for LENGTH_EPSILON?

azoitl: Yes it is. Should stop programming in the middle of the night. Thx for pointing that out.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Point.java
```diff
@@ -349,6 +349,18 @@ public Point getTranslated(Point p) {
...
+	 * {@link Vector}.
+	 *
+	 * @param vec {@link Vector} which provides the translation amounts.
+	 * @return A new Point
+	 * @since 3.23
+	 */
+	public Point getTranslated(Vector vec) {
+		return getCopy().translate(vec);
+	}
+
```
ptziegler, line 363: The `getTranslated(...)` and `translate(...)` method also exist for the `Rectangle` class. Would it make sense to add those methods there as well, to make sure that everything is consistent?

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Vector.java
```diff
@@ -205,6 +207,17 @@ public Vector getMultiplied(double factor) {
...
+	 * Creates a new Vector that is normalized with its length, i.e., a vector with
+	 * length 1.
+	 *
+	 * @return the normalized Vector
+	 * @since 3.23
+	 */
+	public Vector getNormalized() {
+		return getDivided(Math.max(getLength(), LENGTH_EPSILON));
+	}
+
```
ptziegler, line 227: As far as I remember, normalizing a zero-width vector is undefined behavior and I don't think it makes sense to just ignore it, especially when such a vector violates the JavaDoc contract (i.e. the method would still return a zero-width vector). I would rather check whether the length is close to zero and then throw e.g.  an `UnsupportedOperationException`

azoitl: You are right that my solution is a good one. However I'm not sure if an exception is the right thing. I asked my 4 favorite LLMs and all suggested to not throw and exception. But I'm not sure to trust them and what to do.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Point.java
azoitl: good idea will do that.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Vector.java
ptziegler: I had a look at Apache Commons Math and they throw an `MathArithmeticException`. I also looked at some old-school libraries like javax.vecmath, which simply fails with a "division by zero" error. To be honest, I don't have a strong opinion either way. But then the result for a zero-length vector needs to be properly documented.

If you want to avoid an exception but still be mathematically "correct", you could also return an `Optional`, which is empty for a zero-length vector. But it depends on your use case whether that's too verbose.

azoitl: I don't know if I would like to be mathematically correct. I'm thinking what is the main use case of `Vector`? For me to perform some calculations that ends up in some drawings. If I have to check in these calculations everywhere for optionals or exceptions the code will get rather hard to read. Therefore I thought returning some default vector even if not mathematically correct it could make it easier to use for our users and the worst case happening is distorted diagrams. These would be easier to see and maybe even easier to debug. 

At least this is the experience I had when working with my Google Summer of Code student on [this PR](https://github.com/eclipse-4diac/4diac-ide/pull/2421), which was the starting point for this PR here.

ptziegler: As said, I don't have a strong opinion on this. If you want to keep it like that, that's fine. I only suggest that you then explicitly document this edge case in the JavaDoc. :)

azoitl: Made the implementation now a bit nicer and extended the java doc.

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Point.java
azoitl: done

org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Vector.java
ptziegler: Thanks!

ptziegler (approved):

PR #1119, opened 2026-05-15
Fix incorrect palette minimum width

Phillipus:
If on Windows at say 200% display scaling the minumim palette width is twice as big compared with 100% display scaling or on Mac and Linux

Remove scaling factor from minWidth computation

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java
@@ -623,7 +623,7 @@ public class FlyoutPaletteComposite extends Composite {
 					restorePaletteState(pViewer, capturedPaletteState);
 				}
 				capturedPaletteState = null;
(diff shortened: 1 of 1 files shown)


Phillipus:
Before this patch the minimum width on Windows at 200% monitor display:

thick

After (same as Windows 100%, Mac and Linux):

thin

azoitl (approved):
LGTM. @ptziegler WDYT?

ptziegler (approved):
I'm fairly certain I tested this with the original change and the call to `computeSize` would return the size at 100% zoom. But this no longer seems to be the case, meaning that this scaling is definitely wrong with the latest SWT.

ptziegler:
Does this also fix the scrollbar issue?

Phillipus:
> Does this also fix the scrollbar issue?

No, that's different.

Phillipus:
> I'm fairly certain I tested this with the original change and the call to `computeSize` would return the size at 100% zoom. But this no longer seems to be the case, meaning that this scaling is definitely wrong with the latest SWT.

I found a similar inconsistency with `computeSize` [here](https://github.com/eclipse-gef/gef-classic/discussions/1108#discussioncomment-16814713).

PR #1121, opened 2026-05-22
Added Creator Infrastructure for Marquee Selection Feedback Figure

azoitl:
This infrastructure allows clients to provide their own marquee selection feedback figure without the need to subclass MarqueeSelectionTool.

Addresses: https://github.com/eclipse-gef/gef-classic/discussions/1078

### DIFF
--- a/org.eclipse.gef/src/org/eclipse/gef/tools/MarqueeSelectionTool.java
+++ b/org.eclipse.gef/src/org/eclipse/gef/tools/MarqueeSelectionTool.java
@@ -15,6 +15,8 @@ package org.eclipse.gef.tools;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.Objects;
+import java.util.function.Supplier;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.KeyEvent;
@@ -78,6 +80,7 @@ import org.eclipse.gef.util.EditPartUtilities;
 public class MarqueeSelectionTool extends AbstractTool {
 
 	private static class MarqueeRectangleFigure extends Figure {
+		public static final Supplier<IFigure> DEFAULT_MARQUEE_RECTANGLE_FIGURE_CREATOR = MarqueeRectangleFigure::new;
 
 		private static final int DELAY = 110; // animation delay in millisecond
 		private int offset = 0;
@@ -215,6 +218,10 @@ public class MarqueeSelectionTool extends AbstractTool {
 	private Collection<? extends GraphicalEditPart> selectedEditParts;
 
 	private Request targetRequest;
+
+	private static Supplier<? extends IFigure> globalMarqueeRectangleFigureCreator = MarqueeRectangleFigure.DEFAULT_MARQUEE_RECTANGLE_FIGURE_CREATOR;
+
+	private Supplier<? extends IFigure> marqueeRectangleFigureCreator = null;
 
 	/**
 	 * Creates a new MarqueeSelectionTool of default type
@@ -424,9 +431,9 @@ public class MarqueeSelectionTool extends AbstractTool {
 	 *
 	 * @since 3.13
 	 */
-	@SuppressWarnings("static-method")
 	protected IFigure createMarqueeRectangleFigure() {
-		return new MarqueeRectangleFigure();
+		return (marqueeRectangleFigureCreator != null) ? marqueeRectangleFigureCreator.get()
+				: globalMarqueeRectangleFigureCreator.get();
 	}
 
 	private Request getTargetRequest() {
@@ -743,6 +750,33 @@ public class MarqueeSelectionTool extends AbstractTool {
 			throw new IllegalArgumentException("Invalid marquee behaviour specified."); //$NON-NLS-1$
 		}
 		marqueeBehavior = type;
+	}
+
+	/**
+	 * Set a specific MarqueeRectangleFigureCreator to be used per default for all
+	 * MarqueeSelectionTools.
+	 *
+	 * @param marqueeRectangleFigureCreator the creator object, must not be null
+	 * @since 3.26
+	 */
+	public static void setGlobalMarqueeRectangleFigureCreator(
+			Supplier<? extends IFigure> marqueeRectangleFigureCreator) {
+		Objects.nonNull(marqueeRectangleFigureCreator);
+		globalMarqueeRectangleFigureCreator = marqueeRectangleFigureCreator;
+	}
+
+	/**
+	 * Allow to set a dedicated creator for a marquee rectangle figure to be used as
+	 * feedback figure for marquee selections of this MarqueeSelectionTool instance.
+	 *
+	 * @param marqueeRectangleFigureCreator the creator object, null if the global
+	 *                                      marqueeRectangleFigureCreator should be
+	 *                                      used
+	 * @since 3.26
+	 */
+	public void setMarqueeRectangleFigureCreator(Supplier<? extends IFigure> marqueeRectangleFigureCreator) {
+		Objects.nonNull(marqueeRectangleFigureCreator);
+		this.marqueeRectangleFigureCreator = marqueeRectangleFigureCreator;
 	}
 
 	private void setSelectionMode(int mode) {


ptziegler:
The `MarqueeSelectionTool` is generally deeply embedded into the GEF viewer. How exactly would you get the instance that is used for selection? Would you have to subclass the root edit-part, override the `getDragTracker()` method and then modify the rectangle? Or is there an easier way?

Phillipus:
Thanks for looking at this.

For Archi I can hook into `MarqueeSelectionTool` when creating the Palette entries:

```java
MarqueeToolEntry toolEntry = new MarqueeToolEntry("Selection") {
    @Override
    public Tool createTool() {
        MarqueeSelectionTool tool = (MarqueeSelectionTool)super.createTool();
        tool.setMarqueeRectangleFigureCreator(() -> {
            // TODO create figure
            return new Figure();
        });
        return tool;
    }
};

toolEntry.setToolProperty(MarqueeSelectionTool.PROPERTY_MARQUEE_BEHAVIOR, 
        Integer.valueOf(MarqueeSelectionTool.BEHAVIOR_NODES_CONTAINED_AND_RELATED_CONNECTIONS));
```

I'd also need to access `MarqueeDragTracker`:

```java
viewer.setRootEditPart(new ScalableFreeformRootEditPart() {
    @Override
    public DragTracker getDragTracker(Request req) {
        MarqueeDragTracker tracker = (MarqueeDragTracker)super.getDragTracker(req);
        tracker.setMarqueeRectangleFigureCreator(() -> {
            // TODO create figure
            return new Figure();
        });
        return tracker;
    }
});
```

For my use case I would like to use `MarqueeRectangleFigure` but change a few things like line width, style and colour.

azoitl:
@ptziegler I know that `MarqueeSelectionTool` is deep embedded into GEF Viewers. I was searching for a way that allows me to customize it without the need of subclassing like I did in my attempt with the factory method I introduced some years back. Now thinking again on my approach I'm not so sure if this is the best approach. 

Question to both @ptziegler and @Phillipus does it make sense that the marquee selection feedback figure can be configured per tool instance? Or wouldn't it make more sense to have the creator function as class variable that you can globally set?

When we have the design fixed I' planning to provide a view feedback figures as we have discussed  #1078. I guess I would move them to a gef.figures package with the option to override and adjust.

Phillipus:
> Question to both @ptziegler and @Phillipus does it make sense that the marquee selection feedback figure can be configured per tool instance? Or wouldn't it make more sense to have the creator function as class variable that you can globally set?

I'm not sure. I need to think about it.

BTW - I've created my own custom `MarqueeRectangleFigure`. I noticed that the existing one doesn't need to use a set of int[] `points` but can use `graphics.setLineDashOffset()` and draw a rectangle. Here's mine that draws a black outer rectangle and a white inner rectangle and so can be seen on all background colours:

```java
public class CustomMarqueeRectangleFigure extends Figure {
    private static final int DELAY = 110;
    private int offset = 0;
    private boolean schedulePaint = true;
    private int lineWidth = 1;
    
    @Override
    protected void paintFigure(Graphics graphics) {
        graphics.translate(getLocation());

        graphics.setLineDash(new int[]{3, 3});
        graphics.setLineWidth(lineWidth);
        graphics.setLineDashOffset(-offset);
        
        Rectangle rect = new Rectangle(0, 0, bounds.width(), bounds.height());
        
        // Outer line white
        graphics.setForegroundColor(ColorConstants.black);
        rect.shrink(lineWidth, lineWidth);
        graphics.drawRectangle(rect);

        // Inner line white
        graphics.setForegroundColor(ColorConstants.white);
        rect.shrink(lineWidth, lineWidth);
        graphics.drawRectangle(rect);
            
        graphics.translate(getLocation().getNegated());
        
        if(schedulePaint) {
            Display.getCurrent().timerExec(DELAY, () -> {
                offset++;
                if(offset > 5) {
                    offset = 0;
                }
                
                schedulePaint = true;
                repaint();
            });
        }

        schedulePaint = false;
    }
}
```

azoitl:
Now that I thought of it I could even do a global selection figure which can be overridden per instance. Maybe a bit of an overkill but would give us the most flexibility.

Phillipus:
> Now that I thought of it I could even do a global selection figure which can be overridden per instance. Maybe a bit of an overkill but would give us the most flexibility.

I'd be interested to see that. :-)

azoitl:
> > Now that I thought of it I could even do a global selection figure which can be overridden per instance. Maybe a bit of an overkill but would give us the most flexibility.
> 
> I'd be interested to see that. :-)

I updated the PR with my idea. While doing that I also had the crazy idea to allow the current viewer to provide the figure. Then you could easily configure that per viewer? But not sure if I'm now going over the top.

Phillipus:
Well, the global static setter would certainly make things simpler for me:

```java

protected PaletteStack createMarqueeSelectionStack() {
    MarqueeSelectionTool.setGlobalMarqueeRectangleFigureCreator(() -> {
        return new MarqueeSelectionFigure();
    });
    
    PaletteStack stack = new PaletteStack(Messages.AbstractPaletteRoot_0, Messages.AbstractPaletteRoot_1, null);
    stack.add(createMarqueeToolEntry(Messages.AbstractPaletteRoot_2, MarqueeSelectionTool.BEHAVIOR_NODES_CONTAINED_AND_RELATED_CONNECTIONS));
    stack.add(createMarqueeToolEntry(Messages.AbstractPaletteRoot_3, MarqueeSelectionTool.BEHAVIOR_NODES_TOUCHED_AND_RELATED_CONNECTIONS));
    stack.add(createMarqueeToolEntry(Messages.AbstractPaletteRoot_4, MarqueeSelectionTool.BEHAVIOR_CONNECTIONS_CONTAINED));
    stack.add(createMarqueeToolEntry(Messages.AbstractPaletteRoot_5, MarqueeSelectionTool.BEHAVIOR_CONNECTIONS_TOUCHED));
    stack.add(createMarqueeToolEntry(Messages.AbstractPaletteRoot_6, MarqueeSelectionTool.BEHAVIOR_NODES_CONTAINED));
    stack.add(createMarqueeToolEntry(Messages.AbstractPaletteRoot_7, MarqueeSelectionTool.BEHAVIOR_NODES_TOUCHED));
    return stack;
}

protected MarqueeToolEntry createMarqueeToolEntry(String label, int behaviour) {
    MarqueeToolEntry entry = new MarqueeToolEntry(label);
    entry.setToolProperty(MarqueeSelectionTool.PROPERTY_MARQUEE_BEHAVIOR, behaviour);
    return entry;
}
```

> While doing that I also had the crazy idea to allow the current viewer to provide the figure. 

Again, I'd have to see that to judge it.

Thanks for all these great ideas!

azoitl:
> >    While doing that I also had the crazy idea to allow the current viewer to provide the figure.
>
>  Again, I'd have to see that to judge it.

This would be a bit larger but it the `createMarqueeRectangleFigure()`would then most probably look something like this:

```
protected IFigure createMarqueeRectangleFigure() {
     if(getCurrentViewer() != null && getCurrentViewer.getMarqueeRectangleFigureCreator().isPresent()) {
          return getCurrentViewer.getMarqueeRectangleFigure().get().get();  // the double get for handling optional and the function
     } 
     return (marqueeRectangleFigureCreator != null) ? marqueeRectangleFigureCreator.get()
				: globalMarqueeRectangleFigureCreator.get();
}
```

The viewer would then have a getter and a setter for MarqueeRectangleFigureCreator. But I'm not sure if the viewer is right place for this. It would have the advantage that you could easily set the marqueerectagle figure per viewer. So if you have viewers with content in different colors you could adjust for that. But maybe the interface presented in this PR is already sufficient.

ptziegler:
Personally, I don't like those "global" factories at all. Because the moment you integrate multiple GEF-based editors in your application, you always have the risk that two or more projects set their own factory and suddenly, everything breaks. Having a factory per (type of) viewer sounds better.

Would something similar to the `IToolTipHelperFactory` from f93af24cd1b4abc3b7085d25fef3fbde4f060e6b work here as well? You define an `IMarqueeRectangleFactory` interface, where implementations can be provided as an OSGi service and which you can then access from the `InternalGEFPlugin`.

Phillipus:
> Would something similar to the IToolTipHelperFactory from https://github.com/eclipse-gef/gef-classic/commit/f93af24cd1b4abc3b7085d25fef3fbde4f060e6b work here as well?

I tried to use that but my custom `IToolTipHelperFactory` is never picked up. I registered it using `context.registerService(IToolTipHelperFactory.class.getName(), new MyToolTipHelperFactory(), null);`

ptziegler:
> I tried to use that but my custom IToolTipHelperFactory is never picked up. I registered it using context.registerService(IToolTipHelperFactory.class.getName(), new MyToolTipHelperFactory(), null);

The services are loaded when the GEF plugin is started. My guess is that this new service is registered afterwards.

https://github.com/eclipse-gef/gef-classic/blob/d8f3d75ccec63bb5c0ff2aa113a40d1e182c2eeb/org.eclipse.gef/src/org/eclipse/gef/internal/InternalGEFPlugin.java#L60-L70

Perhaps that should be changed so that the list of services are never stored in a local variable... That would also solve the potential problem when one of the contributing plugins is uninstalled at runtime.

Nonetheless, the indented way is via the `@Component` annotation + the `PaletteToolTipHelperFactory.xml` in the OSGI-INF folder. But I can also try to have a look if this approach can be enhanced.

https://github.com/eclipse-gef/gef-classic/blob/d8f3d75ccec63bb5c0ff2aa113a40d1e182c2eeb/org.eclipse.gef.examples/src/org/eclipse/gef/examples/palette/PaletteSnippet2.java#L128-L137

Phillipus:
> Nonetheless, the indented way is via the @Component annotation + the PaletteToolTipHelperFactory.xml in the OSGI-INF folder. 

Thanks. I didn't look closely enough at the example snippet. It works now.

azoitl:
@ptziegler I like this proposal. Haven't yet found time to look into it. Will do that hopefully soon.

PR #1122, opened 2026-05-24
Use PrecisionRectangle when calculating Freeform bounds #1116

### LINKED ISSUE #1116: [Windows] SnapToGeometry helper unnecessary scrollbars appear
- On Windows with display scales of > 100% using a `SnapToGeometry` helper guide and with the Palette open
- When dragging a figure in a diagram and the blue helper guide lines appear, horizontal and vertical scroll bars also appear in the viewport
- If the Palette is closed then the scroll bars do not appear.

I use it like this:

```java
getGraphicalViewer().setProperty(SnapToGeometry.PROPERTY_SNAP_ENABLED, true);
```

With a `SnapToHelper` and `SnapToGeometry` class.

Here's a video. Note that as the figure is moved and the snap guide lines appear scroll bars also appear. When I close the Palette the scroll bars do not appear when the figure is moved.

https://github.com/user-attachments/assets/1a7a1d70-6c9a-4be0-905e-15afaafb983c

If `draw2d.enableAutoscale` is false the scroll bars do not appear so it must be something to do with auto scaling on Windows.

This is not easy to create a snippet for. I'm happy to fix and provide a PR but at this point I have no idea what could cause this. And I don't know why this happens only when the palette is present.

ptziegler:
When the `AutoscaleFreeformViewport` is used, the bounds for the `FreeformFigure`s have to be set using a `PrecisionRectangle`, to account for rounding errors that are introduced when scaling the figure bounds by the monitor zoom.

Closes https://github.com/eclipse-gef/gef-classic/issues/1116

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AutoscaleFreeformViewportTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AutoscaleFreeformViewportTest.java
@@ -1,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.draw2d.test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.eclipse.draw2d.AutoscaleFreeformViewport;
+import org.eclipse.draw2d.FreeformLayer;
+import org.eclipse.draw2d.geometry.Rectangle;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class AutoscaleFreeformViewportTest {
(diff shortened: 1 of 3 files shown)


ptziegler:
Given that we're little late right now, would it make more sense to postpone it for the M1 or would you want to merge it for the current release?

Phillipus:
Thanks for looking at this and finding the problem. :-)

> Given that we're little late right now, would it make more sense to postpone it for the M1 or would you want to merge it for the current release?

I don't want you to merge it if there's a risk of regression. I can always build against M1 or nightly.

ptziegler:
> I don't want you to merge it if there's a risk of regression. I can always build against M1 or nightly.

At least for me it's always about risk vs reward. The problem should've existed since last year, so it's not a regression that was introduced with this release. The problem itself is also not something that I would consider as "critical".

Right now I'm more concerned about why the example editor tests are failing on MacOS, even though this component shouldn't be affected at all... Strange.

Phillipus:
Could this be merged now? I've been using the PR in my fork and it fixes the issue with no side-effects seen.

ptziegler:
Of course. A lot of stuff went on over the past weeks, so this issue slipped my mind. In case anyone is curious, the reason this rounding error happens is because of the implementation of the `Rectangle` class:

https://github.com/eclipse-gef/gef-classic/blob/3a98304627d4ed82f5ecf014bd86b9fbaa4dce65/org.eclipse.draw2d/src/org/eclipse/draw2d/geometry/Rectangle.java#L947-L955

When the client area is scaled by the monitor zoom, the bounds of the rectangle are rounded _up_. So it might happen that the "scaled" client area is larger than the physical client area on the monitor, which then causes the scrollbar to appear.

Phillipus:
Many thanks!

PR #1124, opened 2026-05-25
Build and test on MacOS-14 agent

ptziegler:


ptziegler:
Can be closed, as it's now clear that the problem is caused by a buggy runner. The problem was already reported to the upstream project and I think it's preferable to just sit this out, rather than being stuck on an older version.

PR #1131, opened 2026-06-11
Remove old target platform and baselines

ptziegler:
GEF only supports the Eclipse 2024-06 and newer. There is no benefit in keeping older baselines.


azoitl (approved):
this is a very good idea.

PR #1133, opened 2026-06-22
Make Zest graph styling configurable

ptziegler:
Consumers can already customize the styling of a Zest `Graph` by implementing a custom `IEntityStyleProvider`. However, this is only possible in combination with a JFace viewer and not pure SWT-based graph.

Additionally, this provider implements the `IDisposable` interface, which means that it also can't be used in a pure E4 application. With this change, the corresponding methods in the `IEntityStyleProvider` have been marked for removal and now provide a default implementation.

Instead of this provider, consumers can now use the `GraphColorProvider` to specify the colors used when creating the graph items and which can be set by calling `setColorProvider()` on the `Graph` widget.

The public `Color` fields have been marked for removal in favor of this new color provider.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/1090

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## Zest
 
+- Node styling is node done via the `GraphColorProvider`, as opposed to the `IEntityStyleProvider`. This allows
+  consumers to easily style graph items without relying on JFace classes. The corresponding methods in the
+  `IEntityStyleProvider` have been deprecated for removal.
+
 ## General
 
 
--- a/org.eclipse.zest.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.zest.core/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.zest.core;singleton:=true
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
-Bundle-Version: 1.18.100.qualifier
+Bundle-Version: 1.19.0.qualifier
 Require-Bundle: org.eclipse.zest.layouts,
  org.eclipse.ui;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.draw2d;bundle-version="[3.21.0,4.0.0)";visibility:=reexport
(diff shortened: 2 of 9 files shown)


azoitl (commented):
I see how this is cleaner and more versatile. However I'm missing `IEntityConnectionStyleProvider`and `IEntityConnectionStyleProvider2` and also the `GraphItemStyler`. Wouldn't the later override anything from the new color provider. Or what I'm not understanding I now need to implement both a color provider and a styler if I want to have my own graph style or?

ptziegler:
> However I'm missing IEntityConnectionStyleProvider and IEntityConnectionStyleProvider2 and also the GraphItemStyler.

Yes, I missed the `IEntityConnectionStyleProvider` and also the `IConnectionStyleProvider`. Which means there are three ways to customize the items. Yikes... I don't think there is anything to change in the `GraphItemStyler`, because all it does it to delegate the colors from the provider to the graph. Though eventually, it of course should refresh the colors based on the color provider.

> Wouldn't the later override anything from the new color provider.

Yes, the color provider is called to determine the initial color of the items. Afterwards (also when the items are created, but also when the items are refreshed), the colors are overwritten by the styler.

> Or what I'm not understanding I now need to implement both a color provider and a styler if I want to have my own graph style or?

If you want to define your own colors, you should only need to implement the `GraphColorProvider`. But that you can do for an SWT-based and a JFace-based graph. If you want to have additional functionality (e.g. custom tooltips). But this is only possible for a JFace-based graph.

Phillipus:
I use a `ISelfStyleProvider` in a `IBaseLabelProvider`. Am I affected by this?

ptziegler:
> I use a ISelfStyleProvider in a IBaseLabelProvider. Am I affected by this?

No. The `GraphColorProvider` is always called first. So any changes you do with the other providers will take priority.

ptziegler:
Alternatively, I can also just change the constants. But then we're back to square one where you can't easily change them.

ptziegler:
Maybe as a third option, I could re-use the `ISelfStyleProvider` for the SWT `Graph` widget, instead of creating a new `GraphColorProvider` class.

ptziegler:
Closing in favor of #1142

PR #1136, opened 2026-06-22
ConnectionEndpointPolicy strokes outline to late

### LINKED ISSUE #1018: ConnectionEndpointEditPolicy doesn't scale with editor zoom
When at 100%:

Image

When at 400%:

Image

In one case, the rectangle encloses the connection, in the other it doesn't.

azoitl:
Fixes #1018


ptziegler (approved):
That's cleaner than what I had in mind. I simply would've multiplied the offset by the editor zoom.

azoitl:
> That's cleaner than what I had in mind. I simply would've multiplied the offset by the editor zoom.

This is one thing that I learned in the discussions with @Phillipus: if we just use editor zoom we always are in trouble with native scaling.

PR #1139, opened 2026-06-23
Fix clipping in Outline if at > 100% zoom #1135

### LINKED ISSUE #1135: Outline is clipped if at >100% zoom
Image

ptziegler:
The source size of the thumbnail image needs to be in absolute coordinates, to make sure that the number of tiles encloses the entire client area of the viewer.

Closes https://github.com/eclipse-gef/gef-classic/issues/1135


ptziegler:
I've highlighted the tiles which are currently being painted. This is the result at 150% zoom, where 1/3 is simply not painted.

image

ptziegler:
For comparison, this is what it looks like with this fix:
image

Phillipus:
Bingo! Thanks for this. I spent some time today to find the cause and failed. This fixes it.

Note to self - if in doubt, remember the mantra "Translate To Absolute" 😉

azoitl (approved):

PR #1142, opened 2026-06-25
Separate graph styling from label provider  #1141

### LINKED ISSUE #1141: Harmonize Zest style providers
# Problem

The way the Zest style providers are implemented impose some very harsh restriction when it comes to binary-compatible changes. This was already noticeable with the initial Zest 2.0 merge, where `IConnectionStyleProvider2` and `IEntityConnectionStyleProvider2` were added to provide additional functionality. The reason for this is simple: You can't add new methods to an interface.

Generally, this issue is avoid by adding mentioned `...2` and `...3` interfaces or by adding methods with a default implementation. Former doesn't scale at all, as you can see with e.g. JFace, where a `IDocumentExtension`, `IDocumentExtension2`, `IDocumentExtension3` and `IDocumentExtension4` exists. Latter is also just a workaround because all new methods are optional and you have this weird distinction between these new methods and the mandatory ones.

## Idea:

I would try to use a similar approach to the Draw2D figures, where the provider has an interface marked as `@NoImplement` and `@NoExtend`, where all methods are defined and where new methods may be added in the future. For each interface, there is then an (abstract) base-class available, that may be subclassed by clients and which contains a default implementation for each declared method. Client can and should extend this class. If at some point new methods are added to the interface, the base class can simply be extended and thus avoiding a break in binary compatibility.

The problem: This approach doesn't work well with the way those providers are currently used. Generally, you have something like this:

```java
class CustomLabelProvider extends LabelProvider implements ISelfStyleProvider {
    @Override
    public String getText(Object element) {
        ...
    }

    @Override
    public String getImage(Object element) {
        ...
    }

    @Override
    public void selfStyleConnection(Object element GraphConnection connection) {
        ...
    }

    @Override
    public void selfStyleNode(Object element, GraphNode node) {
        ...
    }
}
```
Because Java doesn't support multi-inheritance, the approach with an abstract base class doesn't work. As a solution to this, I propose to separate the label provider from the style provider. JFace already has a similar concept with the `ILabelDecorator`, which I would simply repurpose.

In short, I would the introduce the following classes:

```java
public interface IGraphLabelDecorator extends ILabelDecorator {
	void decorateConnection(GraphConnection connection);

	void decorateNode(GraphNode node);
}

public class DecoratingGraphLabelProvider extends DecoratingLabelProvider {
	public DecoratingGraphLabelProvider(ILabelProvider provider, IGraphLabelDecorator decorator) {
		super(provider, decorator);
	}

	public IGraphDecorator getGraphDecorator() {
		return (IGraphDecorator) getLabelDecorator();
	}
}
```

In the `GraphStyler`, I would check whether the label provider is of type `DecoratingGraphLabelProvider`, which then calls the decorate methods of the `IGraphLabelDecorator`. For the existing providers, I would create matching decorators. For example an abstract `ConnectionStyleDecorator`, which contains all methods defined in the `IConnectionStyleProvider`, together with a default implementation of the decorate methods.

The existing decorators I would then mark for removal. Because the new decorators don't implement `IDisposable`, it also has the added benefit of solving [Bug 261662](https://bugs.eclipse.org/bugs/show_bug.cgi?id=261662) - `interface IEntityStyleProvider pulls in dependency on org.eclipse.ui.workbench` and eventually allows consumers to use the `GraphViewer` in a pure E4 environment.

As opposed to the example above, you then have something along the lines of:
```java
class CustomLabelProvider extends LabelProvider {
    @Override
    public String getText(Object element) {
        ...
    }

    @Override
    public String getImage(Object element) {
        ...
    }
}

class CustomStyleDecorator extends IGraphDecorator {
    @Override
    public void decorateConnection(GraphConnection connection) {
        ...
    }

    @Override
    public void decorateNode(GraphNode node) {
        ...
    }
}

GraphViewer viewer = ...;
viewer.setLabelProvider(new DecoratedGraphLabelProvider(new CustomLabelProvider(), new CustomStyleDecorator()));

```

This decorator can later also be used as an alternative to #1133.

ptziegler:
This introduces a new `IGraphDecorator` and `IGraphLabelProvider` interface to deprecate the `ISelfStyleProvider` interface, together with the `IConnectionStyleProvider`, `IEntityConnectionStyleProvider` and `IEntityStyleProvider` interface. Default decorator implementations have been added for those classes to simplify the migration.

The `GraphLabelDecorator` can be used directly as label provider for the `GraphViewer`, if labels and icons are not used. Otherwise this decorator needs to be used in combination with an `ILabelProvider` in a `DecoratingLabelProvider`.

Separating both components allows us to extend the providers if needed, without breaking binary compatibility. Something that is currently not possible.

Closes https://github.com/eclipse-gef/gef-classic/issues/1141


ptziegler:
@azoitl What do you think of this approach. The issue with the provider has bothered me for quite a while and I think this is the "cleanest" way to sort things out. If you're fine with that, I'd continue updating the remaining providers and to update the documentation.

@Phillipus Since you mentioned using the `ISelfStyleProvider`.

ptziegler:
I've also migrated/deprecated the remaining style providers.

Phillipus:
Thanks for your work on this.

I applied it to my use of a label provider that implements `ISelfStyleProvider` and I see no issues doing it this new way.

Basically I supplied an empty `LabelProvider` and a custom `GraphLabelDecorator`:

```java
public MyLabelProvider() {
    super(new LabelProvider(), new MyGraphLabelDecorator());
}
```

and put all the code from `selfStyleConnection(Object element, GraphConnection connection)` and `selfStyleNode(Object element, GraphNode node)` into the new `decorateConnection(GraphConnection connection)` and `decorateNode(GraphNode node)` methods. I get the `element` from the `getData()` methods.

In my case I have no use for the `LabelProvider` as all the work is done in `GraphLabelDecorator`. Could there be a constructor that provides a default `LabelProvider`?

```java
public DecoratingGraphLabelProvider(IGraphLabelDecorator decorator) {
	super(new LabelProvider(), decorator);
}
```

ptziegler:
> In my case I have no use for the LabelProvider as all the work is done in GraphLabelDecorator. Could there be a constructor that provides a default LabelProvider?

A `DecoratingLabelProvider` without `LabelProvider` isn't possible. However, the `IGraphLabelDecorator`/`ILabelDecorator` interface extends the `IBaseLabelProvider` interface. If you don't care about labels, you could simply use the `GraphLabelDecorator` directly. I just need to adapt the `GraphStyler` to handle this additional case.

Phillipus:
>  If you don't care about labels, you could simply use the GraphLabelDecorator directly. 

That would be good, thanks.

Phillipus:
> If you don't care about labels, you could simply use the GraphLabelDecorator directly. I just need to adapt the GraphStyler to handle this additional case.

I tried it as PoC using a `GraphLabelDecorator` directly and modified `GraphItemStyler` with an additional `if (labelProvider instanceof IGraphLabelDecorator decorator)` and that works nicely.

Phillipus:
With the latest commit my `GraphLabelDecorator` is setting the node text and image from the label provider methods after I set them in the `decorateNode` method. It falls through to `if (labelProvider instanceof ILabelProvider)` in `GraphItemStyler`

But maybe that's intended so that `getText` and `getImage` come from the label provider and not set in `decorateNode`?

Edit - I can achieve it by my sub-class extending `GraphDecorator` and implementing `IBaseLabelProvider` so the above comment may not be valid.

Phillipus:
I set the connection line style as custom in `decorateConnection`:

```java
public void decorateConnection(GraphConnection connection) {
    PolylineConnection conn = (PolylineConnection)connection.getConnectionFigure();
    connection.setLineStyle(SWT.LINE_CUSTOM);
    conn.setLineDash(new float[] { 2, 4 });
```

But using a `IGraphDecorator` the line dashes are not drawn.

It's being reset here in GraphItemStyler:

```java
int swt = getLineStyleForZestStyle(conn.getConnectionStyle());
conn.setLineStyle(swt);
```

It works when using `ISelfStyleProvider` because that is called after `conn.setLineStyle(swt)`:

```java
int swt = getLineStyleForZestStyle(conn.getConnectionStyle());
conn.setLineStyle(swt);
if (labelProvider instanceof ISelfStyleProvider) {
	((ISelfStyleProvider) labelProvider).selfStyleConnection(conn.getData(), conn);
}
```

ptziegler:
You're right, the `IGraphDecorator` should be called last. Otherwise you'd have the same issue with e.g. `decorateText()` and `decorateImage()`, which would be overridden if the class also implements an `ILabelProvider`. That's also the order when you use the `DecoratingLabelProvider`.

Phillipus:
Latest commit is all working now, thanks!

CHANGELOG.md
```diff
@@ -6,6 +6,23 @@
...
+  `IConnectionStyleProvider`, `IConnectionStyleProvider2`
+  ⇒ Replaced by `ConnectionStyleDecorator`.
+
+  `IEntityConnectionStyleProvider`, `IEntityConnectionStyleProvider2`
+  ⇒ Replaced by `EntityConnectionStyleDecorator`.
+
+  `IEntityStyleProvider`
+  ⇒ Replaced by `EntityStyleDecorator`.
+
+  `ISelfStyler`
```
Phillipus: `ISelfStyleProvider`

org.eclipse.zest.doc.isv/guide-src/migration-guide.adoc
```diff
@@ -0,0 +1,30 @@
...
+- `IConnectionStyleProvider`, `IConnectionStyleProvider2`
+  ⇒ Replaced by `ConnectionStyleDecorator`.
+
+- `IEntityConnectionStyleProvider`, `IEntityConnectionStyleProvider2`
+  ⇒ Replaced by `EntityConnectionStyleDecorator`.
+
+- `IEntityStyleProvider`
+  ⇒ Replaced by `EntityStyleDecorator`.
+
+- `ISelfStyler`
```
Phillipus: `ISelfStyleProvider`

ptziegler: Good catch. I've updated both files.

azoitl (approved):
I like a lot how this turned out.

PR #1144, opened 2026-07-02
Reduce dependency in Zest Core from Platform UI to JFace

ptziegler:
The only dependency to the Eclipse 3.x API is via the now-deprecated style provider interface (via the IDisposable) interface. This means that the full functionality of this plug-in can be used by only using JFace classes. More specifically, this allows users to use Zest in a pure Eclipse 4.x application.

Note that the dependency to `org.eclipse.ui` has to remain until the deprecated classes have been removed. But it is now marked as optional, because it is not needed _as long as_ users use non-deprecated classes.


azoitl (approved):

ptziegler:
I noticed that this doesn't really work if you try to use a `GraphViewer` in a pure Eclipse 4.x application, because you run into a `NoClassDefFoundError` when trying to load the deprecated classes in the `GraphItemStyler`. I'd still like to keep it so that you can use the SWT widgets directly and once the deprecated classes are removed, the problem disappears anyway.

PR #1145, opened 2026-07-02
Deprecate CloudIO activator for removal

ptziegler:
The CloudIO activator is the only dependency to the Eclipse 3.x API by because it extends the AbstractUIPlugin class, which is only used to gain access to a shared `ImageRegistry`. Instead, a `ResourceManager` is used to gain access to the bundle images.

### DIFF
--- a/org.eclipse.zest.cloudio/.project
+++ b/org.eclipse.zest.cloudio/.project
@@ -20,9 +20,15 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.api.tools.apiAnalysisBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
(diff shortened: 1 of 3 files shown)


ptziegler:
I'm not sure how strict we want to be with backwards compatibility here. I don't think many people use the CloudIO plugin, let alone call the activator directly. If I keep the activator in the manifest, I can avoid the `IStartup` workaround but then the dependency to the 3.x API remains for two more years.

ptziegler:
I decided to keep the activator as it is until there is an actual requirement to use it in a pure 4.x application. Otherwise it will be possible in two years anyway.

azoitl (approved):
If we could get rid of the activator it would definitely be good. But this is already a great step forward.

ptziegler:
Yes, once the activator is gone, the dependency to the Eclipse 3.x API disappears as well. But as said, as long as there is no real requirement, we are also in no rush to remove it.

PR #1148, opened 2026-07-13
Simply print logic in `ImagePrintFigureOperation` #1146

### LINKED ISSUE #1146: IllegalArgumentException using ImagePrintFigureOperation and GC.drawImage
I've opened this new issue for the problem I encountered using `ImagePrintFigureOperation` and the Nebula `Gallery` in https://github.com/eclipse-gef/gef-classic/pull/1001.

@ptziegler said:

> As long as there isn't a reproducer, I consider this to be a Nebula bug.

This Snippet reproduces the exception using SWT only. Run on Windows:

```java
package gef;

import org.eclipse.draw2d.Figure;
import org.eclipse.draw2d.ImagePrintFigureOperation;
import org.eclipse.swt.graphics.Image;

public class DrawGC {

        Shell shell = new Shell();

        int width = 310;
        int height = 310;

        Figure figure = new Figure();
        figure.setSize(width, height);

        ImagePrintFigureOperation op = new ImagePrintFigureOperation(shell, figure);
        Image image = op.run();

            event.gc.drawImage(image, 0, 0, width, height, 0, 0, 100, 100);
        });

        shell.open();

        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }

        image.dispose();
    }
}
```

The exception is thrown when `zoom` < 100 in `ImagePrintFigureOperation.getImageDataAtZoom(int zoom)`.

```
Exception in thread "main" java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4915)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.computeSourceRectangle(GC.java:1251)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.lambda$0(GC.java:1223)
	at org.eclipse.swt.graphics.Image.executeOnImageHandleAtBestFittingSize(Image.java:946)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.draw(GC.java:1222)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.apply(GC.java:1204)
	at org.eclipse.swt.graphics.GC.storeAndApplyOperationForExistingHandle(GC.java:6102)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1134)
```

I've tried to simulate what I'm getting using `ImagePrintFigureOperation` to create an image from a `Figure` and displaying it in a Nebula `Gallery`. The Gallery widget calls `drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight)` as in the Snippet.

This may be an SWT issue but I'm wondering if something could be done in `ImagePrintFigureOperation.getImageDataAtZoom` to work around this?

For now I've added:

```java
if(zoom < 100) {
    return null;
}
```

(Edited - happens on all display scales including 100%)

ptziegler:
This removes the explicit caching of the image data for the current zoom level, as the same data is already stored in the SWT image and thus redundant. Additionally, calculating the scaled image size is now done using `Math.round()` instead of `Math.floor()`, to match the size that is expected by SWT which may otherwise cause a validation error.

Closes https://github.com/eclipse-gef/gef-classic/issues/1146

### DIFF
--- a/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ImagePrintFigureOperationTest.java
+++ b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/ImagePrintFigureOperationTest.java
@@ -16,6 +16,7 @@ package org.eclipse.draw2d.test;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.widgets.Shell;
@@ -53,6 +54,56 @@ public class ImagePrintFigureOperationTest {
 				return true;
 			}
 		});
+	}
+
+	/**
+	 * @see <a href="https://github.com/eclipse-gef/gef-classic/issues/1146">Issue
+	 *      1146</a>
+	 */
+	@Test
+	public void testPaintWithAutoScaling() {
+		figure.setSize(310, 310);
+		Image image = new ImagePrintFigureOperation(shell, figure) {
+			@Override
+			protected boolean isDraw2DAutoScalingEnabled() {
+				return true;
+			}
+		}.run();
+
+		GC gc = new GC(shell);
+		gc.drawImage(image, 0, 0, 310, 310, 0, 0, 100, 100);
+		gc.dispose();
+
+		image.dispose();
+	}
+
+	/**
+	 * SWT uses {@link Math#round(double)} for rounding, while Draw2D uses
+	 * {@link Math#floor(double)}. For a zoom < 100% this leads to a mismatch which
+	 * results in an {@link IllegalArgumentException} when trying to paint the
+	 * image, as the "scaled" image doesn't match the expected size due to an
+	 * off-by-one rounding error.
+	 *
+	 * @see <a href="https://github.com/eclipse-gef/gef-classic/issues/1146">Issue
+	 *      1146</a>
+	 */
+	@Test
+	public void testImageBoundsWithAutoScaling() {
+		figure.setSize(310, 310);
+		Image image = new ImagePrintFigureOperation(shell, figure) {
+			@Override
+			protected boolean isDraw2DAutoScalingEnabled() {
+				return true;
+			}
+		}.run();
+
+		for (int i = 1; i < 100; ++i) {
+			ImageData imageData = image.getImageData(i);
+			assertEquals(Math.round(310 * i / 100.0), imageData.width);
+			assertEquals(Math.round(310 * i / 100.0), imageData.height);
+		}
+
+		image.dispose();
 	}
 
 	@Test
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/ImagePrintFigureOperation.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/ImagePrintFigureOperation.java
@@ -15,7 +15,6 @@ package org.eclipse.draw2d;
 
 import java.util.Objects;
 
-import org.eclipse.swt.SWTException;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
@@ -24,7 +23,6 @@ import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 
 import org.eclipse.draw2d.geometry.Dimension;
-import org.eclipse.draw2d.internal.ImageUtils;
 import org.eclipse.draw2d.internal.InternalDraw2dUtils;
 
 /**
@@ -39,7 +37,6 @@ import org.eclipse.draw2d.internal.InternalDraw2dUtils;
  */
 public class ImagePrintFigureOperation {
 	private final Display display;
-	private final Control control;
 	private IFigure printSource;
 
 	/**
@@ -64,7 +61,6 @@ public class ImagePrintFigureOperation {
 	 */
 	protected ImagePrintFigureOperation(Control control) {
 		Objects.requireNonNull(control, "Control must not be null"); //$NON-NLS-1$
-		this.control = control;
 		this.display = control.getDisplay();
 	}
 
@@ -154,14 +150,7 @@ public class ImagePrintFigureOperation {
 	 * @return The image if Draw2D-based auto-scaling is enabled.
 	 */
 	private Image createAutoScaledImage() {
-		double scale = InternalDraw2dUtils.calculateScale(control);
-		Image image = createImageAtScale(scale);
-		try {
-			ImageData imageData = image.getImageData(100);
-			return new Image(display, new ScaledImageDataProvider(imageData, (int) (scale * 100)));
-		} finally {
-			image.dispose();
-		}
+		return new Image(display, new ScaledImageDataProvider());
 	}
 
 	/**
@@ -189,7 +178,9 @@ public class ImagePrintFigureOperation {
 
 		Dimension size = getImageSize().getCopy();
 		if (isDraw2DAutoScalingEnabled()) {
-			size.scale(scale);
+			// size.scale(scale) internally use Math.floor(), but SWT requires Math.round()
+			size.width = (int) Math.round(size.width * scale);
+			size.height = (int) Math.round(size.height * scale);
 		}
 
 		Image image = new Image(display, size.width, size.height);
@@ -218,29 +209,17 @@ public class ImagePrintFigureOperation {
 	}
 
 	/**
-	 * {@link ImageData} provider containing the image data at the current monitor
-	 * zoom and at 100% zoom (as required by SWT). This class is used if
-	 * Draw2D-based scaling to make sure that the image remains sharp.
+	 * {@link ImageData} provider calculating the image data for a given zoom level.
+	 * This class is used if Draw2D-based scaling to make sure that the image
+	 * remains sharp.
 	 */
 	private class ScaledImageDataProvider implements ImageDataProvider {
-		private final ImageData imageData;
-		private final int zoom;
-
-		public ScaledImageDataProvider(ImageData imageData, int zoom) {
-			this.imageData = imageData;
-			this.zoom = zoom;
-		}
-
 		@Override
 		public ImageData getImageData(int zoom) {
-			if (this.zoom == zoom) {
-				return imageData;
-			}
-			try {
+			if (zoom > 0) {
 				return getImageDataAtZoom(zoom);
-			} catch (SWTException e) {
-				return ImageUtils.smoothScaleTo(display, imageData, zoom * 1.0 / this.zoom);
 			}
+			return null;
 		}
 	}
 }
--- a/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/ImageUtils.java
+++ b/org.eclipse.draw2d/src/org/eclipse/draw2d/internal/ImageUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Patrick Ziegler and others.
+ * Copyright (c) 2025, 2026 Patrick Ziegler and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -19,12 +19,7 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ServiceConfigurationError;
 
-import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
-import org.eclipse.swt.graphics.Device;
-import org.eclipse.swt.graphics.GC;
-import org.eclipse.swt.graphics.Image;
-import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.graphics.ImageLoader;
 
 /**
@@ -78,32 +73,5 @@ public final class ImageUtils {
 			isSvgSupported = false;
 		}
 		return isSvgSupported;
-	}
-
-	/**
-	 * Scales the {@link ImageData} by the given factor using anti-alias and
-	 * interpolation.
-	 *
-	 * @param device The device to draw on.
-	 * @param data   The image data to scale.
-	 * @param scale  The factor by which the image is scaled.
-	 */
-	public static ImageData smoothScaleTo(Device device, ImageData data, double scale) {
-		int scaledWidth = Math.max(1, (int) (data.width * scale));
-		int scaledHeight = Math.max(1, (int) (data.height * scale));
-
-		Image source = new Image(device, data);
-		Image target = new Image(device, scaledWidth, scaledHeight);
-		GC gc = new GC(target);
-		try {
-			gc.setAntialias(SWT.ON);
-			gc.setInterpolation(SWT.HIGH);
-			gc.drawImage(source, 0, 0, data.width, data.height, 0, 0, scaledWidth, scaledHeight);
-			gc.dispose();
-			return target.getImageData(100);
-		} finally {
-			target.dispose();
-			source.dispose();
-		}
 	}
 }


ptziegler:
@Phillipus fyi

Phillipus:
Thanks! Will get back to you on this one, some things I need to check with it...

Phillipus:
@ptziegler Thanks for your work on this, I've now done some preliminary tests with this PR:

- IllegalArgumentException with the snippet with low values in [#1146](https://github.com/eclipse-gef/gef-classic/issues/1146) - **fixed**
- IllegalArgumentException with small image sizes in the Nebula Gallery widget as used in Archi - **fixed**
- Increased memory usage as mentioned [here](https://github.com/eclipse-gef/gef-classic/pull/1001#issuecomment-4934128907) - this is improved but still high compared with generating images the "old" way.

**Some thoughts**

`ImagePrintFigureOperation` is definitley needed on Windows when creating Images from a Figure (typically the `LayerConstants.PRINTABLE_LAYERS` layer) with new scaling support otherwise text can be clipped on the right side:

clipped-right

As metioned [here](https://github.com/eclipse-gef/gef-classic/issues/977#issuecomment-3823292440) I had the idea mentioned elsewhere to add `control.setAutoscalingMode(AutoscalingMode.ENABLED)` to the control used in the `GraphicalViewerImpl`. That fixes the clipping on the right-hand side but a new problem is the text height is miscalculated:

clipped-bottom

This is because, as @ptziegler says, "this is functionally identical to `System.setProperty("draw2d.enableAutoscale", "false");`"

So the only way it seems to get correct text on Windows in Images is to use `ImagePrintFigureOperation`:

perfect

**Some issues**

However, the issue I have with this new operation is that it's fine if just a few images are being generated from diagrams, saved to file and immediately disposed. But when there are hundreds generated there is a performance hit in terms of speed and heap memory usage.

When generating 750 diagram images for the Nebula Gallery using the "old" method (which doesn't use an `ImageDataProvider`) there seems to be no additional heap memory usage and overall app heap usage is around 80Mb. But now using the new `ImagePrintFigureOperation` when `ImageDataProvider.getImageData(zoom)` is called the overall app heap memory usage is around 2100Mb and not released until the Images are disposed.

I realise that this is due to the way SWT `Image` and `ImageDataProvider` works now but it's a side effect of trying to solve the original text clipping issue. Is this the only way to solve this?

BTW - this is not a criticism of your great work on this issue.

ptziegler:
> But now using the new ImagePrintFigureOperation when ImageDataProvider.getImageData(zoom) is called the overall app heap memory usage is around 2100Mb and not released until the Images are disposed.

An increase in memory by a factor of 26 sounds incredibly excessive. Were you able to determine what exactly is allocating this much space? My guess is that with the `ImageDataProvider`, the image now stores the image data for multiple zoom levels, rather than just the one at 100% (or rather: the current zoom). So you end up with a lot of data, most of which probably isn't needed.

ptziegler:
> I realise that this is due to the way SWT Image and ImageDataProvider works now but it's a side effect of trying to solve the original text clipping issue. Is this the only way to solve this?

As far as I remember, then "real" magic comes from this line here, because it e.g. forces the font to be drawn at the correct sizes:

https://github.com/eclipse-gef/gef-classic/blob/c54d721629a05e6e81d2145847f48881fcdfdb94/org.eclipse.draw2d/src/org/eclipse/draw2d/ImagePrintFigureOperation.java#L198-L201

So in principle it should be sufficient for you to keep your original implementation and just add an `image.getImageData(100)` before painting the figures.

Phillipus:
> So in principle it should be sufficient for you to keep your original implementation and just add an image.getImageData(100) before painting the figures.

I tried that but the image is blank. This idea was mentioned in https://github.com/eclipse-gef/gef-classic/issues/977#issuecomment-3841465942

Phillipus:
> An increase in memory by a factor of 26 sounds incredibly excessive. Were you able to determine what exactly is allocating this much space? My guess is that with the `ImageDataProvider`, the image now stores the image data for multiple zoom levels, rather than just the one at 100% (or rather: the current zoom). So you end up with a lot of data, most of which probably isn't needed.

I think you're right. Creating the Image using the original method doesn't use an `ImageDataProvider` and so heap memory is not used:

```java
Image image = new Image(null, width, height);
GC gc = new GC(image);
SWTGraphics graphics = new SWTGraphics(gc);
figure.paint(graphics);
gc.dispose();
graphics.dispose();
return image;
```

With 750 images:

low

But using an Image that has an `ImageDataProvider` as in `ImagePrintFigureOperation`:

high

ptziegler:
> I tried that but the image is blank. This idea was mentioned in https://github.com/eclipse-gef/gef-classic/issues/977#issuecomment-3841465942

You're right. That somehow only works when painting the image on a shell with 100% monitor zoom...

ptziegler:
I've slightly modified your example, where I create 800 images, which take up a total of 800MB of memory.

<details>
<summary>Program</summary>

```java
package gef;

import java.util.ArrayList;
import java.util.List;

import org.eclipse.swt.graphics.Image;

import org.eclipse.draw2d.Figure;
import org.eclipse.draw2d.ImagePrintFigureOperation;

public class DrawGC {

		Shell shell = new Shell();

		int width = 310;
		int height = 310;

		Figure figure = new Figure();
		figure.setSize(width, height);

		List<Image> images = new ArrayList<>();
		for (int i = 0; i < 800; ++i) {
			ImagePrintFigureOperation op = new ImagePrintFigureOperation(shell, figure);
			Image image = op.run();
			images.add(image);
		}

			for (Image image : images) {
				event.gc.drawImage(image, 0, 0, width, height, 0, 0, 100, 100);
			}
		});

		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}

		images.forEach(Image::dispose);
	}
}
```
</details>

The suspect is quite clear...

image

...specifically the image-data provider, which consumes 2/3 of the memory. This is without this PR.

image

With this PR, memory consumption drops to 230MB. which is almost exclusively consumed by the image-data cache of the SWT image-data provider wrapper.

image

Creating images without an image-data provider doesn't have a measureable memory footprint. I think that's because the image data isn't stored in the Java heap but directly in the native widget. So you won't see the "actual" consumption with a profiler.

If I draw the image using `drawImage(image, 0, 0)`, I also don't see this massive memory overhead. So I the `DrawScalingImageToImageOperation` that is used internally on Windows might have some unpleasant side-effects. This memory overhead also shows up when calling `drawImage(image, 0, 0, width, height)`, even though that should be equivalent to `drawImage(image, 0, 0)`.

As some noteworthy mention which I think are bugs in the SWT implementation:

- Painting an "empty" image (i.e. an image for which no GC was created) fails with a `NullPointerException`.
- Creating 8000 images and then trying to paint them eventually throws an `IllegalArgumentException`, even though I would expect a "No more handles" error.

ptziegler:
@Phillipus Do you remember how much memory the _process_ was consuming with 750 images? This is what I got from the snippet when creating 4000 images via `new Image(...)`. The Java heap is almost empty and yet this very simple example requires 1GB of memory.

image

If I run the original snippet, where the heap takes up 250MB of memory, the process also consumes 1.1GB of memory. So there is definitely space being wasted here...

Phillipus:
I use the Eclipse Heap memory widget to check the heap size as seen in the 2 screenshots in https://github.com/eclipse-gef/gef-classic/pull/1148#issuecomment-4979903369

In the case of drawing 750 images in Nebula in Archi and with 2100Mb heap usage in the memory widget I have this for the process:

memory

Some observations:

- This PR improves the memory heap usage. Without it the heap usage can grow to around 4000Mb
- If I later call `Image.getImageData(200)` - this increases the heap memory usage and it is not released until the image is disposed

So with a combination of this PR and not calling `Image.getImageData(200)` I can reduce the memory usage to:

new heap

memory2

ptziegler:
Thanks for checking again. So this pretty much confirms what I've seen as well. With this PR, the memory consumption goes down. But it's still bad.

> If I later call Image.getImageData(200) - this increases the heap memory usage and it is not released until the image is disposed

That is expected. The `cachedImageData` map of the `Image` contains the image data for each zoom level. So once this value is loaded, it will stay in memory for as long as the image exists.

Phillipus:
> That is expected. The `cachedImageData` map of the `Image` contains the image data for each zoom level. So once this value is loaded, it will stay in memory for as long as the image exists.

Yes as this shows:

```java
import java.util.ArrayList;
import java.util.List;

import org.eclipse.draw2d.Figure;
import org.eclipse.draw2d.ImagePrintFigureOperation;
import org.eclipse.swt.graphics.Image;

public class MemTest {

        Shell shell = new Shell();
        
        int width = 310;
        int height = 310;

        Figure figure = new Figure();
        figure.setSize(width, height);

        List<Image> images = new ArrayList<>();
        for(int i = 0; i < 800; ++i) {
            ImagePrintFigureOperation op = new ImagePrintFigureOperation(shell, figure);
            Image image = op.run();
            images.add(image);
        }
        
        System.out.println("Memory used before: " + getUsedMemory() + "Mb");
        images.forEach(image -> image.getImageData(200));
        System.out.println("Memory used after: " + getUsedMemory() + "Mb");
        
        images.forEach(Image::dispose);
    }
    
    static long getUsedMemory() {
        Runtime runtime = Runtime.getRuntime();
        runtime.gc(); // GC
        
        long totalMem = runtime.totalMemory();
        long freeMem = runtime.freeMemory();
        return (totalMem - freeMem) / (1024 * 1024);
    }
}
```

I get:

```
Memory used before: 5Mb
Memory used after: 1605Mb
```

ptziegler:
There is already a PR on the SWT side, which may fix the performance/memory overhead: https://github.com/eclipse-platform/eclipse.platform.swt/pull/3429

ptziegler:
I've also started a discussion to see if there's a better way to reduce the memory overhead of such images: https://github.com/eclipse-platform/eclipse.platform.swt/discussions/3445

Phillipus:
@ptziegler Thanks for opening the discussion and spending time on this.

As far as this PR goes it fixes #1146 and reduces memory footprint.

I have other problems with using `ImagePrintFigureOperation` so rather than cluttering up this PR with them I opened a [discussion](https://github.com/eclipse-gef/gef-classic/discussions/1151).

ptziegler:
Sorry for the long delay. I finally managed to get back to this issue now.

Based on https://github.com/eclipse-gef/gef-classic/discussions/1151#discussioncomment-17736978, I made a small example which zooms in on the image created by the `ImageFigurePrintOperation` and an image painted with an `SWTGraphics`, to make sure that there is no degradation in quality.

Screenshot 2026-08-26 210658

@Phillipus This change won't make it into the current release, but perhaps you can still try it out to see if it works with your figures?

I got rid of the ImageDataProvider for the image returned by this operation. This was initially done to resolve a quirk of the SWT implementation, where (because the image data is created at 100% zoom) a blank image is painted when at monitor zoom. So this data has to be converted to make the image usable again.

Phillipus:
@ptziegler Thanks for working on this again.

I tried the latest PR but unfortunately this makes the image quality worse.

This is on Windows 11 with 200% monitor scaling.

**Not** using `ImagePrintFigureOperation` the quality is good but the "g" in "Management" is clipped as expected:

before

Using `ImagePrintFigureOperation` the quality is blurred (but the "g" is not clipped):

after

ptziegler:
> I tried the latest PR but unfortunately this makes the image quality worse.

Yes, I made a silly mistake and accidentally hard-coded the monitor zoom to 150%. So the image data at 100% was scaled to 200%, which is of course blurry.

Phillipus:
I tried again with the latest commit and it's much better but compared with not using `ImagePrintFigureOperation` there is some blurriness on the lines.

Original:

before

With latest PR using `ImagePrintFigureOperation`:

latest

ptziegler:
Then let's try to also disable interpolation... If that also doesn't work, then using an ImageDataProvider will be the only other option. Because there no simple way to create an image with image data that is not for 100% monitor zoom.

Phillipus:
> Then let's try to also disable interpolation... If that also doesn't work, then using an ImageDataProvider will be the only other option. 

I don't think it makes much difference.

This is using the `ImagePrintFigureOperation` in the master branch:

before

And this is with the latest PR:

after

ptziegler:
Can you point me to the class where the image is created/painted? Because at this point I question whether this is an issue with the print operation.

The version on the master branch especially will always produce pixel-perfect data. Any blurriness _must_ therefore come from somewhere else.

Phillipus:
This is basically what I do:

```java
GraphicalViewerImpl viewer = new GraphicalViewerImpl();
viewer.createControl(parent);
viewer.setEditPartFactory(editPartFactory);
RootEditPart rootPart = new FreeformGraphicalRootEditPart();
viewer.setRootEditPart(rootPart);
viewer.setContents(model);
viewer.flush();

LayerManager layerManager = (LayerManager)graphicalViewer.getEditPartRegistry().get(LayerManager.ID);
IFigure rootFigure = layerManager.getLayer(LayerConstants.PRINTABLE_LAYERS);

Shell shell = new Shell();
try {
    ImagePrintFigureOperation op = new ImagePrintFigureOperation(shell, rootFigure);
    Image image = op.run();

    ImageData imageData = image.getImageData(200); // 200 = current display scale
    ImageLoader loader = new ImageLoader();
    loader.data = new ImageData[] { imageData };
    loader.save("image_name.png", SWT.IMAGE_PNG);
}
finally {
    shell.dispose();
}
```

ptziegler:
It would've been really nice to have known how the figures are painted before digging though the implementation...

In short, the "source" image is already blurry after calling:

```java
try {
    preparePrintSource(graphics);
    printSource.paint(graphics);
} finally {
    restorePrintSource(graphics);
}
```

So the issue is not the post-processing or anything else I'm doing in the `ImagePrintFigureOperation`. Instead, this issue happens when two conditions are met:

1) Anti-aliasing must be enabled. This is explicitly set in the [AbstractContainerFigure](https://github.com/archimatetool/archi/blob/036a51fec00ac4ac6dc74209ec71084a5868d517/com.archimatetool.editor/src/com/archimatetool/editor/diagram/figures/AbstractContainerFigure.java#L94-L101).

2) The graphics object must be translated by a fractional value. This is done implicitly in the [AbstractDiagramModelObjectFigure](https://github.com/archimatetool/archi/blob/036a51fec00ac4ac6dc74209ec71084a5868d517/com.archimatetool.editor/src/com/archimatetool/editor/diagram/figures/AbstractDiagramModelObjectFigure.java#L224-L225)

Neither is something this operation has any control over, so I don't see how this is something I can address here. I've updated the snippet to illustrate this problem.

image

The screenshot above is on Linux, so I would argue that this is the correct behavior.

Phillipus:
> It would've been really nice to have known how the figures are painted before digging though the implementation...

Hmmm...it seems my standalone snippets missed something....sorry.

> Anti-aliasing must be enabled. 

Yes. This is what was missing.

> The graphics object must be translated by a fractional value

No, that doesn't affect it, it's only setting `graphics.setAntialias(SWT.ON)`. But the figures must have this on otherwise rounded rectangles and the like are jagged.

My workaround is to do `viewer.createControl(parent).setAutoscalingMode(AutoscalingMode.ENABLED); ` and not use `ImagePrintFigureOperation` (just a simple `figure.paint(graphics)`) when creating the image and live with text height anomalies.

> so I don't see how this is something I can address here. 

I think you're right. The reason the image figure lines are blurry is because anti-aliasing is on and that is how things are in the diagram as well as the image, as noted [here](https://github.com/eclipse-gef/gef-classic/issues/1076).

ptziegler:
> No, that doesn't affect it, it's only setting graphics.setAntialias(SWT.ON). But the figures must have this on otherwise rounded rectangles and the like are jagged.

I think the monitor zoom plays into that because occasionally, I saw that as well. But this is the snippet on Linux without translation.

image

...and this with translation.

image

Phillipus:
A key missing factor (at least on Windows) is `g.setLineWidth(1)`.

Try this:

```java
IFigure printSource = new Figure() {
	@Override
	public void paintFigure(Graphics g) {
		Rectangle rect = getBounds().getCopy().resize(-1, -1);
		g.translate(0.5f, 0.5f);
		g.setAntialias(SWT.ON);
		g.setLineWidth(1);
		g.setBackgroundColor(ColorConstants.cyan);
		g.fillRectangle(rect);
		g.setBackgroundColor(ColorConstants.darkGray);
		g.drawRectangle(rect);
	}
};
```

Image 001

And also eliminating `g.translate(0.5f, 0.5f);` will get the same result:

```
IFigure printSource = new Figure() {
	@Override
	public void paintFigure(Graphics g) {
		Rectangle rect = getBounds().getCopy().shrink(1, 1);
		g.setAntialias(SWT.ON);
		g.setLineWidth(1);
		g.setBackgroundColor(ColorConstants.cyan);
		g.fillRectangle(rect);
		g.setBackgroundColor(ColorConstants.darkGray);
		g.drawRectangle(rect);
	}
};
```

So, these are the important things:

```java
g.setAntialias(SWT.ON);
g.setLineWidth(1); // or an odd number like 3, 5
```

This is on Windows 11 at 200% monitor scaling.

Phillipus:
This is certainly an improvement on the existing `ImagePrintFigureOperation` - uses much less memory too - so perhaps you might like to merge it?

What I'm doing for Archi next version is to add a preference on Windows to use this version of `ImagePrintFigureOperation` (I made a copy of it). The resulting images are not so sharp (anti-aliased) but there is no text clipping or miscalculated text heights.

Archi users can opt in to this but can choose the default option where I use the workaround of setting `viewer.createControl(parent).setAutoscalingMode(AutoscalingMode.ENABLED);` and not using `ImagePrintFigureOperation`. The images are crisper but the text heights can be wrong. Users can make the height of the figures a little higher to accommodate this.

It's just a shame that the root cause (differing text height calculations at different display scales) can't be fixed.

ptziegler:
Yes, at the end of the day, both the text height calculation and the issues with fractional scaling seem to be a limitation of GDI/GDI+. I don't think there is anything more that can be done from our side. Sorry, the past few weeks have been rather hectic, so the GEF tasks took the short end of the stick.

Phillipus:
> Yes, at the end of the day, both the text height calculation and the issues with fractional scaling seem to be a limitation of GDI/GDI+. I don't think there is anything more that can be done from our side. Sorry, the past few weeks have been rather hectic, so the GEF tasks took the short end of the stick.

No need to apologise, you've worked hard on this feature and I greatly appreciate that. Thank-you!

One last question - do you think `gc100.setInterpolation(SWT.NONE);` makes any difference rather than using `SWT.SWT.HIGH` or some other value?

PR #1157, opened 2026-07-24
Remove coupling of custom graph stylers to GraphLabelDecorator

ptziegler:
This changes the `IConnectionStyleDecorator`, `IEntityStyleDecorator` and the `IEntityConnectionStyleDecorator` from abstract classes to interfaces and moves the default implementation to the `GraphLabelProvider`.

Problem with the current implementation is that consumers who implement more than one of the now-deprecated style providers can't easily migrate to the new decorators, as Java doesn't allow sub-classing.

By decoupling those extensions from the `IGraphLabelDecorator` interface, we remove this limitation while at the same time preserving the ability to add new methods to mentioned interface, without breaking backwards compatibility.

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/IConnectionStyleProvider.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/IConnectionStyleProvider.java
@@ -15,7 +15,7 @@ package org.eclipse.zest.core.viewers;
 import org.eclipse.swt.graphics.Color;
 
 import org.eclipse.ui.services.IDisposable;
-import org.eclipse.zest.core.viewers.decorators.ConnectionStyleDecorator;
+import org.eclipse.zest.core.viewers.decorators.IConnectionStyleDecorator;
 
 import org.eclipse.draw2d.IFigure;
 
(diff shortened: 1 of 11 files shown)


ptziegler:
Main problem with these decorators is binary compatibility. The reason we always extend the `GraphLabelDecorator` is so that we can freely extend the `IGraphLabelDecorator` interface in the future. Which is why I decided to not have the custom decorators extend the `IGraphLabelDecorator`, but to remove the relationship alltogether.

So consumers now have the `GraphLabelDecorator` as the single entry point, but can customize the styling of nodes and connections by implementing the now-interfaces.

@merks I suppose that's how you'd expect it to look like?

image

Below the patch for the required changes, based on this PR. It's a little bit tricky because you're already extending `AdapterFactoryLabelProvider.FontProvider`, but that's what the `DecoratingLabelProvider` is for.

[p2repo-aggregator.patch](https://github.com/user-attachments/files/30356987/p2repo-aggregator.patch)

merks:
Yes, that looks as expected.

merks (approved):
This looks more flexible.

merks:
@ptziegler 

Is it planned to merge this for M2?

ptziegler:
Ah, yes. There were some errands I had to do over the weekend, but the plan was to get this merged ASAP. :)

merks:
Thanks.  I tried the nightly build with your patch and it works great!  

---

FYI, all the build warnings from the latest nightly build are a result of the mirror PHP script malfunctioning:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7712#note_8080701

azoitl:
I was about to merge it now (also didn't find time earlier). But @ptziegler was faster. I like the improved design a lot.

PR #1159, opened 2026-08-02
Simplify validation done in `GraphicsTest`

ptziegler:
This change streamlines the way the paint events in our `Graphics` recorder are tracked. Rather than having a field for storing the parameters of each method call, the information is now encoded as a string and stored in the `events` list.

This list is stored in the new `GraphicsRecorder` class, which is a replacement for the `RecordingSwtGraphics` class. Recording is done via two instances of this class. First, the "actual" recorder keeps track of the methods that were called by the draw operation. Next, the "expected" recorder is called with the expected method called. Then both event lists are compared for equality.

The advantage of this approach is that we not only keep track of the method calls, rather than just the last call and/or the aggregated result.


azoitl (approved):
cool improvement

PR #1165, opened 2026-08-05
Improve feedback of figure-under-cursor in GraphSnippet7

ptziegler:
Rather than printing the figure to the console, it is now shown in a label underneath the graph. The information is now also split into the cursor location and the actual figure, to also provide feedback in case no figure is found.

To improve the result of the `toString()` method, default implementations for the `Label` and `AbstractPointListShape` have been added.


ptziegler:
The bind(...) method has only recently been adapted to allow varargs. So I've split this string into the location and figure, because there are dedicated bind(...) methods for one and two parameters, respectively.

image

azoitl (approved):

merks:
FYI, the adjustments were easy and are done now.  Thanks!

I used org.eclipse.draw2d.ColorConstants.lightBlue instead of org.eclipse.zest.core.widgets.Graph.LIGHT_BLUE where it would have been helpful to have that suggested in the deprecation.   😶

ptziegler:
> I used org.eclipse.draw2d.ColorConstants.lightBlue instead of org.eclipse.zest.core.widgets.Graph.LIGHT_BLUE where it would have been helpful to have that suggested in the deprecation.

I think this was meant for https://github.com/eclipse-gef/gef-classic/pull/1170? I was a little hesitant to mention the `ColorConstant` fields because they aren't a drop-in replacement, depending on your use-case. 

My understanding is that the colors are public, so that clients can change the theme for the graph. But when switching to `ColorConstants.lightBlue`, you lose this connection. So when the graph switches to a different color palette, those colors may no longer fit together.

Nonetheless, I've updated the documentation as part of #1172.

merks:
Thanks.

PR #1170, opened 2026-08-06
Mark public color fields in `Graph` for removal

ptziegler:
Having public fields is an anti-pattern because it allows users to modify those fields at any point. The colors should instead be set directly via e.g. `setBackgroundColor(...)` when creating the graph items.


ptziegler:
I think I got confused by the fields being capitalized. They aren't actually static, so the initial argument doesn't hold.

azoitl (approved):

PR #1171, opened 2026-08-06
Distinguish Zest node foreground color between highlighted and not

ptziegler:
Depending on the foreground color used for Zest nodes, it might be hard to read any displayed text when they are selected. For example if the text is shown in white and the highlight color is bright.

This change splits up the foreground color into two separate fields, which can be set with the `getForegroundHighlightColor()` and the `setBackgroundHighlightColor()` (formerly the `setHighlightColor()`) methods.

To ensure backwards compatibility, the graph node will fall back to the "normal" foreground color, if no highlight color is set.

Contributes to https://github.com/eclipse-gef/gef-classic/issues/1090

### DIFF
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/IEntityStyleDecorator.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/IEntityStyleDecorator.java
@@ -94,13 +94,22 @@ public interface IEntityStyleDecorator {
 	Color getForegroundColor(Object entity);
 
 	/**
-	 * Returns the foreground color of this entity. May return {@code null} for
-	 * defaults.
+	 * Returns the foreground highlight color of this entity. May return
+	 * {@code null} for defaults.
 	 *
 	 * @param entity the entity to be styled.
 	 * @return the foreground color of this entity.
 	 */
-	Color getNodeHighlightColor(Object entity);
+	Color getNodeForegroundHighlightColor(Object entity);
+
+	/**
+	 * Returns the background highlight color of this entity. May return
+	 * {@code null} for defaults.
+	 *
+	 * @param entity the entity to be styled.
+	 * @return the background color of this entity.
+	 */
+	Color getNodeBackgroundHighlightColor(Object entity);
 
 	/**
 	 * Returns the tool-tip for this node. If {@code null} is returned Zest will
--- a/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/GraphLabelDecorator.java
+++ b/org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/GraphLabelDecorator.java
@@ -95,7 +95,8 @@ public class GraphLabelDecorator extends LabelProvider implements IGraphLabelDec
 
 			Optional.ofNullable(decorator.getBorderColor(entity)).ifPresent(node::setBorderColor);
 			Optional.ofNullable(decorator.getBorderHighlightColor(entity)).ifPresent(node::setBorderHighlightColor);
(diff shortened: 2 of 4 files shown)


org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/IEntityStyleDecorator.java
```diff
@@ -94,13 +94,22 @@ public interface IEntityStyleDecorator {
...
+	Color getNodeForegroundHighlightColor(Object entity);
+
+	/**
+	 * Returns the background highlight color of this entity. May return
+	 * {@code null} for defaults.
+	 *
+	 * @param entity the entity to be styled.
+	 * @return the background color of this entity.
+	 */
+	Color getNodeBackgroundHighlightColor(Object entity);
```
ptziegler, line 112: The interface was added with this release, so it should be fine with adding new non-default methods. Which is why I'd to include this for the current release. Thought it might introduce compilation errors for e.g. @merks in the M3.

org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/IEntityStyleDecorator.java
```diff
@@ -94,13 +94,22 @@ public interface IEntityStyleDecorator {
 	Color getForegroundColor(Object entity);
 
 	/**
-	 * Returns the foreground color of this entity. May return {@code null} for
-	 * defaults.
+	 * Returns the foreground highlight color of this entity. May return
+	 * {@code null} for defaults.
```
ptziegler, line 98: The highlight color was used for the background, so the documentation is incorrect.

org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/IEntityStyleDecorator.java
merks: I'll have to keep this in mind then when a GEF milestone build is out.    It wouldn't hurt to create an extra milestone sooner rather than later.  😁

ptziegler: > It wouldn't hurt to create an extra milestone sooner rather than later.

I'll be traveling over the weekend, but I'll see what I can do once I'm back.

merks: FYI, if this is done and you merge it, I can do the rest... I have god-like committer powers too.  😀

merks: Also, I can test against a nightly build if you prefer not to have an extra milestone build.

azoitl (approved):

org.eclipse.zest.core/src/org/eclipse/zest/core/viewers/decorators/IEntityStyleDecorator.java
azoitl: @merks I like the idea of an early milestone. Just merged it and would trigger later a milestone build. Should I also update simrel?

merks: Contribute early, contribute often is a good rule of thumb.  :-)  But it's not necessary to update SimRel too, so your choice...

azoitl: milestone is there. I may wait a bit with SimRel.

PR #1176, opened 2026-09-01
Update bundles and features for 2026-12 release

ptziegler:


pom.xml
```diff
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.eclipse.gef_root</groupId>
 	<artifactId>org.eclipse.gef.releng</artifactId>
-	<version>3.28.0-SNAPSHOT</version>
+	<version>3.29.0-SNAPSHOT</version>
```
azoitl: did we forget to update that the last time. because the next release is 3.30.0

org.eclipse.draw2d.tests/pom.xml
azoitl: the update of the test plugin version is missing.

azoitl (changes requested):

pom.xml
ptziegler: Looks like it. I've updated the version to 3.30.0.

org.eclipse.draw2d.tests/pom.xml
ptziegler: The test plugins are not part of the baseline. So formally, there is nothing to increment. I've updated them regardless.

pom.xml
azoitl: thx

org.eclipse.draw2d.tests/pom.xml
azoitl: good point maybe I overdid it here. But as you've changed it we may also keep it.

azoitl (approved):

PR #1182, opened 2026-09-05
Move PDE/JDT prefs to ".eclipse" project and reference by resource link

ptziegler:
The preferences for the GEF projects are currently copy-pasted from one project to another, which causes unnecessary work when trying to keep them synchronized and also when creating new projects.

Instead of a physical file, the projects now only contain a reference to a "shared" preference in the ".eclipse" project. Any changes to this file is then automatically applied to all other projects.


ptziegler:
This is just an idea I had, because 15% of the changes in https://github.com/eclipse-gef/gef-classic/pull/952 are just those preferences. 

image

In the Eclipse, symbolic links seem to work and are also properly restored on a switch.

ptziegler:
@azoitl wdyt? I still have to test whether this works on Windows but otherwise I'd do the same for the other preferences as well. The only issue I see is with the Java 17/21 duality but I believe that can be managed, too.

ptziegler:
Honorary mention to https://github.com/eclipse-platform/eclipse.platform/issues/89, which inspired this idea.

azoitl:
This is an amazing idea! I love it. My main dev machine is a Linux machine so for me it works as a charm. I currently have no easy access to a windows machine but if it works there as well I would definitely go with it and also consider to adopt it for 4diac IDE. There we have over 130 projects where we need to maintain the preferences.

ptziegler:
@azoitl Sadly, the symlink semantic doesn't work on Windows :frowning:. However, perhaps I can inject a small script which automatically creates the link when building the workspace?

merks:
Eclipse does support file links though.  I'm not sure exactly how they work:

image

I don't think it helps with cross project links...

ptziegler:
> Eclipse does support file links though. I'm not sure exactly how they work:

@merks I think that's exactly what we need. Many thanks for the hint! The preferences are show up in the workspace and are detected by JDT/PDE and any changes modify the files in the .eclipse folder.

image

merks:
I'm not sure if you need a task like this to ensure that linking isn't disabled:
```
<?xml version="1.0" encoding="UTF-8"?>
<setup:PreferenceTask
    xmi:version="2.0"
    xmlns:xmi="http://www.omg.org/XMI"
    xmlns:setup="http://www.eclipse.org/oomph/setup/1.0"
    key="/instance/org.eclipse.core.resources/description.disableLinking"
    value="false"/>
```

You can copy and paste that...

ptziegler:
> I'm not sure if you need a task like this to ensure that linking isn't disabled:
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <setup:PreferenceTask
>     xmi:version="2.0"
>     xmlns:xmi="http://www.omg.org/XMI"
>     xmlns:setup="http://www.eclipse.org/oomph/setup/1.0"
>     key="/instance/org.eclipse.core.resources/description.disableLinking"
>     value="false"/>
> ```
> 
> You can copy and paste that...

I always wonder how many people actually use the Oomph installer... But you're obviously right.

merks:
> I always wonder how many people actually use the Oomph installer... But you're obviously right.

I wonder how many people would know they need to change this preference to work with this project’s repository?

ptziegler:
> I wonder how many people would know they need to change this preference to work with this project’s repository?

AFAIK, linking is enabled by default. So this only would've been an issue if people active disable it.

https://github.com/eclipse-platform/eclipse.platform/blob/241fa272393d64bfc28dab0e1b8ad676f4f82647/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/PreferenceInitializer.java#L34

merks:
Hmmm. I’ve always seen it disabled (true) and there is a warning when you stop disabling it. Worse yet, I often see it ends up disabled again and the task runs again in that case (to set it to false). I’ve never tracked down what causes that. Maybe it’s different in the eclipse sdk product. Products can always change defaults. I don’t think I added the task if it was only restoring a default. But it’s worth investigating.

merks:
You're right that linked enabled is the default in the IDE packages and in the SDK too.   

If it's disabled and you enable it, you get this warning:

image

But something in my SDK IDE apparently sets it to true once in a while (and has been doing that for years) and then sometimes and then I see this in the log when it gets set back to false:
```
[2026-09-03 09:00:15] Performing Preference /instance/org.eclipse.core.resources/description.disableLinking = false
```
I see this 3 times in my 3 month old log. Hopefully that never happens to you...

Phillipus:
I'm seeing strange behaviour with this. If I check out the branch with this PR I see API warnings even though I've got my own settings that over-ride them. My settings are ignored.

And in each project's settings the "Enable project specific settings" is not checked:

prefs

It seems that projects are not being updated properly when the `*.prefs` files change.

Phillipus:
> It seems that projects are not being updated properly when the *.prefs files change.

Sometimes the settings get out of sync with the workspace. I found that I can get them to wake up if I edit the `org.eclipse.jdt.core.prefs` file and/or `org.eclipse.pde.api.tools.prefs` file, save them, and undo - in Eclipse (that's the important part).

ptziegler:
> And in each project's settings the "Enable project specific settings" is not checked:

That is "correct" as in "that's how it's implemented in PDE". The checkbox is enabled when one of the preferences specified [here](https://github.com/eclipse-pde/eclipse.pde/blob/523f495f87650156d3548ecbbc1aba9974c6b5c6/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsConfigurationBlock.java#L448-L547) is set. But we're only using the following preferences, neither of which appear in this list.

- CLASS_ELEMENT_TYPE_CHANGED_RESTRICTIONS
- INTERFACE_ELEMENT_TYPE_CHANGED_RESTRICTIONS
- METHOD_ELEMENT_TYPE_CHANGED_RESTRICTIONS
- incompatible_api_component_version_include_major_without_breaking_change
- incompatible_api_component_version_include_minor_without_api_change

But I can't find those in the PDE project and apparently, project-specific settings were removed with 68ab11fc95121e895dd5fe3fe44b5fc78075a7a7. So where do those settings even come from? @azoitl Do you remember? My guess is you simply unchecked the item in the project properties, which AFAIK only removes "known" properties. In that case I would remove those files completely.

The missing update of the preferences I can confirm and reproduce when switching back and forth between this branch and the master. If I switch to a different branch where the preferences are still in the project folders, the update happens again. So this seems like a Platform issue, which effectively blocks this change. As a side-note, I also noticed that Git doesn't automatically pick up the changes when the preferences are edited (unless done directly on the .eclipse project).

Phillipus:
> And in each project's settings the "Enable project specific settings" is not checked:

But it is then checked again when I force an update my editing one of the prefs files in Eclipse and refresh.

ptziegler:
> But it is then checked again when I force an update my editing one of the prefs files in Eclipse and refresh.

See my comment above:

> The missing update of the preferences I can confirm and reproduce when switching back and forth between this branch and the master.

I also get the behavior when I enable the project-specific settings and do a "Replace with HEAD revision" on the `.prefs` file in the `.eclipse` project. Even though no project-specific settings remain, the checkbox is still set. I assume all of those things describe exactly the same phenomenon.

My guess is that this is related to the `ProjectScope` which is used internally. In the workspace model, I would expect one scope for each project, all of which internally point to the same file. When the `.prefs` file is edited, only the scope of the `.eclipse` project is updated, while the others go out-of-sync.

Phillipus:
> That is "correct" as in "that's how it's implemented in PDE". The checkbox is enabled when one of the preferences specified [here](https://github.com/eclipse-pde/eclipse.pde/blob/523f495f87650156d3548ecbbc1aba9974c6b5c6/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsConfigurationBlock.java#L448-L547) is set. But we're only using the following preferences, neither of which appear in this list.

I just realised it was checked because I added some exclusions to the `org.eclipse.pde.api.tools.prefs` file in my fork:

```
LEAK_EXTEND=Ignore
LEAK_METHOD_RETURN_TYPE=Ignore
LEAK_IMPLEMENT=Ignore
ILLEGAL_IMPLEMENT=Ignore
ILLEGAL_REFERENCE=Ignore
```

But, yes, there is still a sync problem when switching branches or editing the `prefs` files.

azoitl:
I must say I don't remember how they came back. When I did the referenced change I wanted to address a similar problem by removing project specific settings to use default Eclipse settings. However by now I learned that this was not the best idea and therefore I re-added project specific settings for JDT. I never thought about the PDE settings so most probably they where just forgotten. Therefore having them now back in and correctly configured for Eclipse GEF Classic  would be for me the better way to go.

ptziegler:
I briefly looked into this and _I_ would claim that this is a bug on the Platform side, or at the very least a severe limitation. The project preferences are updated [automatically](https://github.com/eclipse-platform/eclipse.platform/blob/a68c37a13a73933ebe15b9ddc49b89063577bc4b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java#L590-L594) when a file in the .settings folder is changed.

However, the update mechanism only considers the project the .prefs file is [physically](https://github.com/eclipse-platform/eclipse.platform/blob/a68c37a13a73933ebe15b9ddc49b89063577bc4b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectPreferences.java#L353-L389) located in. Project that dynamically link to this file won't receive any updates and continue to work with the outdated settings.

ptziegler:
On further investigation, the Platform has an `AliasManager`, which keeps track of the linked resources. If one file is changed, the other is refreshed as well. And in my current test project, it works. Strange.

https://github.com/user-attachments/assets/3b414a97-ec27-4f46-8a09-fd4cad12530f

ptziegler:
I think I've got it. The scopes go out of sync if the files are edited by an external tool. Even a refresh won't update the linked preferences. And this just happens to be how JGit works.

Phillipus:
> I think I've got it. The scopes go out of sync if the files are edited by an external tool. Even a refresh won't update the linked preferences. And this just happens to be how JGit works.

Yes, that's what I found, but not just with JGit. That's why I mentioned:

> I found that I can get them to wake up if I edit the org.eclipse.jdt.core.prefs file and/or org.eclipse.pde.api.tools.prefs file, save them, and undo - in Eclipse (that's the important part).

ptziegler:
See https://github.com/eclipse-platform/eclipse.platform/pull/2914

PR #1188, opened 2026-09-14
Use Java 25 in builds

ptziegler:
`org.eclipse.core.filesystem` requires Java 25 since 2026-12.


ptziegler:
@azoitl Premium support for Oracle Java 17 ends this month and I don't really count the "extended support" until 2029. At that point I'd like to bring up the topic for Java 21 again, which raises the minimum required Eclipse version and thus avoids some of the reflection shenanigans we have in Draw2D.

azoitl:
I would be all in to raise to Java 21. Especially if it helps us making our code cleaner.

azoitl (approved):

ptziegler:
Looks like I missed to update the promotion project, which now requires Java 21 :/

https://github.com/eclipse-gef/gef-classic/blob/154869a104e241909cd03698da70642bd4d19348/promotion/pom.xml#L49

> [ERROR] org.eclipse.equinox.launcher 1.8.0.v20260804-1928 requires Execution Environment that matches (&(osgi.ee=JavaSE)(version=21)) but the current resolution context uses [a.jre.javase 17.0.0]

PR #1190, opened 2026-09-15
Use JavaSE-25 for promotion

merks:


merks:
@azoitl @ptziegler 

This is just a simple change to plan ahead for the day when it really will require Java 25.  The PR build will not test this, but I've this changed in many projects already so of course I expect it to work fine.

ptziegler (approved):
Thanks! I only noticed this now, so I guess #1191 is obsolete.

